PR #26, opened 2015-01-31
Fix running 'make': upgrade to infer-annotations-1.4

dimsuz:
It seems that 'buck' tool got infer-annotations upgraded from 1.3 to 1.4, so typing 'make' in console gave a 404 error while trying to build a 'java' target.
Fixed that by updating version to 1.4, now tests run and pass.


vjeux:
@astreet, @kmagiera mind looking at this?

PR #28, opened 2015-02-01
Added package.json

ColinEberhardt:
Thanks for the awesome library - I'm already using it in one of my projects. In order to manage this as a proper dependency (rather than cut and paste the code into my repo - yuck!), I need a `package.json` file. It would be great if you could add this.

PR #31, opened 2015-02-02
Added package.json

ColinEberhardt:
Updated from #28 - rebased to include #26 which fixes the build issues with the previous PR.


vjeux:
Thanks :)

PR #35, opened 2015-02-03
Changed tests to reflect new chrome text size

jimmyhmiller:
Assuming that the error message in "Layout tests consts" is accurate, these test seem to be failing simply because chrome has updated their text measurement functions. So rather than change the code, I changed the tests to reflect this change.

### DIFF
--- a/src/__tests__/Layout-test.js
+++ b/src/__tests__/Layout-test.js
@@ -890,8 +890,8 @@ describe('Layout', function() {
           {style: {margin: 20, measure: text(texts.big)}}
         ]}
       ]},
-      {width: 200, height: 58, top: 0, left: 0, children: [
-        {width: 200, height: 58, top: 0, left: 0, children: [
+      {width: 200, height: 56, top: 0, left: 0, children: [
+        {width: 200, height: 56, top: 0, left: 0, children: [
           {width: textSizes.bigWidth, height: textSizes.smallHeight, top: 20, left: 20}
         ]}
       ]}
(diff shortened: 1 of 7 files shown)


ColinEberhardt:
If these are magic numbers that are going to change from time-to-time it would be useful if they didn't appear in quite so many places!

src/__tests__/Layout-test.js
```diff
@@ -905,8 +905,8 @@ describe('Layout', function() {
           {style: {margin: 20, measure: text(texts.big)}}
         ]}
       ]},
-      {width: 200, height: 76, top: 0, left: 0, children: [
-        {width: 200, height: 76, top: 0, left: 0, children: [
+      {width: 200, height: 72, top: 0, left: 0, children: [
+        {width: 200, height: 72, top: 0, left: 0, children: [
```
vjeux: I replaced this 72 constant with textSizes.bigHeight + 40, this way it'll won't fail with the next update

vjeux:
Thanks for this pull request! I merged in in two separate commits.

The C version was out of date for some (bad) reason. https://github.com/facebook/css-layout/commit/5f8c3b0eb744b91db8e98be005fed0a875961447

Updated the two constants and made one test use the constants so that i doesn't fail anymore https://github.com/facebook/css-layout/commit/0e56641b28ce5a598fe0253f1c76e9d763cd2c8a

@ColinEberhardt the constants are already isolated. Next time there's a change, we just have to update `preDefinedTextSizes` in Layout-test-utils.js and run make to propagate the changes to C and Java. At Facebook we collapse changes for files that are tagged `@generated` but github doesn't :(

PR #37, opened 2015-02-04
computeLayout now adds layout, children and style properties as required

### LINKED ISSUE #24: Example in readme does not run.
Running the example in the readme will result in "node.layout is undefined"  error.

```javascript
computeLayout(
  {style: {padding: 50}, children: [
    {style: {padding: 10, alignSelf: 'stretch'}}
  ]}
);
```

This doesn't occur in the tests because of the preprocessing the node goes through in the testing suite. Particularly in the [computeCSSLayout](https://github.com/facebook/css-layout/blob/28243156e48cc6e3cb8adb809e03406c5970eb41/src/Layout-test-utils.js#L131) function.

I'm not sure if this is the expected behavior for the library or if the preprocessing should be moved internally, but either way the readme example should change to reflect this.

ColinEberhardt:
Closes #24 

I've removed `fillNodes` from the unite tests, moving this logic to `prepareNodes` within `computeLayout`. There's a unit test that ensures backward compatibility.

I've also updated the README to provide a better description of the JS API.

### DIFF
--- a/src/Layout.js
+++ b/src/Layout.js
@@ -9,26 +9,23 @@
 
 var computeLayout = (function() {
 
-  // Ensures that all nodes have layoutm style and children properties. This simplifies
+  // Ensures that all nodes have layout, style and children properties. This simplifies
   // the layout algorithm in that it can assume a uniform structure
-  function prepareNodes(node) {
-    if (!node.layout) {
-      node.layout = {
+  function prepareNode(newNode) {
+    if (!newNode.layout) {
+      newNode.layout = {
         width: undefined,
         height: undefined,
         top: 0,
         left: 0
       };
     }
-    if (!node.style) {
-      node.style = {};
+    if (!newNode.style) {
+      newNode.style = {};
     }
-
-    if (!node.children || node.style.measure) {
-      node.children = [];
+    if (!newNode.children || newNode.style.measure) {
+      newNode.children = [];
     }
-
-    node.children.forEach(prepareNodes);
   }
 
   function capitalizeFirst(str) {
@@ -230,11 +227,13 @@ var computeLayout = (function() {
   var CSS_POSITION_RELATIVE = 'relative';
   var CSS_POSITION_ABSOLUTE = 'absolute';
 
-  function layoutNode(node, parentMaxWidth) {
+  return function layoutNode(node, parentMaxWidth) {
     var/*css_flex_direction_t*/ mainAxis = getFlexDirection(node);
     var/*css_flex_direction_t*/ crossAxis = mainAxis === CSS_FLEX_DIRECTION_ROW ?
       CSS_FLEX_DIRECTION_COLUMN :
       CSS_FLEX_DIRECTION_ROW;
+
+    prepareNode(node);
 
     // Handle width and height style attributes
     setDimensionFromStyle(node, mainAxis);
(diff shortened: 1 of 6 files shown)


ColinEberhardt:
Curse you transpiler! I'll get this fixed shortly ;-)

In the meantime, any review feedback regarding whether this PR is a suitable addition would be appreciated.

src/Layout.js
```diff
@@ -651,6 +673,11 @@ var computeLayout = (function() {
       }
     }
   };
+
+  return function performLayout(node, parentMaxWidth) {
+    prepareNodes(node);
```
vjeux: It's going to be slow to do that during each layout, but since we don't have caching yet on the JS version it's probably fine for now :)

src/__tests__/Layout-test.js
```diff
@@ -14,6 +14,22 @@ var texts = layoutTestUtils.texts;
 var textSizes = layoutTestUtils.textSizes;
 
 describe('Layout', function() {
+
+  it('should accomodate nodes that already have layout and children properties', function() {
```
vjeux: The way I would fix this is to add another describe, and when we do the transpiling of the tests, only transpile the tests that are in this describe.

ColinEberhardt:
OK, spotted the issue - didn't realise that the transpile step only took part of `Layout.js` when compiling to Java and C. Anyhow, I've re-worked the logic so that the C and Java versions of `prepareNode` are stubs as the logic is not needed with statically typed languages.

From my perspective, the important part of this change is that the JavaScript interface is now much simpler and the README example works! ...

```
var nodes = {
  style: {padding: 50}, children: [
    {style: {padding: 10, alignSelf: 'stretch'}}
  ]
};

computeLayout(nodes);
```

vjeux:
Hmm, I really don't like the fact that we're adding some no-ops in C and Java. For any real use case of the api we'll want to have the nodes shaped in a certain way.

We need to store some additional meta data info in order to do caching (which is not implemented in the js version): https://github.com/facebook/css-layout/blob/master/src/Layout.h#L69-L75

Also, the usual way you interact with the API is that you've got some node structure that contains a reference to the style object and a list of children. We have a special get_child function to do that: https://github.com/facebook/css-layout/blob/master/src/Layout.h#L114

So, my recommendation is to have two apis. One simple API as you suggest that's good for playing around and one API that's good when used in a real application with caching, the node indirection, a way to override measure... What do you think?

jimmyhmiller:
I have [a version](https://github.com/jimmyhmiller/css-layout/commit/1ae0b914536f387e395a979975803707b44a60e1) of this that doesn't require any changes to the java or C projects. I need to add a test in and change the readme. But it is working. 

I personally would love to see the javascript broken out a bit more. Separate the parts that need to be transpiled from the rest of it. But I think for right now, this other route might be something to look at.

vjeux:
@jimmyhmiller yeah, the JS part is in a pretty bad shape right now. I only hacked it together to get the C and Java ones that we're using in prod. Feel free to send pull requests to clean up the API :)

facebook-github-bot-6:
Thank you for reporting this issue and appreciate your patience. We've  notified the core team for an update on this issue. We're looking for a response within the next 30 days or the issue may be closed.

PR #41, opened 2015-02-04
Fixed preDefinedTextSizes for Chrome Canary 42.0.2295.0

coodoo:
Updated the default text size so that "RunLayoutTests.html" will pass.

However, "RunLayoutRandomTests.html" is still failing in various random layout tests.

### DIFF
--- a/src/Layout-test-utils.js
+++ b/src/Layout-test-utils.js
@@ -357,11 +357,11 @@ var layoutTestUtils = (function() {
   };
 
   var preDefinedTextSizes = {
-    smallWidth: 34.671875,
+    smallWidth: 34.65625,
     smallHeight: 16,
     bigWidth: 172.421875,
     bigHeight: 32,
-    bigMinWidth: 100.453125
+    bigMinWidth: 100.4375
   };
 
   var textSizes;


vjeux:
Grrr, I'm not sure what to do with this. The current version of Chrome doesn't have the same constants :(

There's a way to compute the constants on the fly: https://github.com/facebook/css-layout/blob/master/src/Layout-test-utils.js#L367-L378

But, we still need to propagate those constants to the C and Java tests. I'm unsure what's the best solution is

vjeux:
For the failing tests, there are two main causes which I think that they are bugs in the current flexbox implementations

https://github.com/facebook/css-layout/blob/master/src/__tests__/Layout-test.js#L1117-L1119
https://github.com/facebook/css-layout/blob/master/src/__tests__/Layout-test.js#L1176-L1178

The sad part is that it's the same behavior in both Chrome and Firefox. So it may be unlikely that they are going to fix it.

coodoo:
Yep, it only happens in Chrome Canary, which means it will eventually appear in standard version of Chrome, probably 2 months from now.

But it is indeed *not* a good idea to manually modify those numbers whenever changes happens.

Closing this for now.

PR #42, opened 2015-02-05
Improvement/add test runner

daviskoh:
Don't know if this is imposing, but I thought testing may be easier if there was a more manageable way to run them. Also, the added benefit is that when ready to test this in other browsers, karma launchers for them can be added to the package.json & karma.conf.js. What do you think?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ a.out
 /**/java/out/*
 /**/.idea/workspace.xml
 /lib/
+/node_modules/
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -1,0 +1,42 @@
+'use strict';
+
+module.exports = function (config) {
+    config.set({
+
+        // base path, that will be used to resolve files and exclude
+        basePath: 'src',
+
+        // frameworks to use
+        frameworks: ['jasmine'],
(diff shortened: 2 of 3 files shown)


daviskoh:
also, the specs in Layout-random-test.js seem to be failing:

fail

fail2

is this a known issue?

vjeux:
Nice, thanks for doing that :) Never used karma, this looks sweet. For the failing tests, there's a few behaviors which I believe are bugs in Chrome.

https://github.com/facebook/css-layout/blob/master/src/__tests__/Layout-test.js#L1117-L1119
https://github.com/facebook/css-layout/blob/master/src/__tests__/Layout-test.js#L1176-L1178

And, there's a few edge cases that are not properly supported yet unfortunately :(

daviskoh:
Ah I see...well np anyways. Glad to help. 

Question: what do you think about adding a linter (like jshint) to the npm pretest script so that the JS files get linted before tests get executed?

Also, good job on Reactjs Conf, the presentation on native looked super cool!

vjeux:
Happy to take it in, however i'm doing a lot of hacks that are probably not eslint friendly :p

PR #43, opened 2015-02-05
Fix firefox iframe timing issue in test runners, fixes #25

### LINKED ISSUE #25: Tests are failing on firefox 31.3.0
To reproduce: open RunLayoutTests.html in firefox.

all tests are failing in firefox

ryanseddon:
All browsers except chrome have timing issues when dynamically creating and injecting content inside an iframe, see [blog post](https://developer.zendesk.com/blog/2014/05/13/rendering-to-iframes-in-react/) if you're interested in why. Boils down to waiting for the `readyState` to be `complete` before appending to the DOM.

All tests pass in Firefox except 2, not sure how to fix these so any help would be good.

screen shot 2015-02-05 at 8 38 47 pm

I also changed the use of `innerText` to `textContent` as Firefox refused to render the CSS string and in Chrome it added a bunch of `<br>` tags and somehow the CSS parser is forgiving enough to still work!

screen shot 2015-02-05 at 8 44 35 pm

### DIFF
--- a/src/Layout-test-utils.js
+++ b/src/Layout-test-utils.js
@@ -44,47 +44,60 @@ var layoutTestUtils = (function() {
     };
   }
 
+  function renderIframe() {
+    var iframe = document.createElement('iframe');
+    document.body.appendChild(iframe);
+    return iframe;
+  }
+
   var cachedIframe;
   function getIframe() {
     if (cachedIframe) {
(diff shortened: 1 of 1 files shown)


vjeux:
Thanks! I couldn't figure out a reliable way to avoid the timing issues :)

PR #49, opened 2015-02-12
Add UMD

alexanderGugel:
Make it work...... everywhere!

Putting computeLayout in the IIFE right away broke the transpiler for me. Nevertheless, I think using an out of the box UMD pattern makes sense here. I don't see a reason why css-layout shouldn't be script loader agnostic.

### DIFF
--- a/src/Layout.js
+++ b/src/Layout.js
@@ -8,7 +8,6 @@
  */
 
 var computeLayout = (function() {
-
   function capitalizeFirst(str) {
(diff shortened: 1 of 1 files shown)


vjeux:
Nice! Thanks

petehunt:
Hmm, would it be better to leave this package as CommonJS and treat umd as a build target?

vjeux:
@petehunt do you know what's the standard procedure for this?

PR #50, opened 2015-02-16
Rework javascript api to match README

jimmyhmiller:
This is a fix for https://github.com/facebook/css-layout/issues/24

It doesn't require any changes to the transpiled code, but exposes a nice api for javascript. I did have to create a new file as the main for package.json. I'd love to work on simplifying the structure of the javascript in the future. Perhaps moving the layout algorithm into its own file. But those sorts of changes can be done later.

Let me know if you have any changes you'd like me to make.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "css-layout",
   "version": "0.0.1",
   "description": "Reimplementation of CSS layout using pure JavaScript",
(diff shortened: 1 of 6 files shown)


vjeux:
So good, thanks for doing that. Please keep the pull requests coming for improving the js api :)

PR #59, opened 2015-03-22
Normalized C and Java definition of isDimDefined.

freakboy3742:
The JavaScript implementation of isDimDefined contains a check to ensure
that the dimension value is positive; the C and Java versions did not
have this check. As a result, a negative style value for 'width' (such
as that used by the "should layout node with negative width" test) would
have different layout under the C/Java implementation to the JavaScript
implementation.

This was hidden because the C/Java transpilers filtered out any negative
instantiation values from the test suite. In effect, the negative value
tests weren't running on the C/Java implementation.

This patch removes the negative value filter from the transpiler, and
makes the isDimDefined definition consistent between the three
implementations.

### DIFF
--- a/src/Layout.c
+++ b/src/Layout.c
@@ -287,7 +287,8 @@ static float getDimWithMargin(css_node_t *node, css_flex_direction_t axis) {
 }
 
 static bool isDimDefined(css_node_t *node, css_flex_direction_t axis) {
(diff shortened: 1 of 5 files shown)


vjeux:
Good catch! Thanks for sending this pull request :)

PR #60, opened 2015-03-23
Make Java measure thread-safe.

tcmulcahy:
The measure method in java previously referenced a global MEASURE_OUTPUT variable. This means that css layout could not be done safely from multiple threads. The naive solution would be to create a MeasureOutput on each measure invocation, but since Android is sensitive to allocation and garbage collection, we instead force a MeasureOutput instance to be passed in to calculateLayout. Since it would be strange and non-intuitive to pass MeasureOutput to calculateLayout, this implementation detail is hidden from clients by wrapping the MeasureOutput in a CSSLayoutContext object. Clients are free to re-use the same CSSLayoutContext object, provided that they ensure it is not used concurrently from multiple threads.

### DIFF
--- a/src/java/src/com/facebook/csslayout/CSSLayoutContext.java
+++ b/src/java/src/com/facebook/csslayout/CSSLayoutContext.java
@@ -1,0 +1,20 @@
+/**
+ * Copyright (c) 2014, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+package com.facebook.csslayout;
+
+/**
+ * A context for holding values local to a given instance of layout computation.
+ *
+ * This is necessary for making layout thread-safe. A separate instance should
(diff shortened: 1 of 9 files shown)


vjeux:
@kmagiera or @astreet can you review this?

src/java/src/com/facebook/csslayout/CSSLayoutContext.java
```diff
@@ -0,0 +1,13 @@
...
+ * Copyright (c) 2014, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+package com.facebook.csslayout;
+
+public class CSSLayoutContext {
```
kmagiera: Please add short class level comment explaining what is this for

kmagiera:
Thanks! Looks good, please just add a class level comment to CSSLayoutContext that will explain the purpose of this class and that a separate instance should be used when calculateLayout is called concurrently for different node hierarchies. Once it's done, ask @vjeux to merge this

tcmulcahy:
Okay, I've added the javadoc, and also fixed a long line.

PR #66, opened 2015-04-01
Fix getIframe test helper

kassens:
Chrome seems to always be ready on the first iteration, so this doesn't affect
Chrome, but on Firefox, the `iframe` is undefined in the recursive call.

This makes most of the test in `RunLayoutTests.html` pass in Firefox. The only
failing test is the one checking font sizes.


vjeux:
Sweet thanks!

PR #68, opened 2015-04-07
Ensure that flex children adopt their parent's cross-axis min dimension.

freakboy3742:
Corrects one of the two tests added in b912acf8e3.


vjeux:
Thanks! Expect more breaking tests soon :p

PR #70, opened 2015-04-24
Remove unused code pointed by Xcode

tadeuzagallo:
Xcode analyze tool pointed out this dead code on react-native, I double checked and everything is still alright after removing it... Was it supposed to be used or can it really just be removed?

### DIFF
--- a/src/Layout.c
+++ b/src/Layout.c
@@ -693,19 +693,6 @@ static void layoutNodeImpl(css_node_t *node, float parentMaxWidth) {
       }
     }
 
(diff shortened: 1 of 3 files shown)


vjeux:
If the tests are still passing this is good to me! Can you run make to also remove the obj-c and java parts? Thx

tadeuzagallo:
I've updated the C and Java files. the `npm` and `make` tests are ok. If I run the `js` tests through `html` files I get 19 failures, but I checked and it's pre-existent. I'll try to fix then on a next PR.

PR #72, opened 2015-04-26
Transpile to update the changes of consts omitted in #69

Kroisse:
#69 includes [the changes of `preDefinedTextSizes`][1] but the corresponding transpiled results are omitted. So currently they have different text size constants that the JavaScript test case and C and Java's one.

[1]: https://github.com/facebook/css-layout/blob/1cbb5eef26478bc79ae4f823cd15e537f90e26c0/src/Layout-test-utils.js#L362-L368


vjeux:
Thanks!

PR #75, opened 2015-05-01
Revert "Ensure that flex children adopt their parent's cross-axis min di...

arasmussen:
...mension."

This reverts commit b8316413b310643ea6555a015b3903f4fde1104e.

This was causing issues for existing components built with React
Native.. we need to investigate more in the future.

cc @freakboy3742


freakboy3742:
Are you able to elaborate on the "issues"? Can you provide a test case (or reproduction instructions)?

PR #76, opened 2015-05-05
Add a comment telling developer how to update react-native-github

arasmussen:


spicyj:
Seems like a funny thing to put here. Any reason we can't make a `@generated` signer script that pulls from github, adds a note referring to itself, and signs the files?

PR #77, opened 2015-05-05
[Layout.c/h]: Minor modifications so that the Microsoft compiler can compile the file as C++ directly ;

prenaux:
Visual C++ doesn't support C99, the simplest way is to compile the C as C++ (by changing the extension or with the /Tp extension on the command line).

These changes account for the fact that its C++ instead of C (one explicit cast), and for the lib C differences.


vjeux:
Nice tricks

PR #80, opened 2015-05-08

lukasreichart:


lukasreichart:
This is a little embarrassing: only changed 3 lines of code and it's failing: think I just copied the code from C to javascript ... :/
I'll create another pull request.

PR #81, opened 2015-05-08

lukasreichart:


vjeux:
Can you run `make` in order to generate the C and Java versions as well?

I have a feeling that what you are working on is going to be glorious! Are you going to let you embed <Image> or arbitrary elements inside of a text node? If yes, <3<3<3

lukasreichart:
@vjeux I have created another [pull](https://github.com/facebook/css-layout/pull/84) request.

PR #82, opened 2015-05-08
Support RTL direction and ROW_REVERSE & COLUMN_REVERSE flex direction

lucasr:
This PR adds support for a direction style property (equivalent to Chrome's `direction` CSS property) that allows you to change the row flow for flexbox allows. By default, all nodes inherit the direction from their parents. If the parent direction is not explicitly defined, left-to-right is used. This means that, by default, nodes from a subtree will share the same direction defined by the root. You can also have node trees with mixed directions, as covered by one of the new tests.

Given that the work to implement RTL support is analogous to `ROW_REVERSE` and `COLUMN_REVERSE`, I went ahead and implemented them as well.

These new features require a small API break in the `layoutNode()` for both C and Java as well as a few new types for the new direction style property and the new flex direction values.

I added a bunch of analogous tests for the reversed layout cases and a few new tests to verify that the new features work correctly.

I split the changes into separate commits to make them a bit easier to review: one with the Javascript changes and another with the updates to the transpiled versions. I'll squash them together before merging.

### DIFF
--- a/src/Layout.js
+++ b/src/Layout.js
@@ -11,8 +11,14 @@ var computeLayout = (function() {
 
   var CSS_UNDEFINED;
 
+  var CSS_DIRECTION_INHERIT = 'inherit';
+  var CSS_DIRECTION_LTR = 'ltr';
+  var CSS_DIRECTION_RTL = 'rtl';
+
   var CSS_FLEX_DIRECTION_ROW = 'row';
+  var CSS_FLEX_DIRECTION_ROW_REVERSE = 'row-reverse';
   var CSS_FLEX_DIRECTION_COLUMN = 'column';
+  var CSS_FLEX_DIRECTION_COLUMN_REVERSE = 'column-reverse';
 
   // var CSS_JUSTIFY_FLEX_START = 'flex-start';
   var CSS_JUSTIFY_CENTER = 'center';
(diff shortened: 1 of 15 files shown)


src/Layout.js
```diff
@@ -11,8 +11,14 @@ var computeLayout = (function() {
 
   var CSS_UNDEFINED;
 
+  var CSS_DIRECTION_INHERIT = 'inherit';
```
vjeux: Please remove inherit

vjeux:
Grrr, this error is super weird

```
FAIL should layout text with flexDirection row in rtl
Input:    {flexDirection: 'column', alignItems: 'stretch', flex: 0, },
Output:   {layout: {width: 200, height: 58, top: 0, left: 0}, },
Expected: {layout: {width: 200, height: 58, top: 0, left: 0}, },
```

vjeux:
Actually it segfaults

```
FAIL should layout text with flexDirection row in rtl
Input:    {flexDirection: 'column', alignItems: 'stretch', flex: 0, },
Output:   {layout: {width: 200, height: 58, top: 0, left: 0}, },
Expected: {layout: {width: 200, height: 58, top: 0, left: 0}, },
/bin/sh: line 1: 34119 Segmentation fault: 11  ./a.out
make: *** [c_test] Error 139
```

vjeux:
Okay, I just pushed a diff that shows all the children.

```
Input:    {flexDirection: 'column', alignItems: 'stretch', flex: 0, children: [
  {flexDirection: 'row', alignItems: 'stretch', flex: 0, width: 200, children: [
    {flexDirection: 'column', alignItems: 'stretch', flex: 0, margin: 20, padding: 20, },
  ]},
]},
Output:   {layout: {width: 200, height: 58, top: 0, left: 0}, children: [
  {layout: {width: 200, height: 58, top: 0, left: 0}, children: [
    {layout: {width: 172.422, height: 18, top: 20, left: 7.57812}, }, // <--- 7.57812
  ]},
]},
Expected: {layout: {width: 200, height: 58, top: 0, left: 0}, children: [
  {layout: {width: 200, height: 58, top: 0, left: 0}, children: [
    {layout: {width: 172.422, height: 18, top: 20, left: 15}, }, // <--- 15
  ]},
]},
```

PR #84, opened 2015-05-10

lukasreichart:

### DIFF
--- a/src/Layout.c
+++ b/src/Layout.c
@@ -431,7 +431,9 @@ static void layoutNodeImpl(css_node_t *node, float parentMaxWidth) {
           getPaddingAndBorderAxis(node, CSS_FLEX_DIRECTION_COLUMN);
       }
     }
-    return;
+    if (node->children_count == 0) {
+      return;
+    }
   }
 
   int i;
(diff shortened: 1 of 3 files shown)


vjeux:
Thanks, that makes sense! I can't wait to see your TextView code. I've always wanted to be able to put arbitrary elements there but couldn't find time to implement it!

paramaggarwal:
@lukasreichart Just like you use this for a `<TextView>`, I use it for `<Image>`. With `measure` I'm able to define an aspect ratio for the image and then have it take the entire width of the device - taking the necessary height as required. Super useful!

Not sure how to add support to also do this other way round - calculate height given a width.

lukasreichart:
@paramaggarwal sounds great and I'd love to try it.  :+1: Have you published a npm module or created a PR for react-native?

paramaggarwal:
@lukasreichart Unfortunately your change here hasn't ended up in RN yet... Open PR: https://github.com/facebook/react-native/pull/1776

I have a generic implementation for all RCTView subclasses here: https://github.com/facebook/react-native/pull/1538 - simply pass an `aspectRatio` prop to your views. So you could make _any_ view behave like a rigid boxed rectangle and all the children would layout normally.

The PR is blocked at the moment as `measure` can only calculate a height given a width and not vice-versa. Any thoughts on how to make `measure` more generic to be able to measure both ways?

PR #86, opened 2015-05-12
Implement (padding|border|margin)-(start|end) support

lucasr:
These changes exposed a bug in my previous diff which is fixed now and covered by the new tests i.e. the trailing position has to be set on both main and cross axis.


vjeux:
Are you going to implement position: absolute; start/end as well? It's not in the spec but without it you can't really implement proper rtl support everywhere

lucasr:
@vjeux I can do it, just want to make sure we agree on this direction before going ahead.

vjeux:
Can you rebase, it doesn't merge cleanly anymore, sorry :(

PR #88, opened 2015-05-14
Add getters for CSSNode's style width and height

lucasr:


vjeux:
@foghina can you review?

obishop:
looks good on the java side.

foghina:
this is how i feel

PR #89, opened 2015-05-15
[java] simplify Spacing, reduce allocations

foghina:
* always resolve spacing in `get()`, since it's not that expensive and allows us to not allocate a result array
* allocate for default spacing on demand

cc @lucasr @andreicoman11

### DIFF
--- a/src/java/src/com/facebook/csslayout/Spacing.java
+++ b/src/java/src/com/facebook/csslayout/Spacing.java
@@ -8,6 +8,8 @@
  */
 package com.facebook.csslayout;
 
+import javax.annotation.Nullable;
+
 /**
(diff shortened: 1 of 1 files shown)


lucasr:
Looks good to me. What was this `mDirty` about? Do we have tests for `Spacing`? Maybe add some?

foghina:
@lucasr `mDirty` was used in `ensureResult()` to determine whether recalculating the result array is necessary. No explicit `Spacing` tests, but the coverage is pretty good from the many many layout tests that implicitly use `Spacing` for padding and margin. The layout tests already caught a couple of issues while writing this patch and the previous one (#87).

lucasr:
:+1:

PR #90, opened 2015-05-15
[java] use getRaw to test if START/END were set

foghina:
This is causing issues when setting, for example, `margin: 10, marginLeft: 20` as `get(START)` will return `10` and take precedence over `get(LEFT)` which would be `20`.


vjeux:
@lucasr can you review?

lucasr:
:+1:

Issue #64, opened 2015-03-26
Style checks force code to go polymorphic or worse


esprehn:
width/height/top/left use isDimDefined and isPosDefined which checks for undefined. margin and padding properties use the "in" operator which isn't fast and also means you need to totally omit those properties from your object if you want them to have "no value" for layout.

ex.
```js
if (pos in node.style) {
    return node.style[pos];
}
return 0;
```

should just be return node.style[pos] || 0;

For getSpacing it should do:

```js
return node.style[key] || node.style[type + suffix] || 0;
```

This means your style object can have that property all the time and just use the value undefined to get the default. The current code forces you to have many different "types" of Style objects which makes v8 and other JS engines go slow.

vjeux:
Totally! The js version has not been written with performance in mind. If you send a pull request i'd gladly accept it :)

JonSonesen:
Hey!
Noticed that this has been sitting for a while and would like to go ahead and implement these optimizations. I can get started right away but want to make sure that @esprehn didn't already tackle this so as to avoid redundancy. I am a CS student who couldn't find an internship this summer so I saved up money to take the summer off and do as much work on open source projects as I can handle. :)

vjeux:
Please send out pull requests :)

JonSonesen:
:+1:

jaredly:
@JonathonSonesen were you going to PR with those fixes?

JonathonSonesen:
Hi Jared, after running the tests I found that there were no notable
performance improvements from my changes so I closed the pull request also,
I am not very experienced so I wasn't sure what I had done was even
correct. I started a summer class and its an accelerated course so my time
has been filled with calculus 2 for the last 7 weeks in addition to doing
udemy courses for js and python. I can definitely reopen the pull request
if you'd like to see what I added.  My final is next Wed. so I can take
another look at it after I wrap that up as well.

jaredly:
Ok cool :) good luck with finals

JonathonSonesen:
Thanks man! I plan to slay them, with the sweet knowledge blades I've been forging ;)

emilsjolander:
Js version is gone, closing this.

PR #94, opened 2015-06-29
fix for #64

JonSonesen:
This is my attempt to implement the optimizations mentioned in the issue. I assumed it was also safe to remove the 'in' check from 'getPositiveSpacing' as well because if it is null then will not be >= 0 I would be happy to make any requested changes if this isn't quite right.
Thanks!

### DIFF
--- a/src/Layout.js
+++ b/src/Layout.js
@@ -68,14 +68,8 @@ var computeLayout = (function() {
       var location = locations[i];
 
       var key = type + capitalizeFirst(location) + suffix;
-      if (key in node.style) {
-        return node.style[key];
-      }
-
-      key = type + suffix;
-      if (key in node.style) {
-        return node.style[key];
-      }
+  
+        return node.style[key] || node.style[type + key + suffix];
     }
 
     return 0;
@@ -121,12 +115,12 @@ var computeLayout = (function() {
       var location = locations[i];
 
       var key = type + capitalizeFirst(location) + suffix;
-      if (key in node.style && node.style[key] >= 0) {
+      if (node.style[key] >= 0) {
         return node.style[key];
       }
 
       key = type + suffix;
-      if (key in node.style && node.style[key] >= 0) {
+      if (node.style[key] >= 0) {
         return node.style[key];
       }
     }


src/Layout.js
```diff
@@ -68,14 +68,8 @@ var computeLayout = (function() {
...
-      if (key in node.style) {
-        return node.style[key];
-      }
-
-      key = type + suffix;
-      if (key in node.style) {
-        return node.style[key];
-      }
+  
+        return node.style[key] || node.style[type + key + suffix];
```
vjeux: this is not safe :(

```javascript
var node = { style: { marginTop: 0, margin: 10 } };
```

```javascript
// before
if ('marginTop' in node.style) {
  return node.style['marginTop'];
}
// returns 0
```

```javascript
// after
return node.style['marginTop'] || node.style['margin'];
// returns 10
```

vjeux: A safe way to do it is by writing `node.style['marginTop'] !== undefined`. But, in order to make sure that it has performance improvement, we need to benchmark it.

Can you use the code that generates a random test to generate a big random tree and then time the layout algorithm before and after? Otherwise we can't know if this is a performance optimization or not.

JonSonesen:
Hey,
I will gladly make this safer then time the tests and  give you the output. I should have thought out my changes a bit more thoroughly, sorry about that!

JonSonesen:
Also would making it  look like
 <code>return node.style['marginTop'] || node.style['margin'] || 0;</code> 
work if I excluded the return 0 on exiting the loop? Otherwise I can add check for being undefined. 
Also for the tests do I use the RunLayoutRandomTests.html to do this?  Or the tests that are in the __tests__ dir?

JonSonesen:
Also, opening RunLayoutRandomTests.html before my changes finishes in about 2.2 seconds on chrome and after it finishes in about 1 second

JonSonesen:
Hey so after working on this for a while yesterday I found that there wasn't much of a performance improvement on my machine, and that the changes caused a slight increase in the number of failures on the layout tests. Thanks for taking a look at what I had. I will close the PR.

PR #95, opened 2015-07-29
Lazy init children array in CSSNode, update removeChildAt to return removed child

astreet:


spicyj:
Neat. Know if this improves perf?

PR #99, opened 2015-08-06
Flex bug

jaredly:
Here's a bug in the flexbox calculation

```js
  <View>
    <View>
      <Text style={{backgroundColor: 'red'}}>Awesome</Text>
    </View>
    <View style={{width: 400, height: 20}}/>
  </View>
```

image

I added a failing test, but I don't immediately know where I would go to fix it.

### DIFF
--- a/src/__tests__/Layout-test.js
+++ b/src/__tests__/Layout-test.js
@@ -2307,6 +2307,21 @@ describe('Layout', function() {
       ]}
     );
   });
+
+  xit('should stretch a nested child', function() {
+    testLayout(
+      {children: [
+        {children: [{}]},
+        {style: {width: 40}}
+      ]},
+      {width: 40, height: 0, top: 0, left: 0, children: [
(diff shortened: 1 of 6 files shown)


src/__tests__/Layout-test.js
```diff
@@ -2397,3 +2397,20 @@ describe('Layout alignContent', function() {
   testAlignContent('flex-end', 'flex-end');
   testAlignContent('flex-end', 'stretch');
 });
+
+describe('nested flex', function() {
+  it('should stretch a nested child', function() {
```
vjeux: can you use xit instead of doing a second describe? This is how I put tests that aren't passing but should

jaredly:
Ok. @vjeux do you know where I should look to fix it?

vjeux:
The way I usually debug is to just run that test and put console.log everywhere in the big function and try hacking around. When I have a fix, I run the other tests and try to make them work as well

PR #109, opened 2015-08-13
Windows build, package C, and updated README

prenaux:
All tested on Windows and OSX.

The package-c task creates a unified css-layout.h file which is a single include containing the header and the implementation.

Example usage:
```c
// some_file.c
#define CSS_LAYOUT_IMPLEMENTATION
#include <css-layout.h> // prototypes and implementation
...
```

```c
// some_other_file.c
#include <css-layout.h> // prototypes only
```

This in my experience is the simplest way to distribute a lightweight C lib, its inspired by https://github.com/nothings/stb.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -19,11 +19,6 @@ module.exports = function(grunt) {
     javaTestFiles: 'org.junit.runner.JUnitCore com.facebook.csslayout.LayoutEngineTest com.facebook.csslayout.LayoutCachingTest com.facebook.csslayout.CSSNodeTest'
   };
 
-  // Create the dist folder if it doesn't exist. It is deleted by the 'clean' task.
-  if (!fs.existsSync(config.distFolder)){
-    fs.mkdirSync(config.distFolder);
-  }
-
   // C compilation configuration
   if (isWindows) {
     // Windows build, assumes cl is in the path (see https://msdn.microsoft.com/en-us/library/f2ccy3wt.aspx).
@@ -41,8 +36,15 @@ module.exports = function(grunt) {
   }
 
   grunt.initConfig({
-
     config: config,
+
+    mkdir: {
+      dist: {
+        options: {
+          create: ['<%= config.distFolder %>']
+        },
+      },
+    },
 
     clean: {
       dist: ['<%= config.distFolder %>'],
@@ -163,13 +165,13 @@ module.exports = function(grunt) {
   grunt.registerTask('test-javascript', ['eslint', 'karma']);
 
   // Packages the JavaScript as a single UMD module and minifies
-  grunt.registerTask('package-javascript', ['includereplace', 'uglify']);
+  grunt.registerTask('package-javascript', ['mkdir:dist', 'includereplace', 'uglify']);
 
   // Packages the Java as a JAR
-  grunt.registerTask('package-java', ['shell:javaPackage']);
+  grunt.registerTask('package-java', ['mkdir:dist', 'shell:javaPackage']);
 
   // Packages the C code as a single header
-  grunt.registerTask('package-c', ['concat']);
+  grunt.registerTask('package-c', ['mkdir:dist', 'concat']);
 
   // Default build, performs the full works!
   grunt.registerTask('build', ['test-javascript', 'transpile', 'clean:dist', 'package-javascript', 'package-java', 'package-c']);
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -6,10 +6,10 @@ module.exports = function(grunt) {
   var isWindows = /^win/.test(process.platform);
 
   require('load-grunt-tasks')(grunt);
-  grunt.loadNpmTasks('grunt-contrib-concat');
 
   // config
   var config = {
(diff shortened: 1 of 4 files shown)


Gruntfile.js
```diff
@@ -1,28 +1,53 @@
...
     srcFolder: 'src',
     testFolder: 'src/__tests__',
-    cTestFiles: 'src/__tests__/Layout-test.c src/Layout.c src/Layout-test-utils.c',
-    cTestOutput: 'c_test',
     javaLibFolder: 'src/java/lib',
     javaSource: 'src/java/tests/com/facebook/csslayout/*.java',
     javaTestFiles: 'org.junit.runner.JUnitCore com.facebook.csslayout.LayoutEngineTest com.facebook.csslayout.LayoutCachingTest com.facebook.csslayout.CSSNodeTest'
   };
 
+  // Create the dist folder if it doesn't exist. It is deleted by the 'clean' task.
```
ColinEberhardt: I'd prefer to see the clean task updated to `dist/**` - this feels a bit like a side-effect.

Gruntfile.js
```diff
@@ -1,28 +1,53 @@
...
+  if (!fs.existsSync(config.distFolder)){
+    fs.mkdirSync(config.distFolder);
+  }
+
+  // C compilation configuration
+  var cTestClean, cTestCompile, cTestExecute;
```
ColinEberhardt: rather than defining new variables, can these be added to the 'config' object?

Gruntfile.js
```diff
@@ -78,18 +103,56 @@ module.exports = function(grunt) {
...
-        command: 'gcc -std=c99 -Werror -Wno-padded <%= config.cTestFiles %> -lm -o "./<%= config.cTestOutput %>"'
+        command: cTestCompile
       },
       cTestExecute: {
-        command: './<%= config.cTestOutput %>'
+        command: cTestExecute
       },
       javaCompile: {
-        command: 'javac -cp <%= config.javaLibFolder %>/junit4.jar:<%= config.javaLibFolder %>/jsr305.jar:<%= config.javaLibFolder %>/infer-annotations-1.4.jar -sourcepath ./src/java/src:./src/java/tests <%= config.javaSource %>'
+        command: 'javac -cp <%= config.javaLibFolder %>/junit4.jar' +
```
ColinEberhardt: rather than mixing templates strings and concatenation, can you place `path.delimiter`onto the config object and just use templating?

README.md
```diff
@@ -43,7 +43,7 @@ padding, paddingLeft, paddingRight, paddingTop, paddingBottom | positive number
 borderWidth, borderLeftWidth, borderRightWidth, borderTopWidth, borderBottomWidth | positive number
 flexDirection | 'column', 'row'
 justifyContent | 'flex-start', 'center', 'flex-end', 'space-between', 'space-around'
-alignItems, alignSelf | 'flex-start', 'center', 'flex-end', 'stretch'
+alignItems, alignSelf, alignContent | 'flex-start', 'center', 'flex-end', 'stretch'
```
ColinEberhardt: well spotted!

Gruntfile.js
```diff
@@ -1,28 +1,53 @@
 'use strict';
 
 module.exports = function(grunt) {
+  var fs = require('fs');
+  var path = require('path');
+  var isWindows = /^win/.test(process.platform);
 
   require('load-grunt-tasks')(grunt);
+  grunt.loadNpmTasks('grunt-contrib-concat');
```
ColinEberhardt: this is unnecessary, the `load-grunt-tasks` should find and load this task.

ColinEberhardt:
Great job - a few comments on the Gruntfile, but otherwise, looks good to me.

It would be good to get some input from someone else on the C package step.

Gruntfile.js
prenaux: Its a side effect, but, since the dist folder can also be removed "manually" by someone that does the cleanup by hand, and the fact that the build fails if the dist folder doesn't exist I opted to make sure the folder exists.

I'm not sure whether grunt has some native task to create directories, which would probably be nicer than this.

Gruntfile.js
prenaux: yup

Gruntfile.js
prenaux: ok

Gruntfile.js
ColinEberhardt: There's a task for that: https://github.com/rubenv/grunt-mkdir ;-)

prenaux: Yes, I feel bad adding a module to replace two lines of code though :package: 

Anyhow, trying to use it, I've added:

prenaux: ok, nevermind I forgot to npm install grunt-mkdir ^^

prenaux:
I've updated the code based on your comments.

vjeux:
@ColinEberhardt feel free to merge it in when it is to your liking, you do not need to wait for me :)

ColinEberhardt:
Thanks @prenaux for turning this around so quickly.

PR #110, opened 2015-08-14
fix broken JSFiddle link in README.md

shamrin:
Original JSFiddle used `http://fb.me/react-js-fiddle-integration.js`, which returns 5xx server error.
Updated JSFiddle uses `https://facebook.github.io/react/js/jsfiddle-integration.js`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ position | 'relative', 'absolute'
 
 Default values
 --------------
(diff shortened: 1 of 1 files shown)


spicyj:
Fixed the fb.me link.

shamrin:
Thank you! It works now. (I had to force reload JSFiddle page though.)

PR #117, opened 2015-08-28
Watch for changes and run tests during development using `grunt watch`.

paramaggarwal:
In development, watch for changes to `src/Layout.js` and keep running tests whenever the file changes. Helps to ensure that a breaking change is not introduced while developing.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -149,7 +149,12 @@ module.exports = function(grunt) {
       javaPackage: {
         command: 'jar cf <%= config.distFolder %>/<%= config.libName %>.jar <%= config.javaSource %>'
       }
-    }
+    },
+
+    watch: {
+      files: ['src/Layout.js'],
+      tasks: ['ci'],
+    },
   });
 
   // Compiles and runs the Java tests
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "description": "Reimplementation of CSS layout using pure JavaScript",
   "main": "dist/css-layout.js",
   "scripts": {
+    "watch": "grunt watch",
     "test": "grunt ci",
     "postversion": "git push --tags upstream HEAD:master"
   },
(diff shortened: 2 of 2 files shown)


paramaggarwal:
That was fast! Thanks @vjeux. (Please also bless https://github.com/facebook/react-native/pull/1776! :stuck_out_tongue:)

vjeux:
Yeah so sorry about this, I want to do a proper review of aspectRatio and merge it but I kept putting it to the next day :(

paramaggarwal:
https://github.com/facebook/react-native/pull/1776 is the tiny css-layout change that was needed to allow the children of a `measure` function based node to also be laid out. 

Regarding the `aspectRatio` PR, I am looking at how to make aspectRatio work horizontally too. Currently because `measure` only works one way (given width, it calculates height) the `aspectRatio` also works in just one direction. To be able to use `aspectRatio` such that given a height, it would calculate the width to maintain ratio, we need to make `measure` more generic.

Any idea how to make `measure` take both width and height, so that it can work both ways?

vjeux:
Oh interesting. Yeah would be nice. You should look at sizeThatFits and layoutSubviews in iOS. It has proven to be a good abstraction to implement "hard" layouts. https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIView_Class/#//apple_ref/occ/instm/UIView/sizeThatFits:

ColinEberhardt:
Thanks for the contribution to the build :-) 

Quick question - is there any reason the watch runs the ci task rather than the build task? (See the comments that describe the difference)

paramaggarwal:
@ColinEberhardt The `build` task would cause chrome to keep opening again and again. This is more like continuous integration if you think about it.

paramaggarwal:
@vjeux The `sizeThatFits` API looks nice. It takes a `CGSize` as input - pretty much what we need for our generic `measure` function in css-layout. Currently `measure` only takes `width` - hence the limitation.

Just need to modify css-layout to let `measure` take a width _and_ a height and then set layout properties based on the returned value.

ColinEberhardt:
@paramaggarwal fair enough, I guess that would be quite disruptive.

PR #118, opened 2015-08-28
Add support for both width and height in measure

paramaggarwal:
The idea is to make `measure` more generic. Like `sizeThatFits:` from iOS.

I'm working on adding a few tests too to go with this. But the Chrome tests interfere - as cannot be tested in Chrome. Also having an issue in the Java tests. The C tests work just fine.


dist/css-layout.js
```diff
@@ -434,30 +435,41 @@ var computeLayout = (function() {
...
         width = node.style.width;
+      } if (isDimDefined(node, resolvedColumnAxis)) {
+        height = node.style.height;
       } else if (!isUndefined(node.layout[dim[resolvedRowAxis]])) {
         width = node.layout[dim[resolvedRowAxis]];
+      } else if (!isUndefined(node.layout[dim[resolvedColumnAxis]])) {
+        height = node.layout[dim[resolvedColumnAxis]];
       } else {
-        width = parentMaxWidth -
+        width = parentMaxWidth -
```
vjeux: Can you setup your editor to trim trailing whitespace? :)

dist/css-layout.h
```diff
@@ -1,7 +1,7 @@
-/*
- * #define CSS_LAYOUT_IMPLEMENTATION
- * before you include this file in *one* C or C++ file to create the implementation.
- */
+/*
+ * #define CSS_LAYOUT_IMPLEMENTATION
```
vjeux: do you know what changed?

dist/css-layout.h
```diff
@@ -724,30 +725,41 @@ static void layoutNodeImpl(css_node_t *node, float parentMaxWidth, css_direction
 
   if (isMeasureDefined(node)) {
     float width = CSS_UNDEFINED;
+    float height = CSS_UNDEFINED;
+    float parentMaxHeight = parentMaxWidth;
```
paramaggarwal: The idea is to use this for images also, so that one could create a custom component with support for fixed aspect ratio. So given a width, the height would be calculated to maintain the aspect ratio of the image.

Currently this works only to calculate the height, given a width. I'm trying to make it work the other way also, that given a height, it should calculate width.

Think of it more like `- (CGSize *)sizeThatFits:(CGSize)size` than `measure`. Meaning it allows custom components to override layout.

dist/css-layout.h
paramaggarwal: haha, nope

dist/css-layout.js
paramaggarwal: yes, will fix

paramaggarwal:
@ColinEberhardt I would like to request some help/suggestions/direction here... What I am trying to do is to make `measure` more like `sizeThatFitsSize:` from iOS. Currently `measure` only takes a `width` and gives `height` as output. Hence it can only be used for text.

I would like to use it for images where I would like to maintain an aspect ratio. The way I go about this is to make a custom image module that implements `measure`. Currently this works in one direction - given a width, the height is maintained as per aspect ratio. But I would like this to work the other way around also.

I started work on this PR towards that direction. But I don't understand the code fully to be able to submit a full-fledged PR with corresponding tests... This is because the new `sizeThatFits` cannot be tested with the browser testing that is there here. It needs the more basic given this input, expect this output kind of testing.

Please help if you would agree that a more generic `sizeThatFits:` kind of API would be useful for `css-layout`. (ref: [Apple Docs](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIView_Class/#//apple_ref/occ/instm/UIView/sizeThatFits:))

ColinEberhardt:
Hi @paramaggarwal sorry but i doubt I will be much help here, I have mostly been responsible for tidying up the JS API and distribution. I don't know enough about the C & Java use cases to be able to determine whether this is a good idea or not.

However, one thought springs to mind, wouldn't a better solution be to expand the CSS vocabulary of this project so that you can create a fixed aspect ratio element via CSS?

paramaggarwal:
@ColinEberhardt Didn't want to extend the CSS vocabulary as it breaks the assumption that the API to this library is CSS itself... This `measure` based approach allows a generic way to override layout behaviour on a particular node in the layout hierarchy.

paramaggarwal:
Fixed in https://github.com/facebook/css-layout/pull/154.

PR #121, opened 2015-09-07
Faster flexbox

lucasr:
This branch has a series of changes that makes css-layout around 60% faster in my preliminary benchmarks on Android. 

Some of these commits also benefit the JS and C implementations as they make the layout algorithm more efficient in some setups.

Here's an overview of the changes:
* Avoid a large amount of method invocations in the layout code by moving the generated Java code closer to how the JS and C implementations work i.e. pre-computed arrays with the decisions for main/cross axis. As a bonus, the Java transpiler a now simpler because the code is mostly analogous to the one in C.
* Inline immutable values so avoid method invocations in some of the loops in the layout code.
* Skip/merge some of the loops to avoid unnecessary work in some setups e.g. no need to traverse absolute children if there are none.
* Reduce the search range for flexible children i.e. quicker loops when flexible children are present.

There's definitely more work to be done in terms of adding faster paths for the most common layout setups. For instance, I think we can do a much better job when a node is a simple vertical/horizontal stack of elements. But this is a good start ;-)

In case you're curious, my benchmarks have two types of layouts: a simple vertical stack of 500 children and a more real world "header"-like layout with avatar+title+menu (one flex child surrounded by non-flex ones). Here are the numbers:

* Vertical stack layout
  * Before: ~20ms
  * After: ~7ms
* Header layout
  * Before: 160μs
  * After: 60μs


vjeux:
This is pretty awesome that you've been able to do all this without changing any test in C and JS :)

cc @kmagiera for reviewing the Java changes

vjeux:
Looks really good to me!

src/Layout.js
```diff
@@ -528,16 +504,62 @@ var computeLayout = (function() {
...
+            !isDimDefined(child, crossAxis)) {
+          child.layout[dim[crossAxis]] = fmaxf(
+            boundAxis(child, crossAxis, node.layout[dim[crossAxis]] -
+              paddingAndBorderAxisCross -
+              getMarginAxis(child, crossAxis)),
+            // You never want to go smaller than padding
+            getPaddingAndBorderAxis(child, crossAxis)
+          );
+        } else if (getPositionType(child) === CSS_POSITION_ABSOLUTE) {
+          absoluteChildrenCount++;
```
kastiglione: Re 41591257: Would there be any performance gain if this were to add `child` (or its index) into an array?…

src/Layout.js
```diff
@@ -933,37 +960,38 @@ var computeLayout = (function() {
...
-        }
-        for (ii = 0; ii < 2; ii++) {
-          axis = (ii !== 0) ? CSS_FLEX_DIRECTION_ROW : CSS_FLEX_DIRECTION_COLUMN;
-          if (isPosDefined(child, trailing[axis]) &&
-              !isPosDefined(child, leading[axis])) {
-            child.layout[leading[axis]] =
-              node.layout[dim[axis]] -
-              child.layout[dim[axis]] -
-              getPosition(child, trailing[axis]);
+    if (absoluteChildrenCount != 0) {
```
kastiglione: Re 41591257: With a collected array of absolute children, this optimization would apply not just for N = 0, but also for N = 1, 2.

src/Layout.js
lucasr: @kastiglione Generally, we try very hard to avoid allocations during layout as 1) allocations are not free (on Android, at least) 2) More allocations -> more likely GC calls and main thread stalls. It's a trade-off: more stable memory behaviour vs computational cost.

src/Layout.js
lucasr: Yeah, there a couple of things that could be more efficient if we allocated extra data structures for book keeping here. But I'm skeptical about it for the reasons I mentioned above.

lucasr: Actually, I've just had an idea on how to make this more efficient without allocating anything. I'll update the commits tomorrow.

kastiglione: :+1:

lucasr:
I've just updated the commits with an extra optimization that only traverses absolute/flexible children instead of traversing all children and checking for their type. This is accomplished by using a private linked list with no extra allocations. It just requires a couple of pointers in each node which is a good enough trade-off here.

PR #126, opened 2015-09-15
Make CSSLayout.java constants public

AaaChiuuu:


lucasr:
Next time, try to avoid the extra merge commit in the PR ;-)

PR #131, opened 2015-09-24
Revert of [Fix dimension check for STRETCH children]

vjeux:
We had a layout issue in Ads Manager after updating to the latest version of css-layout on react-native. I bisected the change to https://github.com/facebook/css-layout/commit/909c14117f118c9b36e6623590e62bff2f996c1b. It changes the condition and breaks the behavior that we had. This change was not covered by a unit test so I added one.

The 3 tabs should have the same height
ios simulator screen shot sep 23 2015 4 57 43 pm


vjeux:
Before the fix  (the three tabs have a different height)

ios simulator screen shot sep 23 2015 4 57 43 pm

After the fix (the three tabs have the same height)

ios simulator screen shot sep 24 2015 12 03 26 pm

vjeux:
@facebook-github-bot import

vjeux:
@facebook-github-bot shipit

vjeux:
screen shot 2015-09-24 at 12 23 09 pm

vjeux:
@facebook-github-bot shipit

PR #137, opened 2015-10-04
JS optimization

devongovett:
This makes the JS version run significantly faster by improving some helper functions that get called a lot. The main optimization here is inlining the margin/padding/border property lookups, rather than using expensive string manipulation operations to compute the property to read. Additionally, checking whether a value is not null is faster than using the `in` operator.

I wrote a benchmark with over 10,000 nodes to test these changes. Here are the results:

```
new x 19.39 ops/sec ±19.66% (38 runs sampled)
old x 3.54 ops/sec ±1.12% (13 runs sampled)
new is the fastest
old is 82% slower
```

Also, in a more real world test, this makes window resizing smooth vs seriously janky before.

### DIFF
--- a/src/Layout.js
+++ b/src/Layout.js
@@ -357,11 +357,11 @@ var computeLayout = (function() {
   }
 
   function isDimDefined(node, axis) {
-    return !isUndefined(node.style[dim[axis]]) && node.style[dim[axis]] >= 0;
+    return node.style[dim[axis]] != null && node.style[dim[axis]] >= 0;
   }
 
   function isPosDefined(node, pos) {
-    return !isUndefined(node.style[pos]);
+    return node.style[pos] != null;
   }
 
   function isMeasureDefined(node) {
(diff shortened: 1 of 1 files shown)


src/Layout.js
```diff
@@ -59,28 +59,6 @@ var computeLayout = (function() {
     'column-reverse': 'height'
   };
 
-  function capitalizeFirst(str) {
-    return str.charAt(0).toUpperCase() + str.slice(1);
```
vjeux: Yeah, this was stupidly slow. I was just lazy when I wrote the js version and since we never used it in prod (only the c and java), I never got around to make it faster. Thanks for doing that!

vjeux:
By the way, you should also implement caching. It is implemented in the C and Java versions but not in the JS one. The way it works is that it wraps all the recursive calls of layout with a small function that checks if we've got a cached version for the same arguments, if yes return it, if not compute it, cache it and return it.

When you mutate any style attribute, you dirty the element but also all the elements upward to the root. (I think it can be upward to the first position absolute parent but never implemented it yet. It's a bit annoying as now you've got to keep track of multiple roots when you re-render)

PR #138, opened 2015-10-04
Implement caching in the JS version

devongovett:
This implements caching for the JS version, as suggested by @vjeux in #137. Here's the results of the same benchmark I used before, compared to the current master (after my previous optimizations):

```
new x 270 ops/sec ±122.16% (94 runs sampled)
old x 22.76 ops/sec ±4.21% (43 runs sampled)
new is the fastest
old is 92% slower
```

### DIFF
--- a/src/Layout.js
+++ b/src/Layout.js
@@ -385,7 +385,7 @@ var computeLayout = (function() {
     return -getPosition(node, trailing[axis]);
   }
 
-  function layoutNode(node, parentMaxWidth, /*css_direction_t*/parentDirection) {
+  function layoutNodeImpl(node, parentMaxWidth, /*css_direction_t*/parentDirection) {
     var/*css_direction_t*/ direction = resolveDirection(node, parentDirection);
     var/*(c)!css_flex_direction_t*//*(java)!int*/ mainAxis = resolveAxis(getFlexDirection(node), direction);
     var/*(c)!css_flex_direction_t*//*(java)!int*/ crossAxis = getCrossFlexDirection(mainAxis, direction);
@@ -1068,8 +1068,32 @@ var computeLayout = (function() {
       child.nextAbsoluteChild = null;
     }
   }
+  
+  function layoutNode(node, parentMaxWidth, parentDirection) {
+    if (!node.lastLayout ||
+         node.lastLayout.requestedHeight !== node.layout.height ||
+         node.lastLayout.requestedWidth !== node.layout.width ||
+         node.lastLayout.parentMaxWidth !== parentMaxWidth) {
+      
+      if (!node.lastLayout)
+        node.lastLayout = {};
+      
+      node.lastLayout.requestedWidth = node.layout.width;
+      node.lastLayout.requestedHeight = node.layout.height;
+      node.lastLayout.parentMaxWidth = parentMaxWidth;
+    
+      layoutNodeImpl(node, parentMaxWidth, parentDirection);
+      
+      for (var key in node.layout)
+        node.lastLayout[key] = node.layout[key];
+    } else {
+      for (var key in node.layout)
+        node.layout[key] = node.lastLayout[key];
+    }
+  }
 
   return {
+    layoutNodeImpl: layoutNodeImpl,
     computeLayout: layoutNode,
     fillNodes: fillNodes
   };
--- a/src/Layout.js
+++ b/src/Layout.js
@@ -1128,9 +1128,10 @@ var computeLayout = (function() {
          node.lastLayout.requestedHeight !== node.layout.height ||
          node.lastLayout.requestedWidth !== node.layout.width ||
          node.lastLayout.parentMaxWidth !== parentMaxWidth) {
-      
-      if (!node.lastLayout)
+           
+      if (!node.lastLayout) {
         node.lastLayout = {};
+      }
       
       node.lastLayout.requestedWidth = node.layout.width;
       node.lastLayout.requestedHeight = node.layout.height;
@@ -1138,11 +1139,13 @@ var computeLayout = (function() {
     
       layoutNodeImpl(node, parentMaxWidth, parentDirection);
       
-      for (var key in node.layout)
+      for (var key in node.layout) {
         node.lastLayout[key] = node.layout[key];
+      }
     } else {
-      for (var key in node.layout)
+      for (var key in node.layout) {
         node.layout[key] = node.lastLayout[key];
+      }
     }
   }
 
--- a/src/Layout.js
+++ b/src/Layout.js
@@ -1124,12 +1124,23 @@ var computeLayout = (function() {
   }
   
   function layoutNode(node, parentMaxWidth, parentDirection) {
-    if (node.isDirty ||
-        !node.lastLayout || 
-         node.lastLayout.requestedHeight !== node.layout.height ||
-         node.lastLayout.requestedWidth !== node.layout.width ||
-         node.lastLayout.parentMaxWidth !== parentMaxWidth) {
-           
+    node.shouldUpdate = true;
+    
+    var direction = node.style.direction || CSS_DIRECTION_LTR;
+    var skipLayout = 
+      !node.isDirty &&
+      node.lastLayout &&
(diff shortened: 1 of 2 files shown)


src/Layout.js
```diff
@@ -1068,8 +1068,32 @@ var computeLayout = (function() {
...
+      if (!node.lastLayout)
+        node.lastLayout = {};
+      
+      node.lastLayout.requestedWidth = node.layout.width;
+      node.lastLayout.requestedHeight = node.layout.height;
+      node.lastLayout.parentMaxWidth = parentMaxWidth;
+    
+      layoutNodeImpl(node, parentMaxWidth, parentDirection);
+      
+      for (var key in node.layout)
```
vjeux: Nit: please always use { } even for one lines

src/Layout.js
```diff
@@ -1068,8 +1068,32 @@ var computeLayout = (function() {
...
   }
+  
+  function layoutNode(node, parentMaxWidth, parentDirection) {
+    if (!node.lastLayout ||
+         node.lastLayout.requestedHeight !== node.layout.height ||
+         node.lastLayout.requestedWidth !== node.layout.width ||
+         node.lastLayout.parentMaxWidth !== parentMaxWidth) {
+      
+      if (!node.lastLayout)
+        node.lastLayout = {};
```
vjeux: nit: always use { }

src/Layout.js
```diff
@@ -1068,8 +1068,32 @@ var computeLayout = (function() {
       child.nextAbsoluteChild = null;
     }
   }
+  
+  function layoutNode(node, parentMaxWidth, parentDirection) {
```
vjeux: Would you mind porting the caching implementation to be closer to the C version this way it won't diverge in subtle ways

https://github.com/facebook/css-layout/blob/master/src/Layout.c#L1205-L1235

vjeux:
This is awesome :)

Would you mind sending a pull request with the benchmark as well, this way other people working on optimizations can have some kind of reference to play with

src/Layout.js
```diff
@@ -1068,8 +1068,32 @@ var computeLayout = (function() {
...
+    if (!node.lastLayout ||
```
vjeux: can you support node.isDirty as well

devongovett:
Actually... the benchmark results above are not really valid since it used the same tree every run, so everything was cached (duh). Still way better though if you're just modifying a child rather than the whole tree. I'll update it.

devongovett:
Fixed nits and added node.isDirty, and node.shouldUpdate support. Also has to reset child layouts for any invalidated nodes before recomputing or the results will be wrong.

src/Layout.js
```diff
@@ -1122,8 +1122,50 @@ var computeLayout = (function() {
...
+      node.lastLayout.requestedHeight = node.layout.height;
+      node.lastLayout.parentMaxWidth = parentMaxWidth;
+      
+      // Reset child layouts
+      node.children.forEach(function(child) {
+        child.layout.width = undefined;
+        child.layout.height = undefined;
+        child.layout.top = 0;
+        child.layout.left = 0;
+        child.layout.bottom = 0;
```
vjeux: hmm, what's going on with those two? We shouldn't compute those. They can be obtained from width and height

src/Layout.js
```diff
@@ -1122,8 +1122,50 @@ var computeLayout = (function() {
       child.nextAbsoluteChild = null;
     }
   }
+  
+  function layoutNode(node, parentMaxWidth, parentDirection) {
+    if (node.isDirty ||
```
vjeux: Can you write this as a local variable skipLayout the same way the C version does it?

src/Layout.js
```diff
@@ -1122,8 +1122,50 @@ var computeLayout = (function() {
...
+        !node.lastLayout || 
+         node.lastLayout.requestedHeight !== node.layout.height ||
+         node.lastLayout.requestedWidth !== node.layout.width ||
+         node.lastLayout.parentMaxWidth !== parentMaxWidth) {
+
```
vjeux: missing direction

src/Layout.js
ColinEberhardt: @vjeux - this should probably be added to the static code analysis as part of the CI build. I'll look into it.

src/Layout.js
```diff
@@ -1122,8 +1122,50 @@ var computeLayout = (function() {
...
+        child.layout.right = 0;
+      });
+      
+      layoutNodeImpl(node, parentMaxWidth, parentDirection);
+      
+      for (var key in node.layout) {
+        node.lastLayout[key] = node.layout[key];
+      }
+      
+      node.isDirty = false;
```
vjeux: This isn't this function job to dirty/clean the nodes. It's up to whoever processes the update (see shouldUpdate)

src/Layout.js
```diff
@@ -1122,8 +1122,50 @@ var computeLayout = (function() {
...
+    } else {
+      for (var key in node.layout) {
+        node.layout[key] = node.lastLayout[key];
+      }
+    }
+    
+    node.shouldUpdate = true;
```
vjeux: can you put it at the beginning of the function so that people don't accidentally add a return somewhere and this doesn't get called anymore

src/Layout.js
ColinEberhardt: Wow, just checked and the eslint files only has one rule! That definitely needs fixing ;-)

src/Layout.js
```diff
@@ -1122,8 +1122,50 @@ var computeLayout = (function() {
...
+        child.layout.height = undefined;
+        child.layout.top = 0;
+        child.layout.left = 0;
+        child.layout.bottom = 0;
+        child.layout.right = 0;
+      });
+      
+      layoutNodeImpl(node, parentMaxWidth, parentDirection);
+      
+      for (var key in node.layout) {
```
vjeux: there's only 4 properties. Can you list them explicitly. This way it's faster and we don't accidentally forward more things than needed

devongovett:
Updated.

src/Layout.js
```diff
@@ -1122,8 +1122,53 @@ var computeLayout = (function() {
...
+      if (!node.lastLayout) {
+        node.lastLayout = {};
+      }
+      
+      node.lastLayout.requestedWidth = node.layout.width;
+      node.lastLayout.requestedHeight = node.layout.height;
+      node.lastLayout.parentMaxWidth = parentMaxWidth;
+      node.lastLayout.direction = direction;
+      
+      // Reset child layouts
```
vjeux: You shouldn't have to reset children layout, we're not doing that in the C version. Can you figure out what's going on here?

devongovett: Looks like it's done in react-native instead of in this library. https://github.com/facebook/react-native/blob/e665acca4b8839424936b38b686daf2c0f9f7f74/React/Views/RCTShadowView.m#L161-L164

In the Java version it's done here though: https://github.com/facebook/css-layout/blob/cefd6ccb967ffccc877ee7bd80500c69c3a0292b/src/java/src/com/facebook/csslayout/LayoutEngine.java#L218-L220

vjeux: Hmm, looks like we either need to update the C version or the Java one. Sucks that we have 2 different ways of reseting the children :x

I think that the C version is more "correct" as we're reseting the nodes that have been dirtied, not some random children.

devongovett: Isn't it expected that all children of a node not have layout properties defined prior to processing that node? `layoutNodeImpl` seems to assume so. Without resetting, some [isUndefined checks](https://github.com/devongovett/css-layout/blob/js-caching/src/Layout.js#L718) might fail where they shouldn't, and the position of a node [is added](https://github.com/devongovett/css-layout/blob/js-caching/src/Layout.js#L457-L464) to it's position from the previous layout pass. I think all children have to be reset. The C version just does this after a layout pass instead of before the next one.

vjeux: You are right, it doesn't matter much indeed if we do it before the beginning or after the end. It's probably better to do it inside of the layout algorithm this way the callsite doesn't have to care and possibly introduce bugs.

PR #139, opened 2015-10-05
Eslint - enforces many more rules.

ColinEberhardt:
Based on the review comments I saw for #138 I thought the linting rules could do with being 'tightened up'.

I checked the `.eslintrc` and found that the only thing being checked was the quotes style, and that was only being applied to `Layout.js`!

I've auto-configured ESLint based on the existing codebase, then added a few rules that I think are most important. Let me know what you think.


src/Layout.js
```diff
@@ -136,11 +136,11 @@ var computeLayout = (function() {
       case 'column-reverse': value = node.style.marginTop;    break;
     }
 
-    if (value != null) {
+    if (value !== null && typeof value !== 'undefined') {
```
vjeux: You should do

value !== undefined

The typeof check is when the variable may or may not exist. This is useful for checking globals. But in this case we know it exists so we can just check against the undefined value.

src/Layout.js
```diff
@@ -186,11 +186,11 @@ var computeLayout = (function() {
       case 'column-reverse': value = node.style.paddingTop;    break;
     }
 
-    if (value != null && value >= 0) {
+    if (value !== null && value >= 0) {
```
vjeux: Probably want undefined here as well and not null

devongovett: `!=` null covers both cases (`undefined == null`, but `undefined !== null`).

ColinEberhardt:
@vjeux @devongovett - I've relaxed the eqeqeq rule to permit eqeq checks against null, as @devongovett rightly points out this allows you to check for both null and undefined with a single test.

ColinEberhardt:
As an side, it does feel like the `getLeadingMargin` function, and its neighbours, share a lot of common code. I'll have a look at refactoring as a separate PR.

spicyj:
See how React and Relay have eslint set up – they use a common config from the fbjs-scripts package.

ColinEberhardt:
@spicyj thanks for the heads up, didn't realise there was a centralised lint configuration. I've updated the PR to use the `fbjs-scripts` eslint configuration and fixed all errors / warnings.

src/Layout.js
```diff
@@ -410,7 +410,7 @@ var computeLayout = (function() {
   // When the user specifically sets a value for width or height
   function setDimensionFromStyle(node, axis) {
     // The parent already computed us a width or height. We just skip it
-    if (node.layout[dim[axis]] != null) {
+    if (typeof node.layout[dim[axis]] !== 'undefined') {
```
vjeux: please use `x === undefined` instead of `typeof x === 'undefined'`

src/Layout.js
```diff
@@ -391,10 +391,10 @@ var computeLayout = (function() {
     }[axis];
 
     var boundValue = value;
-    if (max != null && max >= 0 && boundValue > max) {
+    if (max !== null && max >= 0 && boundValue > max) {
```
vjeux: should be `!== undefined`

vjeux:
Sweet! Outside of the typeof undefined and null, looks very good! Thanks for doing that

ColinEberhardt:
> please use x === undefined instead of typeof x === 'undefined'

Fair point, guarding against redefinition of `undefined` is a little unnecessary, and does make the code less readable. I've made these updates.

> Thanks for doing that

No problems :-)

vjeux:
Feel free to land if travis passes :)

ColinEberhardt:
Travis is happy - all done :+1:

PR #140, opened 2015-10-05
Add reset() method to CSSNode

lucasr:
This allows users of css-layout in Java to perform faster resets on
CSSNode in cases when you want to recycle instances.

CSSLayout.resetResult() was renamed to simply reset() for consistency.

### DIFF
--- a/src/java/src/com/facebook/csslayout/CSSNode.java
+++ b/src/java/src/com/facebook/csslayout/CSSNode.java
@@ -400,4 +400,19 @@ public class CSSNode {
       dirty();
     }
   }
+
+  /**
+   * Resets this instance to its default state. This method is meant to be used when
+   * recycling {@link CSSNode} instances.
+   */
+  public void reset() {
+    if (mParent != null || (mChildren != null && mChildren.size() > 0)) {
+      throw new IllegalStateException("You should not reset an attached CSSNode");
+    }
+
+    style.reset();
+    layout.resetResult();
+    lineIndex = 0;
+    mLayoutState = LayoutState.DIRTY;
+  }
 }
--- a/src/java/src/com/facebook/csslayout/CSSStyle.java
+++ b/src/java/src/com/facebook/csslayout/CSSStyle.java
@@ -8,40 +8,62 @@
  */
 package com.facebook.csslayout;
 
+import java.util.Arrays;
+
 /**
  * The CSS style definition for a {@link CSSNode}.
  */
 public class CSSStyle {
 
-  public CSSDirection direction = CSSDirection.INHERIT;
-  public CSSFlexDirection flexDirection = CSSFlexDirection.COLUMN;
-  public CSSJustify justifyContent = CSSJustify.FLEX_START;
-  public CSSAlign alignContent = CSSAlign.FLEX_START;
-  public CSSAlign alignItems = CSSAlign.STRETCH;
-  public CSSAlign alignSelf = CSSAlign.AUTO;
-  public CSSPositionType positionType = CSSPositionType.RELATIVE;
-  public CSSWrap flexWrap = CSSWrap.NOWRAP;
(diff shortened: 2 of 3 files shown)


lucasr:
@kmagiera @foghina ping?

src/java/src/com/facebook/csslayout/CSSStyle.java
```diff
@@ -8,40 +8,62 @@
...
-  public CSSPositionType positionType = CSSPositionType.RELATIVE;
-  public CSSWrap flexWrap = CSSWrap.NOWRAP;
+  public CSSDirection direction;
+  public CSSFlexDirection flexDirection;
+  public CSSJustify justifyContent;
+  public CSSAlign alignContent;
+  public CSSAlign alignItems;
+  public CSSAlign alignSelf;
+  public CSSPositionType positionType;
+  public CSSWrap flexWrap;
```
foghina: Why? I'd venture default values are there for a reason?

src/java/src/com/facebook/csslayout/CSSNode.java
```diff
@@ -400,4 +400,20 @@ public void setDefaultPadding(int spacingType, float padding) {
...
+   * recycling {@link CSSNode} instances.
+   */
+  public void reset() {
+    style.reset();
+    layout.reset();
+    lineIndex = 0;
+    mLayoutState = LayoutState.DIRTY;
+    mMeasureFunction = null;
+
+    if (mParent != null || (mChildren != null && mChildren.size() > 0)) {
```
foghina: Perform check before actually resetting stuff.

src/java/src/com/facebook/csslayout/CSSNode.java
```diff
@@ -400,4 +400,20 @@ public void setDefaultPadding(int spacingType, float padding) {
...
+  /**
+   * Resets this instance to its default state. This method is meant to be used when
+   * recycling {@link CSSNode} instances.
+   */
+  public void reset() {
+    style.reset();
+    layout.reset();
+    lineIndex = 0;
+    mLayoutState = LayoutState.DIRTY;
+    mMeasureFunction = null;
```
foghina: The measure function is not something that is set by the client / layout, at least not in RN. Not sure it makes sense to reset it.

src/java/src/com/facebook/csslayout/CSSStyle.java
foghina: Ah, just saw `reset()` below, ignore.

foghina:
See my inlines, otherwise

looks good to me

Not a big fan of the breaking API change, but as long as you fix usages, I'll allow it.

src/java/src/com/facebook/csslayout/CSSNode.java
lucasr: :+1:

src/java/src/com/facebook/csslayout/CSSNode.java
lucasr: My use case is for a system that recycles nodes across different layout trees. This requires the reset() call to restore the CSSNode to its original state.

lucasr: i.e. we don't subclass CSSNode per type. Layout trees are always homogeneous.

lucasr:
@foghina What API break?

foghina:
> What API break?

`resetLayout()` -> `reset()`.

lucasr:
@foghina Ah, I thought CSSLayout was package-private. Reverted this change to avoid the API break.

lucasr:
Update the branch:
* Reverted resetLayout() rename to avoid API break
* Don't reset measure function on reset to avoid problems in RN
* Attachment check before resetting in CSSNode as suggested

PR #144, opened 2015-10-14
return correct value from Width property (C#)

woehrl01:
`CSSNode.Width` returns the wrong value. It returns `DIMENSION_HEIGHT` instead of `DIMENSION_WIDTH`.


pragmatrix:
@woehrl01 great catch. Thanks!

PR #146, opened 2015-10-21
New style getters and cleanups in CSSNode

lucasr:
This PR adds all missing getters for style properties in CSSNode for consistency and because now we have some use cases for it.

Furthermore, I've renamed the existing getters and some setters to follow a more consistent pattern. This involves an API break so I'm considering this part kinda optional (even though I really think we should do it).

Having "Style" in the method names seems redundant given that we currently have setters that implicitly set style and don't have "style" in their name. For example, we have `setPadding()` but the getter is called `getStylePadding()`. We should be more consistent about this.

The following methods are renamed:
* getStyleWidth() -> getWidth()
* getStyleHeight() -> getHeight()
* setStyleWidth() -> setWidth()
* setStyleHeight() -> setHeight()
* getStyleDirection() -> getDirection()
* getStylePadding() -> getPadding()

With these changes, the CSSNode API will be more like "all setters/getters members affect style, unless they "Layout" in their name, which sounds simpler and more consistent.

### DIFF
--- a/src/java/src/com/facebook/csslayout/CSSNode.java
+++ b/src/java/src/com/facebook/csslayout/CSSNode.java
@@ -228,11 +228,25 @@ public class CSSNode {
     }
   }
 
+  /**
+   * Get this node's flex direction, as defined by style.
+   */
+  public CSSFlexDirection getFlexDirection() {
+    return style.flexDirection;
+  }
+
   public void setFlexDirection(CSSFlexDirection flexDirection) {
     if (style.flexDirection != flexDirection) {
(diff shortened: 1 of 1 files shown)


kmagiera:
looks good.

@facebook-github-bot import

lucasr:
Changed my mind after fiddling a bit more with this new API. I think we should keep the "Style" in getters that are present in both style and layout for clarity e.g. node bounds and layout direction. It still makes sense to rename getStylePadding() to getPadding() as it's a style-only attribute.

PR #150, opened 2015-11-11
grammatical fix

wpcarro:
also, on line 134, I could be wrong, but be wary of this sentence: "No C implementation of this function is provided in provided..." Perhaps this is a mistake.

thanks for all the awesome code!


ColinEberhardt:
Well spotted - thanks :-)

PR #161, opened 2016-01-05
Check layout instead of style before setting height and width to whatever is set in layout

emilsjolander:
`isDimDefined` checks is node.style[axis] is defined while in these places we do not care about `node.style[axis]` but `node.layout[axis]`. This caused parent height not to be correctly propagated to children.


vjeux:
Can you add a test case?

emilsjolander:
@vjeux will do

lucasr:
@emilsjolander Does this fix alone that bug with a specific height being passed to the measure function?

emilsjolander:
@lucasr no. that is more a limitation than a bug

emilsjolander:
@vjeux rebased to include test case. Also changed implementation to avoid some duplicate logic.

emilsjolander:
nvm this broke a lot of things in the transpiling. Fixing.

lucasr:
Nice, thanks!

Issue #160, opened 2016-01-05
Concept of max width/height for custom measure function


emilsjolander:
When implementing a custom measure function there is currently no way of knowing of the width and height passed in refer to an exact height and width or a maximum value. For example when centering a node with a custom measure function it will be passed the width and height of its parent as this is the maximum size it can be measured at. The node being measured does not have this information however and will have to make a guess.

I suggest we add something along the lines of android's concept of measurement modes (MeasureSpec). so that nodes being measured know whether they must measure to a size exactly, at most, or if the can just ignore it totally (this is currently encoded in the size by looking for NaN).

@lucasr @vjeux

NekR:
Just hit same problem today. I expected measure function to have both `width` and `height` as `NaN` for centered element with no sizes. But only `height` is `NaN` in this situation, while `weight` is the width of parent.

NekR:
Sounds more like a bug to me, does measure function really expects sizes of parent node?

For example this makes it impossible to implement measurement for text views, I have following algorithm in mind and it doesn't work: if dimension of node is set -- use it, else -- use size of contained text.

PR #162, opened 2016-01-07
Add support for measure mode

emilsjolander:
Implements #160


emilsjolander:
I messed up, will open again soon

PR #163, opened 2016-01-07
Add support for measure mode

emilsjolander:
Implements #160


emilsjolander:
@lucasr updates with requested changes

emilsjolander:
These changes break current usages. Any suggestions of how to best handle this?

PR #145, opened 2015-10-14
[Issue facebook/css-layout#127]: content re-layout of stretched items

alebo:
Trying to fix #127.

PR #171, opened 2016-02-21
Make 'stretch' aligned children fill cross-axis in containers of undefined size in the cross-axis

jsendros:
Summary: Fetched and rebased
https://github.com/facebook/css-layout/pull/145 and updated it to do 2
things:
1) Make 'stretch' aligned children grow to fill their containers in the
cross-axis if the cross-axis size is undefined. The added test might be
the best explanation of this.
2) Make sure this doesn’t decrease the frequency of the
“simpleStackCross” optimization in a relatively complex sample
environment.

Test Plan: Added a test: Container with 3 row directioned children and
no explicit height set. One child is of height 0, one of 0 height with
stretch alignment, and one of 150 height. Expected the 0 height child
remains of height 0, stretch grows to 150, and 150 stays the same.


lucasr:
Looks good, just needs the suggested cleanups. Please communicate about these changes here in #145 as this is based on his work.

jsendros:
The reason all the tests change in LayoutCachingTest.java is because they all have roots which are all default styled (flexDirection: column, alignItems: stretch). The width of the root changes, so the width of the leaves must change.

lucasr:
Looks good but you should also land the generated code for each language in the repo.

lucasr:
Never mind, the commit code does have the generated code. One C# test is failing though.

src/csharp/Facebook.CSSLayout.Tests/LayoutCachingTest.cs
```diff
@@ -155,7 +155,7 @@ public void testInvalidatesFullTreeWhenParentWidthChanges()
             root.calculateLayout();
             markLayoutAppliedForTree(root);
 
-            c0.Height = 200;
+            c0.Width = 200;
```
jsendros: This right here... This is why I wish there wasn't code(or test) duplication.

lucasr:
Landed in #145

dist/css-layout.h
```diff
@@ -1203,15 +1208,31 @@ static void layoutNodeImpl(css_node_t *node, float parentMaxWidth, float parentM
...
+                child->layout.position[leading[mainAxis]] -= getLeadingMargin(child, mainAxis) +
+                  getRelativePosition(child, mainAxis);
+                child->layout.position[trailing[mainAxis]] -= getTrailingMargin(child, mainAxis) +
+                  getRelativePosition(child, mainAxis);
+                child->layout.position[leading[crossAxis]] -= getLeadingMargin(child, crossAxis) +
+                  getRelativePosition(child, crossAxis);
+                child->layout.position[trailing[crossAxis]] -= getTrailingMargin(child, crossAxis) +
+                  getRelativePosition(child, crossAxis);
+
+                layoutNode(child, maxWidth, maxHeight, direction);
```
javache: This causes us to call layoutNode recursively a second time, which has side-effects. The bug we're seeing because of this is that items that are centered get their leading dimension incremented twice.

PR #178, opened 2016-03-25
Reset child position/sizes before calculating layout

### LINKED ISSUE #175: Positioning issues for centered items when nested under stretched items
Seeing this issues after pulling in https://github.com/facebook/css-layout/pull/171 in RN on iOS.

When there's a container view with unspecified cross-axis dimension, even though its calculated dimensions are known, it will perform a second recursive call. This second recursive call doesn't reset the position of recursively laid out children, and the offset to align any centered children will be applied twice.

Changing `isStyleDimDefined` on https://github.com/facebook/css-layout/blob/master/src/Layout.c#L1039 to `isLayoutDimDefined` mitigates the issue for us, but I'm not sure if the underlying issue of offsets being applied twice is just hidden by this.

javache:
For consistency with the other native versions, we should do layout reset inline with the `layoutNode` calculation, since `layoutNode` can potentially be invoked multiple times.

I'll need to remove/update the custom reset logic in React Native to make use of this when merging in that project.

This fixes #175.

**Test plan**: I added a new test in `Layout-test.js`, which passed the JS and Java version but failed in the C version. After my changes the C version also passes.


vjeux:
Looks good to me! Sorry for the trouble

src/__tests__/Layout-test.js
```diff
@@ -2459,6 +2458,31 @@ describe('Layout', function() {
       ]}
     );
   });
+
+  it('should center tems correctly inside a stretched layout', function() {
```
lucasr: tems -> items

src/csharp/Facebook.CSSLayout.Tests/LayoutEngineTest.cs
```diff
@@ -8392,6 +8392,96 @@ public void TestCase187()
...
+            node_3.layout.position[POSITION_TOP] = 0;
+            node_3.layout.position[POSITION_LEFT] = 25;
+            node_3.layout.dimensions[DIMENSION_WIDTH] = 50;
+            node_3.layout.dimensions[DIMENSION_HEIGHT] = 50;
+          }
+        }
+      }
+    }
+
+    test("should center tems correctly inside a stretched layout", root_node, root_layout);
```
lucasr: Ditto, items

javache:
Thanks, removed the typos.

lucasr:
Great, thanks for fixing this!

PR #182, opened 2016-03-30
Correct skipLayout condition in Layout.c

### LINKED ISSUE #155: layoutNode doesn't always skip layout when it could
(This was found while I was reading the code, not by running it or anything.)

In 36a46673f9479f94f83277464c5c43766989c417 (from @lucasr), support for reverse layout and rtl was added.

At that time, the check for whether or not to skip layout in `layoutNode` was augmented and became this:

```c
  bool skipLayout =
    !node->is_dirty(node->context) &&
    eq(layout->last_requested_dimensions[CSS_WIDTH], layout->dimensions[CSS_WIDTH]) &&
    eq(layout->last_requested_dimensions[CSS_HEIGHT], layout->dimensions[CSS_HEIGHT]) &&
    eq(layout->last_parent_max_width, parentMaxWidth);
    eq(layout->last_direction, direction);
```

I'm pretty sure that the 2nd to last line should be a ``&&`` so that the check to see if `direction` has changed would be included.

javache:
Fixes #155


vjeux:
Silly error. Accepted

PR #186, opened 2016-04-28
Force measure flag

marco-cova:
We can force the measure function of a node, if set, to be called.
The resulting sizes will be used for the layout traversal.

### DIFF
--- a/src/CSharpTranspiler.js
+++ b/src/CSharpTranspiler.js
@@ -111,6 +111,7 @@ function __transpileSingleTestToCSharp(code) {
         function(str, match1, match2, match3) {
           return 'style.' + match1 + match2.toUpperCase() + match3;
         })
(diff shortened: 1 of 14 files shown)


emilsjolander:
Changes look good. For sake of documentation please write a paragraph (in this pull request) about why we want this and when it would be useful to set this flag.

marco-cova:
We can force the measure function of a node to be called if we need to perform operations based on the sizes of the node. 
e.g. We might want to add/operate on the children of the current node if it's wide enough. https://github.com/facebook/css-layout/pull/187 will properly support changing the number of children of a node while measuring it.

emilsjolander:
Sounds good.

marco-cova:
We decided not to change css-layout but to work around this on higher level.

PR #187, opened 2016-04-29
Account for children change during measure()

marco-cova:
During a measure call, the children of the node being measured could change.
This diff make sure that nothing bad happens: the layout operate on the children of a node only after it has been measured.

### DIFF
--- a/src/Layout.c
+++ b/src/Layout.c
@@ -457,6 +457,12 @@ static bool isMeasureDefined(css_node_t *node) {
   return node->measure;
 }
 
+static void resetChildrenLayout(css_node_t *node) {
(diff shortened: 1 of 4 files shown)


src/Layout.c
```diff
@@ -457,6 +457,12 @@ static bool isMeasureDefined(css_node_t *node) {
   return node->measure;
 }
 
+static void resetChildrenLayout(css_node_t *node) {
```
emilsjolander: Why do we need this?

emilsjolander:
This looks good.

marco-cova:
We decided not to change css-layout but to work around this on a higher level.

PR #192, opened 2016-06-01
Expose getters and setters for some new properties

rigdern:
PR #185 added support for some new properties. We forgot to include getters/setters for these properties in some of the languages. This PR fixes that.

The properties include overflow and min/max height/width.


vjeux:
Is min/max actually spec compliant? If yes, this would be huge! A lot of people have been waiting for that.

rigdern:
@vjeux it is not fully spec compliant. When resolving min/max sizes, section 9.7.4 of the spec calls for n passes whereas [this only does 2 passes](https://github.com/facebook/css-layout/blob/9d62cee68bcf3247e7cb1cb3ebfd9b67b92b4dd7/src/Layout.js#L929-L940). The 2 pass approach was borrowed from code that was in css-layout prior to PR #185.

I'm not familiar with how well the min/max implementation worked prior to PR #185. However, the current implementation has worked very well for our app.

emilsjolander:
Thank you!

PR #193, opened 2016-06-02
Measure with exact measurement when stretch is defined

emilsjolander:
An optimization which will measure with EXACTLY when possible given stretch alignment. Text measurement specifically is much more performant when it knows the exact measurement up front and given that stretch alignment is the default this has quite a big performance impact on certain types of layouts.


emilsjolander:
I'm assuming my environment is set up differently than the person who last generated the code. That is why so many lines are whitespace removal.

emilsjolander:
@rigdern Would love it if you took a look

dist/css-layout.h
```diff
@@ -1064,31 +1064,42 @@ static void layoutNodeImpl(css_node_t* node, float availableWidth, float availab
           }
         }
 
+        // If child has no defined size in the cross axis and is set to stretch, set the cross
+        // axis to be measured exactly with the available inner width
+        if (!isMainAxisRow && !isUndefined(availableInnerWidth) && !isStyleDimDefined(child, CSS_FLEX_DIRECTION_ROW) && widthMeasureMode == CSS_MEASURE_MODE_EXACTLY && getAlignItem(node, child) == CSS_ALIGN_STRETCH) {
```
rigdern: Maybe add a comment to indicate that these branches are for an optimization rather than a functional requirement.

dist/css-layout.h
```diff
@@ -1266,14 +1277,17 @@ static void layoutNodeImpl(css_node_t* node, float availableWidth, float availab
...
         deltaFreeSpace -= updatedMainSize - childFlexBasis;
-        
+
         if (isMainAxisRow) {
           childWidth = updatedMainSize + getMarginAxis(currentRelativeChild, CSS_FLEX_DIRECTION_ROW);
           childWidthMeasureMode = CSS_MEASURE_MODE_EXACTLY;
-          
-          if (!isStyleDimDefined(currentRelativeChild, CSS_FLEX_DIRECTION_COLUMN)) {
+
+          if (!isUndefined(availableInnerCrossDim) && !isStyleDimDefined(currentRelativeChild, CSS_FLEX_DIRECTION_COLUMN) && heightMeasureMode == CSS_MEASURE_MODE_EXACTLY && getAlignItem(node, currentRelativeChild) == CSS_ALIGN_STRETCH) {
```
rigdern: Maybe add a comment to indicate that this branch and the one below are for an optimization rather than a functional requirement.

rigdern:
@emilsjolander Where does the performance win come from? Do you have an example layout?

In case you weren't aware, you can view the changes while ignoring whitespace by adding `?w=1` to the end of the URL. For example: https://github.com/facebook/css-layout/pull/193/files?w=1. However, I don't think you can write or view comments in this view :(

emilsjolander:
I tried to describe it in the PR description but I'll expand on it a bit.

In a simple layout such as
```html
<View style={{width: 1000}}>
  <Text>zZzZ</Text>
  <Text>zZzZzZzZ</Text>
  <Text>zZzZzZzZzZzZ</Text>
</View>
```
`View` defaults to `alignItems: 'stretch'` which previous to this diff would initially measure children with `AT_MOST` width. Not only is this wrong as they should be measured with an exact width due to there alignment it is also bad for performance as the text itself needs to figure out how big it is instead of being told a size.

dist/css-layout.h
emilsjolander: It is not just for perf though. It is also more correct to measure children with EXACTLY as that is what and alignment of stretch enforces.

src/Layout.js
```diff
@@ -809,6 +809,17 @@ var computeLayout = (function() {
             }
           }
 
+          // If child has no defined size in the cross axis and is set to stretch, set the cross
+          // axis to be measured exactly with the available inner width
+          if (!isMainAxisRow && !isUndefined(availableInnerWidth) && !isStyleDimDefined(child, CSS_FLEX_DIRECTION_ROW) && widthMeasureMode == CSS_MEASURE_MODE_EXACTLY && getAlignItem(node, child) == CSS_ALIGN_STRETCH) {
```
lucasr: Break this down into multiple lines?

rigdern:
@emilsjolander Sorry for the delay. I tested this change today in our app and didn't notice any issues.

Your change looks good to me.

emilsjolander:
Thanks Adam!

dist/css-layout.h
```diff
@@ -1273,7 +1292,10 @@ static void layoutNodeImpl(css_node_t* node, float availableWidth, float availab
           childWidth = updatedMainSize + getMarginAxis(currentRelativeChild, CSS_FLEX_DIRECTION_ROW);
           childWidthMeasureMode = CSS_MEASURE_MODE_EXACTLY;
 
-          if (!isStyleDimDefined(currentRelativeChild, CSS_FLEX_DIRECTION_COLUMN)) {
```
lucasr: wrap in multiple lines

dist/css-layout.h
```diff
@@ -1273,7 +1292,10 @@ static void layoutNodeImpl(css_node_t* node, float availableWidth, float availab
...
+          if (!isUndefined(availableInnerCrossDim) && !isStyleDimDefined(currentRelativeChild, CSS_FLEX_DIRECTION_COLUMN) && heightMeasureMode == CSS_MEASURE_MODE_EXACTLY && getAlignItem(node, currentRelativeChild) == CSS_ALIGN_STRETCH) {
```
lucasr: Ditto

lucasr:
Looks good with nits fixed.

PR #194, opened 2016-06-02
Allow measuring the root with a specific measure mode

emilsjolander:
It was previsously not possible to measure a root node with AT_MOST. This diff enables users of the api to specify the measure mode to measure with if they want.


emilsjolander:
@marco-cova

lucasr:
This needs some discussion, seems like it could lead to root style disagreeing with the inputs for layoutNode.

emilsjolander:
min/max sizes can be used instead.

PR #195, opened 2016-06-03
Improve heuristic for cache re-use

emilsjolander:
Implement better heuristics in canUseCachedMeasurement avoiding 40% measure calls compared to before (this stat comes from internal e2e tests).


emilsjolander:
@rigdern You might find this interesting as well. Saw some pretty big wins on the number of times text was measured.

src/Layout.js
```diff
@@ -1431,35 +1431,70 @@ var computeLayout = (function() {
...
+      isTextNode,
+      availableWidth,
+      availableHeight,
+      marginRow,
+      marginColumn,
+      widthMeasureMode,
+      heightMeasureMode,
+      cachedLayout) {
+
+    var heightSame =
```
lucasr: isSameHeight

src/Layout.js
```diff
@@ -1431,35 +1431,70 @@ var computeLayout = (function() {
...
+      (cachedLayout.heightMeasureMode == CSS_MEASURE_MODE_UNDEFINED && heightMeasureMode == CSS_MEASURE_MODE_UNDEFINED) ||
+        (cachedLayout.heightMeasureMode == heightMeasureMode && cachedLayout.availableHeight == availableHeight);
+
+    var widthSame =
```
lucasr: isSameWidth

src/Layout.js
lucasr: hasSameHeight

lucasr: hasSameHeight

src/Layout.js
```diff
@@ -1431,35 +1431,70 @@ var computeLayout = (function() {
...
+    if (heightSame && widthSame) {
       return true;
     }
 
-    // If the width is an exact match, try a fuzzy match on the height.
-    if (cachedLayout.availableWidth === availableWidth &&
-        cachedLayout.widthMeasureMode === widthMeasureMode &&
-        heightMeasureMode === CSS_MEASURE_MODE_EXACTLY &&
-        availableHeight - marginColumn === cachedLayout.computedHeight) {
+    var heightValid =
```
lucasr: isValidHeight

src/Layout.js
```diff
@@ -1431,35 +1431,70 @@ var computeLayout = (function() {
...
+      (cachedLayout.heightMeasureMode == CSS_MEASURE_MODE_UNDEFINED && heightMeasureMode == CSS_MEASURE_MODE_AT_MOST && cachedLayout.computedHeight <= (availableHeight - marginColumn)) ||
+        (heightMeasureMode == CSS_MEASURE_MODE_EXACTLY && cachedLayout.computedHeight == (availableHeight - marginColumn));
+
+    if (widthSame && heightValid) {
+      return true;
+    }
+
+    var widthValid =
```
lucasr: isValidWidth

lucasr: hasValidWidth

src/Layout.js
```diff
@@ -1431,35 +1431,70 @@ var computeLayout = (function() {
...
-        widthMeasureMode === CSS_MEASURE_MODE_EXACTLY &&
-        availableWidth - marginRow === cachedLayout.computedWidth) {
+    if (heightValid && widthValid) {
       return true;
     }
 
+    // We know this to be text so we can apply some more specialized heuristics.
+    if (isTextNode) {
+      if (widthSame) {
+        if (heightMeasureMode == CSS_MEASURE_MODE_UNDEFINED) {
```
lucasr: Add comment explaining each early return case.

src/Layout.js
```diff
@@ -1431,35 +1431,70 @@ var computeLayout = (function() {
...
+          return true;
+        }
+
+        if (heightMeasureMode == CSS_MEASURE_MODE_AT_MOST && cachedLayout.computedHeight < (availableHeight - marginColumn)) {
+          return true;
+        }
+
+        cachedLayout.computedHeight = availableHeight - marginColumn;
```
lucasr: No need to update cachedLayout.computedHeight in the cases above?

src/Layout.js
```diff
@@ -1431,35 +1431,70 @@ var computeLayout = (function() {
...
+        return true;
+      }
+
+      if (cachedLayout.widthMeasureMode == CSS_MEASURE_MODE_UNDEFINED) {
+        if (widthMeasureMode == CSS_MEASURE_MODE_UNDEFINED || (widthMeasureMode == CSS_MEASURE_MODE_AT_MOST && cachedLayout.computedWidth <= (availableWidth - marginRow))) {
```
lucasr: nit: break into multiple lines.

src/Layout.js
```diff
@@ -1506,13 +1541,13 @@ var computeLayout = (function() {
 
       // First, try to use the layout cache.
       if (layout.cachedLayout &&
-          canUseCachedMeasurement(availableWidth, availableHeight, marginAxisRow, marginAxisColumn,
+          canUseCachedMeasurement(node.isTextNode, availableWidth, availableHeight, marginAxisRow, marginAxisColumn,
```
lucasr: Slightly mixed about calling this "isTextNode" instead of something like "isHeightForWidth". I don't feel strongly about this though. Maybe text is a more intuitive than the concept of height-for-width layouts.

lucasr:
Looks good with nits fixed. Good stuff!

src/Layout.js
emilsjolander: No, the height has not been restricted to be smaller than it was measured so no need to change it

src/Layout.js
emilsjolander: There are some optimizations which are specific to how text is measured which might not apply to all height-for-width measured nodes. This makes it easier to think about and more clear where this optimization will work correctly.

PR #199, opened 2016-06-29
Correctly size cross axis when measuring flex basis

emilsjolander:
- Nodes were measured with the assumption of being text nodes (height depends on width) when determining flex basis. This is not always true. Even when we are just interested in the main axis (flex basis) we need to correctly constrain the cross axis.
- Some tests were wrong. Measuring texts.big and expecting it to have textSizes.smallHeight which doesn't make a lot of sense.


rigdern:
> Some tests were wrong. Measuring texts.big and expecting it to have textSizes.smallHeight which doesn't make a lot of sense.

In the context of `texts.big`, I think that `smallHeight` means that the text fits on one line whereas `bigHeight` means that the text wrapped and takes up multiple lines.

emilsjolander:
@rigdern Interesting, I had not thought of that. Super confusing. With that in mind i'll have another look. thanks

PR #200, opened 2016-07-02
Use AT_MOST measurespec when onyl max dimension is defined

emilsjolander:
Make use of max dimension styles to allow root to be measured with AT_MOST measurespec


molon:
I think this pr only support for the top-most node, it's not enough.
https://github.com/facebook/css-layout/issues/201

emilsjolander:
@molon I'll follow up with another PR. thanks for pointing this out.
@rigdern This was an oversight, i'll follow up

emilsjolander:
@rigdern #203

PR #202, opened 2016-07-06
Java: Expose extra functionality on CSSNode

rigdern:
This change is needed to support inline views on React Native for Android.


emilsjolander:
We typically just set the width/height/maxWidth/maxHeight of the root which layoutNode() is called on. Have you tried doing that? If so why does that not work for you

rigdern:
@emilsjolander That's a good point, I have not tried that. I will try that approach and get back to you. Note that the addition of `getChildrenIterable` will still be needed.

rigdern:
@emilsjolander You were right. Changing the signature of `calculateLayout` was unnecessary. I updated the PR.

emilsjolander:
@rigdern Due to large changes to the structure of the library syncing taking longer to turn on than expected master branch has been scrubbed of history. The old version is still available on the `deprecated` branch. I can add the `getChildrenIterable` method you needed if you don't want to add it yourself.

emilsjolander:
I got most if not all of the history back! rebasing might be enough

rigdern:
@emilsjolander Okay. I rebased and resubmitted the change as PR #212.

PR #218, opened 2016-08-31
Fix sizing of absolutely positioned nodes inside overflow:hidden parent

rigdern:
When an absolutely positioned node appeared inside an overflow:hidden
parent, we were limiting its height. This is inconsistent with how
layout behaves on the web.

### DIFF
--- a/CSSLayout/CSSLayout.c
+++ b/CSSLayout/CSSLayout.c
@@ -1764,16 +1764,6 @@ static void layoutNodeImpl(const CSSNodeRef node,
           childWidthMeasureMode = CSSMeasureModeAtMost;
         }
 
-        // The W3C spec doesn't say anything about the 'overflow' property,
-        // but all major browsers appear to implement the following logic.
-        if (node->style.overflow == CSSOverflowHidden) {
-          if (isMainAxisRow && CSSValueIsUndefined(childHeight) &&
-              !CSSValueIsUndefined(availableInnerHeight)) {
-            childHeight = availableInnerHeight;
-            childHeightMeasureMode = CSSMeasureModeAtMost;
-          }
-        }
-
         layoutNodeInternal(currentAbsoluteChild,
                            childWidth,
                            childHeight,
--- a/java/com/facebook/csslayout/LayoutEngine.java
+++ b/java/com/facebook/csslayout/LayoutEngine.java
@@ -1336,15 +1336,6 @@ public class LayoutEngine {
             childWidthMeasureMode = CSSMeasureMode.AT_MOST;
           }
 
-          // The W3C spec doesn't say anything about the 'overflow' property,
-          // but all major browsers appear to implement the following logic.
-          if (node.style.overflow == CSSOverflow.HIDDEN) {
-            if (isMainAxisRow && Float.isNaN(childHeight) && !Float.isNaN(availableInnerHeight)) {
-              childHeight = availableInnerHeight;
-              childHeightMeasureMode = CSSMeasureMode.AT_MOST;
-            }
-          }
-
           layoutNodeInternal(layoutContext, currentAbsoluteChild, childWidth, childHeight, direction, childWidthMeasureMode, childHeightMeasureMode, false, "abs-measure");
           childWidth = currentAbsoluteChild.layout.measuredDimensions[DIMENSION_WIDTH] + (currentAbsoluteChild.style.margin.getWithFallback(leadingSpacing[CSS_FLEX_DIRECTION_ROW], leading[CSS_FLEX_DIRECTION_ROW]) + currentAbsoluteChild.style.margin.getWithFallback(trailingSpacing[CSS_FLEX_DIRECTION_ROW], trailing[CSS_FLEX_DIRECTION_ROW]));
           childHeight = currentAbsoluteChild.layout.measuredDimensions[DIMENSION_HEIGHT] + (currentAbsoluteChild.style.margin.getWithFallback(leadingSpacing[CSS_FLEX_DIRECTION_COLUMN], leading[CSS_FLEX_DIRECTION_COLUMN]) + currentAbsoluteChild.style.margin.getWithFallback(trailingSpacing[CSS_FLEX_DIRECTION_COLUMN], trailing[CSS_FLEX_DIRECTION_COLUMN]));
--- a/tests/CSSLayoutAbsolutePositionTest.cpp
+++ b/tests/CSSLayoutAbsolutePositionTest.cpp
@@ -22,6 +22,11 @@
  * <div id="absolute_layout_width_height_left_top_right_bottom" style="width: 100px; height: 100px;">
  * <div style="width:10px; height: 10px; position: absolute; left: 10px; top: 10px; right: 10px; bottom: 10px;"></div>
  * </div>
+ * <div id="do_not_clamp_height_of_absolute_node_to_height_of_its_overflow_hidden_parent" style="height: 50px; width: 50px; overflow: hidden; flex-direction: row;">
+ * <div style="position: absolute; left: 0; top: 0;">
+ * <div style="width: 100px; height: 100px;"></div>
+ * </div>
+ * </div>
  *
  */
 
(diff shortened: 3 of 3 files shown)


emilsjolander:
Love seeing you make use of gentest.sh! Any feedback on that tool? Up until now i've been the only one using it. Looking for feedback on any improvements that can be made.

The generated code looks a bit odd. There are a bunch of added whitespace as well as the html used for generating the test is formatted wrong.

```html
<div id="do_not_clamp_height_of_absolute_node_to_height_of_its_overflow_hidden_parent" style="height: 50px; width: 50px; overflow: hidden; flex-direction: row;">
<div style="position: absolute; left: 0; top: 0;">
<div style="width: 100px; height: 100px;"></div>
</div>
</div>
```

should instead be 


```html
<div id="do_not_clamp_height_of_absolute_node_to_height_of_its_overflow_hidden_parent" style="height: 50px; width: 50px; overflow: hidden; flex-direction: row;">
<div style="position: absolute; left: 0; top: 0;"></div>
<div style="width: 100px; height: 100px;"></div>
</div>
```

I took the code from the generated test and ran it in a jsfiddle. But I can't really see the problem this is solving. The absolute layout in your test has no height or width so it does not test wether it is being limited by its parent or not. Running it in a jsfiddle setting its height and width to larger than the parent it does indeed seem to be limited by the parent's size. http://jsfiddle.net/y11txxv9/216/

rigdern:
It is nice being able to describe the test inputs using HTML rather than handwriting the equivalent C code. It reminds me of the nice JavaScript representation of the tests from the old implementation. I'll give you some feedback on gentest tomorrow.

I'm not sure why the spacing changed when I generated the tests. I copied the test code out of the Google Chrome console on `test.html`.

The original HTML is correct. The confusion might be due to the lack of indentation. The last div (100px x 100px) should be a child of the position absolute div (like in the first snippet you posted) rather than a sibling (like in the second snippet you posted). Here's the test I wrote indented properly:

```html
<div id="do_not_clamp_height_of_absolute_node_to_height_of_its_overflow_hidden_parent" style="height: 50px; width: 50px; overflow: hidden; flex-direction: row;">
  <div style="position: absolute; left: 0; top: 0;">
    <div style="width: 100px; height: 100px;"></div>
  </div>
</div>
```

I made a [jsfiddle for that snippet](https://jsfiddle.net/eo96cqur/). The absolutely positioned element has a size of 100px by 100px. Prior to my change, css-layout gave it a height of 50px. I verified that the test I added fails before my change and passes after it.

emilsjolander:
Looking forward to the feedback! Ah yes sorry, I totally misread the html due to lack of indentation. That is something I hope to fix with the gentest output at some point 👍  Thanks for the update. This look legit, I'll update the spacing for you while importing this. Thanks for yet another awesome contribution 🐹

rigdern:
@emilsjolander, here is some feedback on gentest.

I found there are a few problems with the current output of gentest. Currently, the comment it generates looks like this:

```javascript
/**
 * @Generated by gentest/gentest.sh with the following input
 *
 * <div id="absolute_layout_width_height_left_top" style="width: 100px; height: 100px;">
 * <div style="width:10px; height: 10px; position: absolute; left: 10px; top: 10px;"></div>
 * </div>
 * <div id="absolute_layout_width_height_right_bottom" style="width: 100px; height: 100px;">
 * <div style="width:10px; height: 10px; position: absolute; right: 10px; bottom: 10px;"></div>
 * </div>
 * <div id="absolute_layout_left_top_right_bottom" style="width: 100px; height: 100px;">
 * <div style="position: absolute; left: 10px; top: 10px; right: 10px; bottom: 10px;"></div>
 * </div>
 * <div id="absolute_layout_width_height_left_top_right_bottom" style="width: 100px; height: 100px;">
 * <div style="width:10px; height: 10px; position: absolute; left: 10px; top: 10px; right: 10px; bottom: 10px;"></div>
 * </div>
 * <div id="do_not_clamp_height_of_absolute_node_to_height_of_its_overflow_hidden_parent" style="height: 50px; width: 50px; overflow: hidden; flex-direction: row;">
 * <div style="position: absolute; left: 0; top: 0;">
 * <div style="width: 100px; height: 100px;"></div>
 * </div>
 * </div>
 *
 */
```

I would change it to look something like this:

```javascript
/**
 * @Generated by gentest/gentest.sh with the following input
 *
   <div id="absolute_layout_width_height_left_top" style="width: 100px; height: 100px;">
     <div style="width:10px; height: 10px; position: absolute; left: 10px; top: 10px;"></div>
   </div>

   <div id="absolute_layout_width_height_right_bottom" style="width: 100px; height: 100px;">
     <div style="width:10px; height: 10px; position: absolute; right: 10px; bottom: 10px;"></div>
   </div>

   <div id="absolute_layout_left_top_right_bottom" style="width: 100px; height: 100px;">
     <div style="position: absolute; left: 10px; top: 10px; right: 10px; bottom: 10px;"></div>
   </div>

   <div id="absolute_layout_width_height_left_top_right_bottom" style="width: 100px; height: 100px;">
     <div style="width:10px; height: 10px; position: absolute; left: 10px; top: 10px; right: 10px; bottom: 10px;"></div>
   </div>

   <div id="do_not_clamp_height_of_absolute_node_to_height_of_its_overflow_hidden_parent" style="height: 50px; width: 50px; overflow: hidden; flex-direction: row;">
     <div style="position: absolute; left: 0; top: 0;">
       <div style="width: 100px; height: 100px;"></div>
     </div>
   </div>
 *
 */
```

Here's the reasoning:
  - The `*`s that begin each line of the output are removed. This enables you to directly copy that code into the gentest tool. Currently, I have to manually remove the `*`s before copying it.
  - The HTML is properly indented to make it easier to read.
  - There's a new line between each HTML example to make it a little easier to read.

`test.html` contains an element that looks like this:

```html
<div id='default'></div>
```

Instead of including this div in `test.html`, I would create it in JavaScript with `document.createElement('div')`. The reasoning is that this makes it easier to edit `test.html` by hand (You might want to do this if you are iterating on a test. It's a faster workflow than rerunning gentest.sh each time.). When I tried editing `test.html` by hand, I ran into several problems due to the "default" div:
  1. I wrote my HTML below the "default" div instead of above it. This resulted in the "default" div becoming a test and my last test was never generated.
  2. I tried deleting the "default" div but this resulted in an exception in the code that was trying to lookup the "default" div.

Aside from that, `test.html` works pretty well.

I struggled with getting `gentest.sh` to work well. Here are the problems I hit:
  1. I didn't have the `EDITOR` environment variable set. This resulted in `gentest.sh` giving an unhelpful error message:

  ```
  ./gentest/gentest.sh: line 2: ./gentest/test.html: No such file or directory
  cat: test.html: No such file or directory
  cat: ./gentest/test.html: No such file or directory
```

  1. I tried setting the `EDITOR` environment variable but I didn't do it correctly. I still got the same unhelpful error message and I had to add logging to the script to realize `EDITOR` still wasn't set.
  1. Finally, I set `EDITOR` correctly to use VS Code. Unfortunately, the script didn't block on opening VS Code. It continued on to the other commands (generating and opening test.html in Chrome). However, I hadn't actually written anything into or saved `test.html` yet so it wasn't good that `gentest.sh` already ran those other commands and exited.

I finally got `gentest.sh` to work properly after setting the `EDITOR` to `vi`. This caused the script to wait for me to save the file before running the rest of the script.

Given all of the trouble I had with `gentest.sh` and that editing `test.html` by hand offers a quicker workflow when iterating on a test, perhaps you should get rid of `gentest.sh` and focus on having people edit `test.html` directly. If you want to do that, you should have an HTML comment in `test.html`indicating where the markup for the tests should be written.

As I mentioned earlier, describing the test inputs with HTML is much nicer than describing them with C. The HTML notation is easier to read for this particular kind of data. It's more concise and it's easier to see the parent-child relations of the elements. 😄

emilsjolander:
Thanks! The spacing and indentation issues were addressed in a commit yesterday :) I only ever tested setting `EDITOR` to `vi` so i'm not all to surprised it doesn't work with anything else. Editing test.html was never a workflow I tried so it was never meant to work seamlessly, though I'll think of a way to make it better, your suggestions here are great.

I'll create a couple tasks to fix some of these issues when I get back from vacation 👍

PR #223, opened 2016-09-23
Removes unnecessary CSSLayoutContext param from CalculateLayout

### LINKED ISSUE #222: CalculateLayout no longer requires CSSLayoutContext
Since the native layer through JNI and P/Invoke do not use the `CSSLayoutContext`, should we get rid of that parameter at least in the C# version, if not the Java version as well?

mattpodwysocki:
Closes #222 for the `csharp` implementation which does not use the `CSSLayoutContext` parameter in the `CalculateLayout` method call.


emilsjolander:
Thanks! looks good. I'll merge this on Monday morning

facebook-github-bot-4:
Thanks for importing. If you are a Facebook employee, you can view this diff [on Phabricator](https://phabricator.intern.facebook.com/D3918948).

PR #233, opened 2016-10-25
Remove children from the end to prevent copying over

woehrl01:
This PR leads to removing the children from the end. [So we don't have the overhead of copying them to the front](https://github.com/facebook/css-layout/blob/dcaef7ecb0ba35a1132386e3a542fe5c7eb11fa5/CSSLayout/CSSNodeList.c#L62).


splhack:
👍

Issue #236, opened 2016-10-29
css_applyLayout crack


singno:
![image](https://cloud.githubusercontent.com/assets/5654690/19830092/4c5b62f6-9e23-11e6-8c0b-d2604274a040.png)

![image](https://cloud.githubusercontent.com/assets/5654690/19830096/5b87c080-9e23-11e6-8827-969271633d7e.png)

emilsjolander:
@dshahidehpour Can you take a look?

gimenete:
It fails because `node->children` is NULL. The problem seems to have been introduced here https://github.com/facebook/css-layout/commit/01c2ac3369b6e403e5f7a24b91a01b7bb792eed7?diff=unified#diff-937e78d20ebabc2f92fe6f07dc70d4a3L198

Using `node->children = CSSNodeListNew(4);` in `CSSNodeInit` works for me but it's not the optimal solution.

dshahidehpour:
Thanks for spotting! I'm adding a fix right now.

dshahidehpour:
Just landed a fix with #ced779b25966448feb9bd792efedd29dfa6ab9e0

gimenete:
It works! Thanks :)

PR #237, opened 2016-10-31
Prevent array out of bound access

woehrl01:
Even so the problem of #236 has been fixed via ced779b there is still a case where you have a out of bound array access of undefined behaviour. This PR prevents the out of bound access of the underlying c array. The out of bound access happens if you already have 4 subViews and add another one. Then you will access CSSNodeGetChild with 4 which points into uninitialized memory.


emilsjolander:
Looks good! merging

PR #239, opened 2016-11-02
Add inline definition for Microsoft Visual Studio

anql:
The inline keyword is available only in C++. The __inline and __forceinline keywords are available in both C and C++. For compatibility with previous versions, _inline is a synonym for __inline.

https://msdn.microsoft.com/en-us/library/z8y1yy88(v=vs.120).aspx

### DIFF
--- a/CSSLayout/CSSLayout.c
+++ b/CSSLayout/CSSLayout.c
@@ -16,6 +16,10 @@
 #include <float.h>
 #define isnan _isnan
 
+#ifndef __cplusplus
+#define inline __inline
+#endif
+
 /* define fmaxf if < VC12 */
 #if _MSC_VER < 1800
 __forceinline const float fmaxf(const float a, const float b) {


emilsjolander:
Thanks! although __inline is vendor specific (msvc) so can we only define it for msvc?

anql:
Only define it for msvc to fix issues like this
http://stackoverflow.com/questions/24736304/unable-to-use-inline-in-declaration-get-error-c2054

emilsjolander:
Exactly, could you update the PR?

emilsjolander:
@anql Why did you close this PR?

anql:
:(   sorry, It’s my mistake. 
Google search "how to update pull request github",
http://stackoverflow.com/questions/9790448/how-to-update-a-pull-request
reopen and click "Update branch".
:)

emilsjolander:
@anql with the update you are still using `__inline` on non msvc compilers. Could you instead check for the _MSC_VER macro? All compilers with support for c99 will support the standard `inline` keyword.

anql:
default

emilsjolander:
ah sorry! Totally missed that in the diff view 👍 I'll import and merge this. Thanks!

PR #257, opened 2016-11-21
Use ASSERT_FLOAT_EQ instead of ASSERT_EQ

woehrl01:
Changes the unit test comparsion to use ```ASSERT_FLOAT_EQ``` instead of ```ASSERT_EQ``` as they check float values.


splhack:
@woehrl01 could you check Travis-CI error log? https://travis-ci.org/facebook/css-layout/builds/177720507

woehrl01:
Strange, as I'm not that familiar with clang whats the problem here? In VS it compiles. Do I need to write the digits as ```10.0f```?

emilsjolander:
@woehrl01 don't worry about it. Easy fix, I'll do it myself tomorrow before merging.

emilsjolander:
@woehrl01 Why did you add the `f` suffix in the first place? Does VS not implicitly treat the number literal as a float if used in a floating point context?

woehrl01:
@emilsjolander for ```integer``` value, it's true. But decimal numbers are treated as ```double```, which fail to compile (in VS). I think we need to detect if it's a fractial number and add the ```f``` in that case only.

emilsjolander:
Or we just prepend a float cast `(float) value`

emilsjolander:
@woehrl01 Could you validate that that would work in VS? I'll update the code in that case

woehrl01:
Sorry. Just had a look, you can omit the ```f``` it "just" leads to a warning. I had "treat warning as error" enabled. Casting to float also works. I'd prefer the cast, to prevent the warning.

emilsjolander:
@woehrl01 casting produces very verbose test code though :/ Could you instead investigate if you can pass some flag to VS so it treats `0.5` as a float?

emilsjolander:
Maybe you could suppress the warning for tests?

splhack:
how about something like in gentest*.js

    function isFloat(n) {
        return Number(n) === n && n % 1 !== 0;
    }

    ... v0 + (isFloat(v0) ? 'f' : '') ...

emilsjolander:
@splhack that works

PR #258, opened 2016-11-22
Add feature to use percentage as value unit

woehrl01:
Adds the feature to use percentage as a value unit.

You can use the function ```YGPx(float)``` and ```YGPercent(float)``` for convenience. 

I did some benchmarks:

```
Without Percentage Feature - Release x86:

Stack with flex: median: 0.000000 ms, stddev: 0.146683 ms
Align stretch in undefined axis: median: 0.000000 ms, stddev: 0.136525 ms
Nested flex: median: 0.000000 ms, stddev: 0.490101 ms
Huge nested layout: median: 23.000000 ms, stddev: 0.928291 ms

Stack with flex: median: 0.000000 ms, stddev: 0.170587 ms
Align stretch in undefined axis: median: 0.000000 ms, stddev: 0.143384 ms
Nested flex: median: 0.000000 ms, stddev: 0.477791 ms
Huge nested layout: median: 22.000000 ms, stddev: 2.129779 ms


With Percentage Feature - Release x86:

Stack with flex: median: 0.000000 ms, stddev: 0.132951 ms
Align stretch in undefined axis: median: 0.000000 ms, stddev: 0.136525 ms
Nested flex: median: 0.000000 ms, stddev: 0.489570 ms
Huge nested layout: median: 21.000000 ms, stddev: 1.390476 ms

Stack with flex: median: 0.000000 ms, stddev: 0.146969 ms
Align stretch in undefined axis: median: 0.000000 ms, stddev: 0.129271 ms
Nested flex: median: 0.000000 ms, stddev: 0.481787 ms
Huge nested layout: median: 21.000000 ms, stddev: 1.695898 ms
```

It is slightly faster, as I was able to reduce the calls to ```isnan()```.


emilsjolander:
This is really cool! I'm worried about perf, we need to benchmark this on low end android phones to get a sense of what the impact is. In a perfect world perf should only be effected when using percentage values. Percentage values are quite rare so impacting perf negatively for the rest of use cases is not great.

Some things I would like to see:
- Description of why this should be added to css-layout. What use cases does it solve that are hard to solve with the current library?
- Many more tests. We need to test how this works for width/height/flexBasis/minHeight/maxHeight/minWidth/maxWidth in both the main and cross axis. Also how does this interact with flex-grow and flex-shrink.

Looking forward to working together on this.

emilsjolander:
> Should the Px() and Percent() functions be prefixed with CSSLayout?
Prefix them with `CSS`.

woehrl01:
> We need to benchmark this on low end android phones to get a sense of what the impact is. 

As I have non available in near future, could someone please try this on a low end device? 

> What use cases does it solve that are hard to solve with the current library?

Lot's of use cases can be solved via "spacer" views. But in my opinion this leads to less readable code on these places, as the intent isn't that clear anymore. 

Additionally if you try to position an element absolute with percentage (e.g ```top: 20%```, ```left: 10%```) this shouldn't be possible at all, without having to creating a spacer view grid.

One use case I have is to layout boxes with wrapping, where I want to have two elements in a row. I would express this as `FlexShrink: 0, FlexGrow: 0, FlexBasis: 50%`.

But my personal prefered use case is to make my view code more readable and maintainable.

> Many more tests. 

Will do

> Additional benefit

This is the start for units, ```em``` should be pretty easy to add (even so this can be calculated manually by now)

> In a perfect world perf should only be effected when using percentage values

not sure if this is possible without doing a big refactoring. But I'll have a look.

emilsjolander:
Would love concrete examples of what views benefit from this, not description of what it could be used for. I'm not really questioning the usefulness of it but it would be great to get a sense of the impact this has. In practice what layout problems does it solve. I myself find it very seldom that I would like to position something a percentage from the top/left as this generally does not scale very nicely. If there is a performance hit due to adding units I want to be able to show all the concrete benefits it brings.

emilsjolander:
I can make sure to run some tests on perf on android. I won't have much time in the coming week though.

woehrl01:
Sorry for bringing up a fictional example. The only real use case I had is the wrapping stuff, where I currently have to reimplement the wrapping logic, by boxing each line in its own container and distribute the elements across it. But I'm not able to express a minimal width on the containig elements, which I would like to.

e.g:
```
<div style="width: [dynamic]px; flex-wrap: wrap;flex-direction:row;display:flex">
 <div style="flex-shrink: 0; flex-grow: 1; flex-basis: 50%; height: 20px; min-width: 50px;background-color:blue;">
 </div>
 <div style="flex-shrink: 0; flex-grow: 1; flex-basis: 50%; height: 20px; min-width: 50px;background-color:red;">
 </div>
 <div style="flex-shrink: 0; flex-grow: 1; flex-basis: 50%; height: 20px; min-width: 50px;background-color:green;">
 </div>
 <div style="flex-shrink: 0; flex-grow: 1; flex-basis: 50%; height: 20px; min-width: 50px;background-color:yellow;">
 </div>
 <div style="flex-shrink: 0; flex-grow: 1; flex-basis: 50%; height: 20px; min-width: 50px;background-color:pink;">
 </div>
</div>
```

for a width with >= 100px you have two elements per row and < 100px you have a single element per row.

emilsjolander:
@woehrl01 Thanks for the update!

woehrl01:
Ups, got it even faster than without this feature 😆 (same maschine), still interested in how this performs on android.

```
Stack with flex: median: 0.000000 ms, stddev: 0.132951 ms
Align stretch in undefined axis: median: 0.000000 ms, stddev: 0.136525 ms
Nested flex: median: 0.000000 ms, stddev: 0.489570 ms
Huge nested layout: median: 21.000000 ms, stddev: 1.390476 ms

Stack with flex: median: 0.000000 ms, stddev: 0.146969 ms
Align stretch in undefined axis: median: 0.000000 ms, stddev: 0.129271 ms
Nested flex: median: 0.000000 ms, stddev: 0.481787 ms
Huge nested layout: median: 21.000000 ms, stddev: 1.695898 ms
```

emilsjolander:
@woehrl01 awesome! I landed a diff yesterday for experiment support in gentest. I would suggest you focus on added a bunch of test cases now 👍

woehrl01:
@emilsjolander Merged master branch. And got the first failing test. Will look into this furter in the next days.

woehrl01:
@emilsjolander the failing test is not introduced via this implementation. created a new issue #261 for tracking this.

emilsjolander:
Please change enums.py and re-generate the enum files by running `python enums.py` 👍

emilsjolander:
I think once this is rebased it will be pretty close to mergable

woehrl01:
@emilsjolander sadly not. I already did a local rebase, but I found a test layout which doesn't get calculated correctly. Still want to find that bug before I'll update the PR.

emilsjolander:
@woehrl01 Ok! Feel free to update the PR with your progress and I can help debug if you want.

woehrl01:
@emilsjolander I have a trace, we need to calculate the padding/margin percentage only based on the width. see https://www.w3.org/TR/CSS2/box.html#margin-properties

> The percentage is calculated with respect to the width of the generated box's containing block. Note that this is true for 'margin-top' and 'margin-bottom' as well. If the containing block's width depends on this element, then the resulting layout is undefined in CSS 2.1.

sorry for letting this slipping through. 😞

emilsjolander:
wait.. margin-top: 50% is 50% of the width?? this makes no sense.

woehrl01:
No, its 50% of the parents width. This SO answer explains it best why: http://stackoverflow.com/a/18295172

Couldn't believe this either at first 😄

emilsjolander:
Oh wow that is super confusing.

woehrl01:
@emilsjolander test is passing 👍 😄

emilsjolander:
This is starting to look very promising.

Could you
- Remove unrelated changes
- Add tests for absolute positioning
- Add tests for border
- Add tests for margin (only tested in nested test now would like something more specific as well)
- Add tests for padding (only tested in nested test now would like something more specific as well)

Also i'm not sure about the best way on constructing these values. `YGPx` etc is pretty verbose but i'm also not a fan of exposing a global unprefixed `px` function.

gentest/fixtures/YGPercentageTest.html
```diff
@@ -0,0 +1,86 @@
...
+</div>
+
+<div id="percentage_margin_should_calculate_based_only_on_width" experiments="Rounding" style="width: 200px; height: 100px;">
+ <div style="flex-grow: 1; margin: 10%;">
+	<div style="width: 10px; height: 10px;">
+	</div>
+ </div>
+</div>
+
+<div id="percentage_padding_should_calculate_based_only_on_width" experiments="Rounding" style="width: 200px; height: 100px;">
```
emilsjolander: Indentation here seems a bit off. Could you please update your tests to look more like the others?
```html

<div id="percentage_padding_should_calculate_based_only_on_width" experiments="Rounding" style="width: 200px; height: 100px;">
    <div style="flex-grow: 1; padding: 10%;">
        <div style="width: 10px; height: 10px;"></div>
    </div>
</div>
```

gentest/fixtures/YGPercentageTest.html
```diff
@@ -0,0 +1,86 @@
+<div id="percentage_width_height" experiments="Rounding" style="width: 200px; height: 200px; flex-direction: row;">
```
emilsjolander: Still missing tests for % border values

woehrl01:
> Remove unrelated changes
> Add tests for absolute positioning
> Add tests for margin (only tested in nested test now would like something more specific as well)
> Add tests for padding (only tested in nested test now would like something more specific as well)

added those tests

> Add tests for border

there is no need to write tests for border, as border can't have percentage in css.

> Also i'm not sure about the best way on constructing these values. YGPx etc is pretty verbose but i'm also not a fan of exposing a global unprefixed px function.

I already added two function e.g. ```YGNodeStyleSetWidthWithUnit``` and one only ```YGNodeStyleSetWidth``` which takes only ```float``` and means pixel.

My idea was to use ```*WithUnit``` in any interop code, it maybe should be used with an enum instead of the YGPx, e.g. ```YGNodeStyleSetWidthWithUnit(root, 14, YGUnitPercent)```. And to keep the "old" for not having a breaking change and to be able to set those values less verbose on ```C``` code.

For C# I'd suggest having a struct with implicit convertion from ```float``` and an extension method to make something like:

```
node.Height = 10f;
node.Width = 5.Percent()
```

For Java you can use overloads

```
node.SetHeight(10);
node.SetHeight(10, YGUnitPercent);
```

gentest/fixtures/YGPercentageTest.html
woehrl01: there is no need to write tests for border, as border can't have percentage in css.

emilsjolander:
> there is no need to write tests for border, as border can't have percentage in css.

So consistent.

> YGNodeStyleSetWidthWithUnit

Didn't see that, cool. It's pretty verbose and ads a lot of functions to the API. I think the API for this is worth thinking a bit about before merging, I'll try to see if I can figure out something better. Feel free to post more suggestions as well.

> 5.Percent()

This seems super magical to me. But if that is the standard why for doing such things in C# then by all means. I would probably prefer an enum in both java and c#.

** Travis is failing **

woehrl01:
Regarding C:

I think we should have multiple functions to not break many users and still make this opt-in for the ones who want to use percentage. But maybe as a second layer? ```YGNodeStyleSetWidth``` is calling ```YGNodeStyleSetWidthWithUnit``` in a thin wrapper?

Regarding C#:

Was just an idea, could also be something like:
```csharp
node.Width = Percent.Of(5f);
```
If you use an enum for C# you miss the change to use Properties, also there is this ```YogaNode.Create()```. I would personaly still prefere the "object initializer syntax":

```csharp
var n = new YogaNode{
    Width = 5,
    Height = Percent.Of(10)
}
```

emilsjolander:
I'll finish up reviewing the algorithm + test changes later today / early next week.

> Regarding C

I like that you are thinking about not breaking the API but in this case i'm holding off on marking a stable release for changes such as this one. Also very few projects depend directly on the C version and instead use the wrappers. As I said I think it is worth taking some time and thinking about this API.

> Regarding C#

Makes sense.

woehrl01:
Travis is still failing, because of the not exposed possibility of percentage in Java, I expect for C# to have the same problem. 

Take your time reviewing and thinking about a good API. As soon as we have a well-conceived idea I'm happy to make the required changes.

yoga/YGEnums.h
```diff
@@ -21,6 +21,12 @@ typedef enum YGFlexDirection {
   YGFlexDirectionCount,
 } YGFlexDirection;
 
+typedef enum YGUnit {
+  YGUnitPixel,
+  YGUnitPercent,
+  YGUnitModeCount,
```
emilsjolander: YGUnitModeCount?? should be YGUnitCount probably. Did you forget to re-generate after changes?

yoga/Yoga.h
```diff
@@ -38,6 +38,16 @@ typedef struct YGSize {
   float height;
 } YGSize;
 
+typedef struct YGValue{
+  float value;
+  YGUnit unit;
+  bool defined;
+
```
emilsjolander: remove empty line

yoga/Yoga.h
```diff
@@ -38,6 +38,16 @@ typedef struct YGSize {
   float height;
 } YGSize;
 
+typedef struct YGValue{
```
emilsjolander: space before { (actually just run format.sh)

yoga/Yoga.h
```diff
@@ -82,7 +92,8 @@ WIN_EXPORT bool YGNodeIsDirty(const YGNodeRef node);
 
 WIN_EXPORT void YGNodePrint(const YGNodeRef node, const YGPrintOptions options);
 
-WIN_EXPORT bool YGValueIsUndefined(const float value);
+WIN_EXPORT bool YGFloatIsUndefined(const float value);
+WIN_EXPORT bool YGValueIsUndefined(const YGValue value);
```
emilsjolander: No reason to have this function as we export the `defined` member of `YGValue`

yoga/Yoga.h
```diff
@@ -107,12 +118,24 @@ WIN_EXPORT void YGNodeCopyStyle(const YGNodeRef dstNode, const YGNodeRef srcNode
   WIN_EXPORT void YGNodeStyleSet##name(const YGNodeRef node, const type paramName); \
   WIN_EXPORT type YGNodeStyleGet##name(const YGNodeRef node);
 
+#define YG_NODE_STYLE_PROPERTY_UNIT(type, name, paramName)                                \
+  YG_NODE_STYLE_PROPERTY(float, name, paramName); \
+  WIN_EXPORT void YGNodeStyleSet##name##WithUnit(const YGNodeRef node, const type paramName); \
```
emilsjolander: Let's remove the *WithUnit APIs and just use units everywhere. This makes the API smaller and more concise. Also it forces the user to think about units.

yoga/Yoga.c
```diff
@@ -179,10 +179,26 @@ static inline float YGComputedEdgeValue(const float edges[YGEdgeCount],
...
+    result.unit = YGUnitPixel;
+    return result;
   }
 
-  return defaultValue;
+  YGValue result;
+  result.value = defaultValue;
+  result.defined = (defaultValue < 0 || defaultValue >= 0); /* is faster than a nan function call and enough at this point */
+  result.unit = YGUnitPixel;
+  return result;
```
emilsjolander: return YGValue {
  .value = ...,
  .defined = ...,
  .unit = ...,
};

yoga/Yoga.c
```diff
@@ -194,7 +210,9 @@ static void YGNodeInit(const YGNodeRef node) {
   node->style.flex = YGUndefined;
   node->style.flexGrow = YGUndefined;
   node->style.flexShrink = YGUndefined;
-  node->style.flexBasis = YGUndefined;
+  node->style.flexBasis.value = YGUndefined;
```
emilsjolander: Create a `YGValueUndefined` static instance which you just copy here and below.

yoga/Yoga.c
```diff
@@ -241,6 +279,22 @@ static void YGNodeInit(const YGNodeRef node) {
 
 int32_t gNodeInstanceCount = 0;
 
+YGValue YGPx(const float value){
+  YGValue result;
+  result.value = value;
+  result.defined = !YGFloatIsUndefined(value);
+  result.unit = YGUnitPixel;
+  return result;
```
emilsjolander: return YGValue { ... };

yoga/Yoga.c
```diff
@@ -241,6 +279,22 @@ static void YGNodeInit(const YGNodeRef node) {
...
+}
+
+YGValue YGPercent(const float value){
+  YGValue result;
```
emilsjolander: return YGValue { ... };

yoga/Yoga.c
```diff
@@ -482,17 +603,34 @@ bool YGLayoutNodeInternal(const YGNodeRef node,
...
                           const bool performLayout,
                           const char *reason);
 
-inline bool YGValueIsUndefined(const float value) {
+inline bool YGFloatIsUndefined(const float value) {
   return isnan(value);
 }
 
-static inline bool YGFloatsEqual(const float a, const float b) {
+inline bool YGValueIsUndefined(const YGValue value) {
```
emilsjolander: just check for `.defined`. this function just makes it more verbose.

yoga/Yoga.c
```diff
@@ -179,10 +179,26 @@ static inline float YGComputedEdgeValue(const float edges[YGEdgeCount],
...
 
-  return defaultValue;
+  YGValue result;
+  result.value = defaultValue;
+  result.defined = (defaultValue < 0 || defaultValue >= 0); /* is faster than a nan function call and enough at this point */
+  result.unit = YGUnitPixel;
+  return result;
+}
+
+static inline float YGValueToFloat(const YGValue unit, const float parentSize) {
```
emilsjolander: `YGValueToFloat` is not very descriptive of what the function does. I would prefer `YGValueResolve` or something similar. What do you think?

yoga/Yoga.c
```diff
@@ -1138,38 +1289,38 @@ static void YGNodeAbsoluteLayoutChild(const YGNodeRef node,
       childHeight = node->layout.measuredDimensions[YGDimensionHeight] -
                     (YGNodeLeadingBorder(node, YGFlexDirectionColumn) +
                      YGNodeTrailingBorder(node, YGFlexDirectionColumn)) -
-                    (YGNodeLeadingPosition(child, YGFlexDirectionColumn) +
-                     YGNodeTrailingPosition(child, YGFlexDirectionColumn));
-      childHeight = YGNodeBoundAxis(child, YGFlexDirectionColumn, childHeight);
+                    (YGNodeLeadingPosition(child, YGFlexDirectionColumn, width) +
```
emilsjolander: Position percentages are calculated based on the same axis as the position (https://jsfiddle.net/jckmwztt/15/) this is different from margin and padding and instead similar to width and height. So this (and other locations) should be `YGFlexDirectionColumn, height`

yoga/Yoga.c
woehrl01: Please don't let me use this syntax here and at the other spots. Visual Studio (at least VS13 which I use) doesn't recognize this syntax, which would need me to change this one everytime I need to compile. (thats the reason I missed some tests at first)

yoga/Yoga.c
woehrl01: sounds good!

yoga/Yoga.c
woehrl01: you mean a "global" instance? I could use this in ```YGComputedEdgeValue```, too.

yoga/Yoga.c
woehrl01: nice catch, thanks!

yoga/Yoga.h
emilsjolander: The getter should return a YGValue as well.

yoga/Yoga.c
emilsjolander: VS15 has support for this syntax and it is used throughout the project so i suggest upgrading. I would prefer it and it is part of C99. Is there a reason you are using a old VS build? if there are legit usages for old VS then I would be OK with leaving as is.

emilsjolander: I think it does from searching google :p please double check yourself

woehrl01: sorry, just downloaded VS15 community and I can't get it to compile with this syntax 😞

yoga/Yoga.h
woehrl01: I would like to do this, but setting everything up on my maschine would be a huge job. As I have no linux runtime currently available, and also no clang localy. It would be greate if you could run this, after your import, if you don't mind. Same for enum.py, currently I have no python installed. If not I'll try my best to set this up locally, could take a while (the clang environment).

yoga/Yoga.c
woehrl01: ups, got it to work, just can't return it at the same time. 😕

yoga/Yoga.h
```diff
@@ -38,6 +38,15 @@ typedef struct YGSize {
   float height;
 } YGSize;
 
+typedef struct YGValue {
+  float value;
+  YGUnit unit;
+  bool defined;
```
emilsjolander: isDefined

yoga/Yoga.h
```diff
@@ -107,12 +116,22 @@ WIN_EXPORT void YGNodeCopyStyle(const YGNodeRef dstNode, const YGNodeRef srcNode
   WIN_EXPORT void YGNodeStyleSet##name(const YGNodeRef node, const type paramName); \
   WIN_EXPORT type YGNodeStyleGet##name(const YGNodeRef node);
 
+#define YG_NODE_STYLE_PROPERTY_UNIT(type, name, paramName)                                \
+  WIN_EXPORT void YGNodeStyleSet##name##(const YGNodeRef node, const type paramName); \
```
emilsjolander: ##name##( -> ##name(

yoga/Yoga.h
woehrl01: great idea!

emilsjolander:
There are some merge conflicts according to github, otherwise looks good. Will try to import internally tomorrow.

woehrl01:
did a merge with the upstream and modified the c# wrapper

emilsjolander:
Seems like the benchmarks, csharp, and java tests are still failing. I can take a look tomorrow

woehrl01:
It would be great if you could take a look at java. It's ages for me the last time I did some JNI.

emilsjolander:
The java version will take a bit more work to update. I'm not super familiar with JNI either tbh. But i'll have a look this week.

woehrl01:
Did some performance improvements (~20%) on my pc: 

```
Stack with flex: median: 0.000000 ms, stddev: 0.143384 ms
Align stretch in undefined axis: median: 0.000000 ms, stddev: 0.129271 ms
Nested flex: median: 0.000000 ms, stddev: 0.432434 ms
Huge nested layout: median: 17.000000 ms, stddev: 3.835411 ms

Stack with flex: median: 0.000000 ms, stddev: 0.132951 ms
Align stretch in undefined axis: median: 0.000000 ms, stddev: 0.121552 ms
Nested flex: median: 0.000000 ms, stddev: 0.439180 ms
Huge nested layout: median: 16.000000 ms, stddev: 1.645686 ms
```

emilsjolander:
@woehrl01 While implementing the JNI bindings I thought of a slightly different way to implement this. Wanted to get your thoughts on it.

Instead of `YGValue` we store the `YGUnit` and the `isDefined` flag as part of the numeric value itself. We could reserve a couple bits in the unit's value for these flags. This would require switching from `float` values to `uint64` types, the first 32 bits storing the meta data and the second 32 bits storing the floating point value.

This would allow for a couple things.
- Not changing the API. `YGNodeStyleSetWidth(node, 10)` would still set the width of a node to the pixel value of 10.
- Only require calling a `YGValue` constructor when using percentage (which should be much more seldom than pixel values). `YGNodeStyleSetWidth(node, YGPercent(10))` would set the width of the node to 10%.

The major downside is that we would no longer allow setting a floating point value directly as the parameter is now a integer, `YGNodeStyleSetWidth(node, 10.5)` would be a compile time error. We might be OK with this though as setting half pixels on styles is generally discouraged anyways. The thing to keep in mind here is platforms not using Yoga with pixels but points instead (iOS). This would require them to transform pixel value to points before passing values to Yoga (but we control the bindings here so the end user should not need to care). @dshahidehpour What do you think about this?

This idea came from the implementing the java bindings as there we need to do this either way to avoid allocations and passing larger objects over JNI which can be much more expensive than primitive types such as floats and integers.

woehrl01:
@emilsjolander 
Sounds good! I have an extended idea on top of it.

if you call ```YGNodeStyleSetWidth(node, 10)``` you will set an Int64 where the upper is fully 0. Which means we provide an int in the lower. If we call it with ```YGNodeStyleSetWidth(node, YGPx(10.5))``` we store the lower as a float and add an additional flag in the upper bits to detect this. So we can still set floating values via the api, we just need to be explicitly use ```YGPx``` then.

Idea 1: Regarding points and pixel. Maybe we should rename it and use Points instead of Pixel all over.
Idea 2: We extend the Api with an additonal Points value factor (like setting the logger). So we can also do the rounding with higher density + an additional YGUnitPoints where we calculate ```unit.value * densityFactor```

dshahidehpour:
> The major downside is that we would no longer allow setting a floating point value directly as the parameter is now a integer, YGNodeStyleSetWidth(node, 10.5) would be a compile time error.

I think this is a huge piece of functionality to throw away. As you correctly stated in your comment, iOS deals with Points, not Pixels. If someone gave us a design where they wanted to something to be 21 *pixels* wide (very reasonable), it would no longer be possible to use Yoga on iOS because this is how you would define value so of 21 pixels:

```objc
YGNodeStyleSetWidth(node, 10.5); // non-retina phones.
YGNodeStyleSetWidth(node, 7) // retina-phones.
```
So, I'm very much against that.

> Only require calling a YGValue constructor when using percentage (which should be much more seldom than pixel values). YGNodeStyleSetWidth(node, YGPercent(10)) would set the width of the node to 10%.

I like this API, but, I think we would have to rename `YGPx` to something else. I don't think that Yoga should know (or care) what the units of a coordinate system are. It should only care that you are dealing with absolute values or ratio values (percents). For all we care, someone could be using Yoga to calculate the layout of their new backyard and all the measurements are in meters. 

cc: @emilsjolander

emilsjolander:
We are going to need to introduce some notion of Points into Yoga at some point. Currently pixel rounding (behind experiment) is kinda broken on iOS as it does not take the point scale factor into account.

So we could end up with an API similar to this where pixel values have to be integers but both points and percentage values can be floating point. The point scale factor would also be taken into account when performing pixel rounding.

```C
YGSetPointScaleFactor([UIScreen mainScreen].scale)
YGNodeStyleSetWidth(node, 10)
YGNodeStyleSetWidth(node, YGPoint(10.0))
YGNodeStyleSetWidth(node, YGPercent(10.0))
```

I agree that the API for iOS developers still has to be nice. But I think this is where `YogaKit` plays a vital role. `YogaKit` allows us to hide some of this detail behind the scenes and always just default to point values and setting the scale factor automatically.

@dshahidehpour how does that sound? I'm not by any means certain about this.

woehrl01:
@emilsjolander do you want to replace the struct inside the c, too? Could reduce the creation time of new nodes as currently we do not perfectly memory align due to the bool flag.

emilsjolander:
@woehrl01 yeah, as I see it `YGValue` would just be an alias for a `uint64_t`. But please don't start making any changes yet as we aren't certain about this yet 👍 Don't want you to waste any time

woehrl01:
@emilsjolander no worries 😄 . But how about negative values? ```YGNodeStyleSetPosition(node, YGEdgeTop, -10)``` would possibly throw some warnings.

emilsjolander:
haha yeah, so perhaps `int64_t` and not `uint64_t` :p

dshahidehpour:
> We are going to need to introduce some notion of Points into Yoga at some point. Currently pixel rounding (behind experiment) is kinda broken on iOS as it does not take the point scale factor into account.

Regardless of whether or not the experiment is running, we always round the values in `YogaKit` to make sure they [correct](https://github.com/facebook/yoga/blob/master/YogaKit/UIView%2BYoga.m#L381). `YGRoundPixelValue` does take the devices scale into account.

> So we could end up with an API similar to this where pixel values have to be integers but both points and percentage values can be floating point. The point scale factor would also be taken into account when performing pixel rounding.

That sounds kinda complex to me. Wouldn't it make more sense to remove units from this library completely and generalize it? I'd rather leave the burden of points, pixels to `YogaKit` and keep the C-library agnostic of different units.

emilsjolander:
> Regardless of whether or not the experiment is running, we always round the values in YogaKit to make sure they correct. YGRoundPixelValue does take the devices scale into account.

Other platforms also have some notion of 'Points'. Moving this into Yoga would allow us to keep the rounding logic in one place.

> That sounds kinda complex to me.

I don't disagree. I would prefer if we could have an api more like this. Where the default scale factor is 1, making pixels the default but allowing users to specify another scale factor to use the value as a point (would not be part of this PR but a possible future addition).

```c
YGSetPixelScaleFactor([UIScreen mainScreen].scale)
YGNodeStyleSetWidth(node, 10.0)
YGNodeStyleSetWidth(node, YGPercent(10.0))
```

However this is not possible as we can't store bit flags on a floating point value. So the alternative then is to have any api where percentage values are specified via a different function call, which I would be fine with as well.

```c
YGNodeStyleSetWidth(node, 10.0)
YGNodeStyleSetWidthPercent(node, 10.0)
```

--

I think it boils down to having on of these two APIs. (I included the scale factor api for show but that would be a different pull request ofc).

### 1
```c
YGSetPointScaleFactor([UIScreen mainScreen].scale)
YGNodeStyleSetWidth(node, 10)
YGNodeStyleSetWidth(node, YGPoint(10.0))
YGNodeStyleSetWidth(node, YGPercent(10.0))
```

### 2
```c
YGSetPixelScaleFactor([UIScreen mainScreen].scale)
YGNodeStyleSetWidth(node, 10.0)
YGNodeStyleSetWidthPercent(node, 10.0)
```

--

I think I prefer 2 as that would allow all existing code to function as it currently does and make it clear when using percentage values. A previous version of this PR had a very similar API `YGNodeStyleSetWidthWithUnit` but in that case there would be two ways of setting pixel values which I would prefer not to have. The argument can be made that exposing a `Unit` setter would allow the addition of more units in the future without adding to the API, however I don't see us adding other units in the near future (I could of course be very wrong on this point). 

What option do you both prefer? Or would you prefer a third alternative?

dshahidehpour:
I would also prefer #2!

woehrl01:
I would still prefer to provide the unit with one function call. I think this would make the interop call much simpler. 

As we don't have to call different functions, I would not like to have a ```.width``` + ```.widthPercent``` method inside my code or do some magic trickery behind the setters. 

So if want to have an "no unit" setter and no "struct" setter, I would make a 

```
YGNodeStyleSetWidthWithUnit(node, 10.0, YGUnitPercent)
```

Not sure how you want to return the value in the getter, if there is percentage set?

you would need to guess if you need to call ```YGNodeStyleGetWidthPercent()``` or ```YGNodeStyleSetWidth()```

Calling set with a different type, where get returns another is also not very ideal.

emilsjolander:
Even with option 2 the return value of the getters would be of the current `YGValue` type, otherwise as you say you would have to guess the correct getter.

> I think this would make the interop call much simpler.

I don't think the language bindings have to match the C API exactly. The language bindings should use an API with suites that language best. in java this will probably be `setWidthPercent()` but as the .NET bindings use properties and not methods a `YGValue` type abstraction in c# would probably make most sense.

woehrl01:
> I don't think the language bindings have to match the C API exactly

I fully agree, thats the reason I would like to make the C API as generic as possible. Which leads me to your first suggestion. I agree that in Java ```setWidthPercent()``` could be a great name.

emilsjolander:
I think leaving `YGNodeStyleSetWidth(node, 10.0)` working as it does today is a good idea because
1. Most code does not use percentage values so keeping that short and concise is nice.
2. It does not break any current usages. (I know I previously mentioned this is not a huge issue. It is a nice benefit however and I have been made aware of more users of the pure C api)

So I think the only open question right now is wether to have
1. `YGNodeStyleSetWidthPercent(YGNode, float)`
2. `YGNodeStyleSetWidthWithUnit(YGNode, YGUnit, float)`

In 2, `YGNodeStyleSetWidth(YGNode, float)` would just be a shorthand for `YGNodeStyleSetWidthWithUnit(YGNode, YGUnitPixel, float)`.

I prefer `YGNodeStyleSetWidthPercent(YGNode, float)` as it is a more clear name and the main benefit of the more generic `YGNodeStyleSetWidthWithUnit(YGNode, YGUnitPixel, float)` is the ability to add more units later which I don't expect we will and if we do then we can tackle it then. I don't want to create a very generic solution where one is not needed.

As I earlier the return value of the getters will still need to be of type `YGValue`.

woehrl01:
I totally understand the thoughts behind that API. The only thing the bugs me, is that the getter and setter have different types or at least there isn't any matching method pair.

Having this kind of api, is not very self explaining, imo:
```
void YGNodeStyleSetWidth(YGNode, float)
void YGNodeStyleSetWidthPercent(YGNode, float)
YGValue YGNodeStyleGetWidth(YGNode)
```

woehrl01:
@emilsjolander @dshahidehpour By the way: if you both think this is the prefered way, I'm totally fine doing this. I just want to be some kind of devil's advocate to challenge the decision.

woehrl01:
@emilsjolander some additional ideas / thoughts:

What I like about suggestion 2:
1.  It's hard to abuse it. We don't have to check if the provided values, are a matching UnitType as we provide the unit via the function name.

1. We could hide the unit enum completely and add an additonal "UndefinedUnit", so we don't have to do some magic bit fiddling in the upper part for an "isDefined" bit.

Not sure if it's a good idea to do a typedef for ```YGValue``` because of the implicit cast possibilities. I'd prefere a struct like ```YGSize```.

emilsjolander:
> Having this kind of api, is not very self explaining, imo:

Agreed, it is not perfect. I think it is the best option now though. Let's make sure to document it!

> Not sure if it's a good idea to do a typedef for YGValue...

No reason to do any bit fiddling with this api. Keep the `YGValue` struct and `YGUnit` as is in the current state.

woehrl01:
> No reason to do any bit fiddling with this api. Keep the YGValue struct and YGUnit as is in the current state.

what's your opinion about the idea of putting Definied as a "unit"? As a undefined YGPx + undefined YGPercent doesn't really make sence, imo.

emilsjolander:
I'm fine with that 👍

woehrl01:
@emilsjolander Cool 👍 ! If there aren't any other opinions, I'd like to summarize our decision before I "get my hands dirty": 👏 

1. Change the setter into two setters -> ```SetWidth(node, float)``` + ```SetWidthPercent(node, float)```
1. Substitute the bool undefined flag, with an additonal enum type ```YGUnitUndefined```

Questions/Brainstorm:
How about the naming of the unit for pixel? As @dshahidehpour pointed out, some users won't care, should this be named differently? Maybe ```YGUnitAbsolute```, ```YGUnitRelative``` instead of pixel and percent?

emilsjolander:
Summary looks good. Again, you can leave the JNI stuff to me if you want. Regarding `YGUnitPixel` I think it is fine as is. Most usages are pixels or some form of pixel abstraction such as points. Trying to generalize to any possible unit just makes it more confusing imho.

woehrl01:
>  you can leave the JNI stuff to me if you want
That would be great, thanks!

> Trying to generalize to any possible unit just makes it more confusing 
agree 👍

.gitignore
```diff
@@ -8,6 +8,12 @@
 
 # Visual studio code
 .vscode
+*.pdb
```
emilsjolander: please mirror these changes in .hgignore

enums.py
```diff
@@ -17,6 +17,11 @@
         'LTR',
         'RTL',
     ],
+	'Unit': [
+	    'Undefined',
```
emilsjolander: indentation

emilsjolander:
> Tests still include a bunch of statements such as YGNodeStyleSetWidth(root, YGPx(100)) which should be YGNodeStyleSetWidth(root, 100) according to the previous discussions.

Disregard. I see you still have commits you are pushing 👍

enums.py
```diff
@@ -18,7 +18,7 @@
         'RTL',
     ],
 	'Unit': [
```
emilsjolander: 'Unit' is still wrongly indented compared to other enums

woehrl01: 😖 sorry, will do (all my editors are configured to use tabs instead of spaces)

emilsjolander:
benchmarks, C tests, and c# tests are failing. Once those are passing i'll import and fix the java and obj-c bits.

woehrl01:
As I don't have clang, could you please lead me to the right direction how to fix this:

```
yoga/Yoga.c:171:169: error: suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
    .border = { [YGEdgeLeft] = __builtin_nanf("0x7fc00000"), [YGEdgeTop] = __builtin_nanf("0x7fc00000"), [YGEdgeRight] = __builtin_nanf("0x7fc00000"), [YGEdgeBottom] = __builtin_nanf("0x7fc00000"), [YGEdgeStart] = __builtin_nanf("0x7fc00000"), [YGEdgeEnd] = __builtin_nanf("0x7fc00000"), [YGEdgeHorizontal] = __builtin_nanf("0x7fc00000"), [YGEdgeVertical] = __builtin_nanf("0x7fc00000"), [YGEdgeAll] = __builtin_nanf("0x7fc00000"), },

```

emilsjolander:
@woehrl01 In addition to the float extension methods could we add an implicit conversion operator from float -> pixel YGValue? https://msdn.microsoft.com/en-us/library/85w54y0a.aspx

woehrl01:
> @woehrl01 In addition to the float extension methods could we add an implicit conversion operator from float -> pixel YGValue? https://msdn.microsoft.com/en-us/library/85w54y0a.aspx

should be already there

emilsjolander:
> should be already there

Then we should be able to remove the `.Px()` from the tests right?

woehrl01:
> Then we should be able to remove the .Px() from the tests right?

true, will do. gimme a minute. 🕐

emilsjolander:
I'm happy to start importing this now. Just want to make sure you don't have anything else in the process of being committed?

woehrl01:
No additional changes 👍 if you have no further requests 😉

srhise:
This may be a dumb question, but how and when can one start to utilize this feature in RN? I'm not familiar with how yoga is versioned into RN. Amazing and very useful work here! Thank you!

emilsjolander:
@srhise once this is merged RN will be able to expose it by adding percent to their API. I have no ETA to give on this.

PR #263, opened 2016-11-27
Enable `-fPIC` for C library

passy:
At least or x86_64 linux, it appears that you need position-independent
code is required in order to link against it. Otherwise you see errors
like this:

```
/usr/bin/ld: lib_sys.rlib(CSSLayout.c.o): relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: lib_sys.rlib(CSSNodeList.c.o): relocation R_X86_64_32 against `.rodata.str1.8' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: Nonrepresentable section on output
collect2: error: ld returned 1 exit status
```

If I understand it correctly, you don't strictly need PIC for static libraries, but as soon as you want to link *other* dynamically linked library into it, it is required - which makes logical sense to me. Let me know if you want to enable this by default or leave it up to the developers to enable this if needed.

### DIFF
--- a/BUCK
+++ b/BUCK
@@ -20,7 +20,7 @@ GMOCK_OVERRIDE_FLAGS = [
   '-Wno-inconsistent-missing-override',
 ]
 
-COMPILER_FLAGS = BASE_COMPILER_FLAGS + ['-std=c11']
+COMPILER_FLAGS = BASE_COMPILER_FLAGS + ['-std=c11', '-fPIC']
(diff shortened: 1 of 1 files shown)


emilsjolander:
Any downside in enabling this flag by default?

passy:
I'm not aware of any, but I'm also a massive n00b in this area.

emilsjolander:
Reading up on this it seems legit. Importing. thanks!

PR #273, opened 2016-12-07
Use transparent border to avoid search box moving

zpao:
cc @JoelMarcey

I think the transition change is probably fine. The border color change looked strange without it, but it also felt weird to add another property.


JoelMarcey:
I trust you 👍

PR #274, opened 2016-12-07
Use https when we can.

zpao:
Not sure if the _config one really matters but the other is in the nav and makes sure we avoid a redirect (GitHub is all https anyway).

There are a few others but it looks like those were just copypasta and don't actually get used, so I skipped them.

cc @JoelMarcey


JoelMarcey:
Cool. Thanks. Both look good. I enforce `https` through GitHub pages, but this is a good thing to do a well.

PR #275, opened 2016-12-07
Fix href to Cocoapods

rupak:
Replace https://gradle.org/ with https://cocoapods.org/


JoelMarcey:
Oh. Nice catch. Thanks!

PR #291, opened 2016-12-19
Add yg_isLeaf readonly property.

### LINKED ISSUE #290: YogaKit does not correctly account for composite UIView when doing layout
I've been trying to render an instance of `MKMapView` but failed to do so.

Here is a code that works **without** using Yoga

```swift
import UIKit
import MapKit

class ViewController: UIViewController {

    override func viewDidLoad() {
        super.viewDidLoad()
        let m = MKMapView()
        m.frame.size = CGSize(width: self.view.frame.size.width, height: 100)
        self.view.addSubview(m)
    }

}
```

which renders the following view

simulator screen shot dec 19 2016 1 24 12 am

but when I try to render a same map view using Yoga with the following code

```swift
import UIKit
import MapKit

class ViewController: UIViewController {

    override func viewDidLoad() {
        super.viewDidLoad()
        let m = MKMapView()

        self.view.yg_usesYoga = true
        self.view.yg_setWidth(self.view.bounds.size.width)
        self.view.yg_setHeight(self.view.bounds.size.height)
        
        m.yg_usesYoga = true
        m.yg_setWidth(self.view.bounds.size.width)
        m.yg_setHeight(100)
        
        self.view.addSubview(m)
        self.view.yg_applyLayout()
    }

}
```

I get this 

simulator screen shot dec 19 2016 1 26 10 am

Which is basically a "white" map

guidomb:
This fixes #290. It is not clear to me if there a programatic way of deciding if a view is a "composite" view or not. Because if there is that would be much better than having to add a category for each view that is a composite view overwriting the `yg_isLeaf` method.

### DIFF
--- a/YogaKit/UIView+Yoga.h
+++ b/YogaKit/UIView+Yoga.h
@@ -21,6 +21,7 @@
  The property that decides during layout/sizing whether or not yg_* properties should be applied. Defaults to NO.
  */
 @property (nonatomic, readwrite, assign, setter=yg_setUsesYoga:) BOOL yg_usesYoga;
+@property (nonatomic, readonly) BOOL yg_isLeaf;
 
 - (void)yg_setDirection:(YGDirection)direction;
 - (void)yg_setFlexDirection:(YGFlexDirection)flexDirection;
--- a/YogaKit/UIView+Yoga.h
+++ b/YogaKit/UIView+Yoga.h
@@ -21,7 +21,7 @@
  The property that decides during layout/sizing whether or not yg_* properties should be applied. Defaults to NO.
  */
 @property (nonatomic, readwrite, assign, setter=yg_setUsesYoga:) BOOL yg_usesYoga;
-@property (nonatomic, readonly) BOOL yg_isLeaf;
+@property (nonatomic, readwrite, assign, setter=yg_setLeaf:) BOOL yg_isLeaf;
 
 - (void)yg_setDirection:(YGDirection)direction;
 - (void)yg_setFlexDirection:(YGFlexDirection)flexDirection;
--- a/YogaKit/UIView+Yoga.m
+++ b/YogaKit/UIView+Yoga.m
@@ -54,7 +54,7 @@
 
 - (NSUInteger)yg_numberOfChildren
 {
-  return YGNodeChildCount([self ygNode]);
+  return YGNodeGetChildCount([self ygNode]);
 }
 
 #pragma mark - Setters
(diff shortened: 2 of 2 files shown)


emilsjolander:
- I don't think Yoga should implement these categories as it does not scale well, leave that to application developers instead.
- Can you make `yg_isLeaf` readwrite? This would allow it to be used without adding a category.

Hopefully we can solve this in a better way in the future removing the need for this property.

hartbit:
> I don't think Yoga should implement these categories as it does not scale well, leave that to application developers instead.

Why do you think Yoga should not implement those? Isn't it worse to have everybody hit the MKMapView *bug* and have to implement their own category?

emilsjolander:
Ideally I don't think Yoga should implement this as it is not really scalable. However I do see your point. Let's keep it for now, I hope we can find a better solution soon. Can you also add UIButton as i've seen others have the same issue with that?

Also I still think having `yg_isLeaf` as a setter is good as it would give anyone the opportunity to quickly fix this themselves for whatever view they are using without the need for a category.

guidomb:
@emilsjolander I do agree with @hartbit on avoiding other users to hit this bug. I understand that when you say this solution "does not scale" you mean that it is unclear how many components are subject to this bug and on each case a category needs to be implemented. In such case I understand the maintenance burden but I think it is a small price to pay to improve user experience (I am clearly biased when I say that it is small price because I am not the maintainer of the project).

I'll add the setter and the category to `UIButton` although I haven't experience this bug with that class. @emilsjolander I think it would be better to have a separate file to put this "hacks". Maybe `UIKitCompositeViewFix.m` and maybe explain this in the documentation. What do you think?

@emilsjolander Could you explain why yoga fails to detect views like `MKMapView` as leafs? What is the difference between a composite view and a "regular" view that has subviews that were all laid out using Yoga?

emilsjolander:
> Maybe UIKitCompositeViewFix.m and maybe explain this in the documentation. What do you think?

Yes pease! :+1:

> Could you explain why yoga fails to detect views like MKMapView as leafs? What is the difference between a composite view and a "regular" view that has subviews that were all laid out using Yoga?

Because we don't require children of a yoga container to be marked as using yoga for them to be added to the yoga layout process. This is probably a mistake. If we required children to be marked as using yoga we would not have this problem at all.

guidomb:
> Because we don't require children of a yoga container to be marked as using yoga for them to be added to the yoga layout process. This is probably a mistake. If we required children to be marked as using yoga we would not have this problem at all.

@emilsjolander I am probably missing some important detail but would it be better if instead of doing this patch we make the `yg_usesYoga` method recursively call itself on all its subviews?

dshahidehpour:
I left a comment and some code in #290. A fix is on the way!

PR #292, opened 2016-12-19
Transform the Count enum values into private constants

hartbit:
Hides implementation details for the C, Objective-C and Swift APIs.


emilsjolander:
Awesome! Will merge tomorrow

emilsjolander:
Changing this a bit before landing. Some RN stuff was depending on the count so I moved the defines into YGEnums.h

Issue #295, opened 2016-12-21
C# wrapper can crash


itsff:
It looks like the C# wrapper is not following "best practices" of calling native code. I think you are just getting lucky that the code is not crashing on you (likely since the reference is long-lived).

Please use **GC.KeepAlive()** to ensure the GC doesn't finalize YodaNode object while you are within a native function call. Some background:
https://goo.gl/WeL3H4
https://msdn.microsoft.com/en-us/library/ms182300.aspx
https://blogs.msdn.microsoft.com/oldnewthing/20100813-00/?p=13153

I would also suggest creating a light-weight "handle" class whose sole purpose would be holding the native IntPtr and finalization of it. It is a poor practice to have a finalizable object that's holding many managed references. Why? Because now they cannot be collected until the owning object has fully died (as they could potentially be needed by its finalizer method). Literature and a much deeper explanation is available in this [excellent book by Marcus Heege](https://books.google.com/books?id=wZoQyVi5f60C&lpg=PA266&ots=Eh2Xb7EeuD&dq=marcus%20heege%20graph%20promotion%20problem&pg=PA266#v=onepage&q=marcus%20heege%20graph%20promotion%20problem&f=false).

splhack:
@itsff Can you share a concrete example of the scenario that we need to call `GC.KeepAlive()`? I think all managed objects already live longer than the associated unmanaged objects.

woehrl01:
@splhack I just read through most of the provided links and did some research. It seems the following is true if you run this outside of visual studio

If if you are inside a object method the finalizer can be run:

```csharp
YogaNode node = new YogaNode();
node.Width = 10;
node.CalculateLayout();
```

That means, it could be that the node gets finalized inside the CalculateLayout() function.
Which means the memory gets freed while the calculation runs.

The suggest fix is to use ```SafeHandle``` instead of ```IntPtr```as the interop marshaler handles this in a special way to ensure that this doesn't get freed inside an interop call.

I think the following should work, do you agree @itsff ?

```csharp
    internal class YogaNodeHandle : SafeHandleZeroOrMinusOneIsInvalid
    {
        private YogaNodeHandle()
            : base(true)
        {
        }

        [ReliabilityContract(Consistency.WillNotCorruptState, Cer.MayFail)]
        protected override bool ReleaseHandle()
        {
            Native.YGNodeFree(handle);
            return true;
        }

    }
```

and change the ```Native.cs``` like, e.g.:
```csharp
[...]

PR #297, opened 2016-12-23
Switched to using SafeHandle for native code interrupt

itsff:
This PR fixes issue #295 

- Created a new internal YGNodeHandle class extending SafeHandle.
- YogaNode now stores a reference to YGNodeHandle.
- Removed finalizer from YogaNode.
- Pulling in System.Runtime.Handles 4.3.0.

### DIFF
--- a/csharp/Facebook.Yoga/Native.cs
+++ b/csharp/Facebook.Yoga/Native.cs
@@ -20,18 +20,40 @@ namespace Facebook.Yoga
         private const string DllName = "yoga";
 #endif
 
+        internal class YGNodeHandle : SafeHandle
+        {
+            private YGNodeHandle() : base(IntPtr.Zero, true)
+            {
+            }
+
+            public override bool IsInvalid
+            {
+                get
+                {
+                    return this.handle == IntPtr.Zero;
+                }
+            }
+
+            protected override bool ReleaseHandle()
+            {
+                Native.YGNodeFree(this.handle);
+                GC.KeepAlive(this);
+                return true;
+            }
+        }
+
         [DllImport(DllName)]
         public static extern void YGInteropSetLogger(
             [MarshalAs(UnmanagedType.FunctionPtr)] YogaLogger.Func func);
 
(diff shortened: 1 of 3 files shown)


csharp/Facebook.Yoga/Native.cs
```diff
@@ -46,237 +68,237 @@ internal static class Native
...
-        public static extern void YGNodeInsertChild(IntPtr node, IntPtr child, uint index);
+        public static extern void YGNodeInsertChild(YGNodeHandle node, YGNodeHandle child, uint index);
 
         [DllImport(DllName)]
-        public static extern void YGNodeRemoveChild(IntPtr node, IntPtr child);
+        public static extern void YGNodeRemoveChild(YGNodeHandle node, YGNodeHandle child);
 
         [DllImport(DllName)]
-        public static extern IntPtr YGNodeGetChild(IntPtr node, uint index);
+        public static extern YGNodeHandle YGNodeGetChild(YGNodeHandle node, uint index);
```
splhack: I'll remove this declaration in merge commit since managed code doesn't use it

itsff: You guys beat me to it. Thank you for noticing this one (@woehrl01)  and removing it (@splhack). It could have caused some confusion and lifetime problems. We should similarly remove `YGNodeSetContext` and `YGNodeSetContext` since this functionality is achieved by the `YogaNode::Data` property.

csharp/Facebook.Yoga/Native.cs
```diff
@@ -20,18 +20,40 @@ internal static class Native
...
+                get
+                {
+                    return this.handle == IntPtr.Zero;
+                }
+            }
+
+            protected override bool ReleaseHandle()
+            {
+                Native.YGNodeFree(this.handle);
+                GC.KeepAlive(this);
```
itsff: Likely not necessary, but I wasn't sure of the implementation of the parent class. Shouldn't hurt.

PR #299, opened 2016-12-23
WIP: JavaScript port via emscripten

mikearmstrong001:
Putting initial PR up for feedback

built using `emcc (Emscripten gcc/clang-like replacement) 1.36.4`

Notes:
- Passing all text with exclusion of memory tests.
- Handling of context is a little different as objects can't be passed to the `YGNodeSetContext`
- Currently tests are hand ported from cpp as getting test generation was proving too time consuming. Script generation would be considered an integral part of a full PR.
- works with JavaScript based measure callbacks, see tests

To test, point browser at JavaScript/tests/index.html and view the console.log for errors

Next Steps:
- remove the _YG
- export as module
- Script generation of tests

Feedback welcome


emilsjolander:
I am super excited about this and I know many people in the community are as well! This is great. 

> Script generation would be considered an integral part of a full PR.

Agreed! @splhack and I have written much of the test generation. We will be happy to help get this up and running.

Generally this looks very good. Some few things I would like us to look into apart from the next steps you have already mentioned.

- Move JavaScript/build/build_js.sh over to use buck. Mostly so that we have a general interface for building things in Yoga. I can help with this. I don't consider this a blocker, this could be done in a follow up PR.
- Look into a nicer way to list the `EXPORTED_FUNCTIONS`. Maintaining a separate list of exported symbols is OK (we do this for java JNI already) but it would be preferable not to require this. I don't consider this a blocker though.

splhack:
maybe https://github.com/google/gjstest/ is a good option for test? Travis-CI can use this with `brew install gjstest`. I can help to make gentest-js.js :)

emilsjolander:
https://facebook.github.io/jest is of course also a good alternative. @cpojer could probably help here.

emilsjolander:
Add the code from #301 into this PR. I think it makes more sense to be a single pull request. Then also all the js tests to the travis config

mikearmstrong001:
Now exporting as a module and removed _YG, this will break the gentests-js @splhack put together so quickly! sorry

From an end user perspective this seems much more pleasant for an API

emilsjolander:
@mikearmstrong001 yeah, that's why it makes more sense to have the gentest additions as part of this PR 👍

emilsjolander:
@mikearmstrong001 Can we use https://facebook.github.io/jest and node for testing instead of doing the testing in the browser? Will be much easier to hook up to CI.

mikearmstrong001:
browser is useful for debugging so want to support both, which this should via `var Yoga = Yoga || require('../src/Yoga');`

Currently, Jest is mocking my require('../src/Yoga') but I'm probably doing something wrong because I'm just getting short work periods of time on this. If I turn automock off I get an out of memory error. Any quick thoughts?

emilsjolander:
No idea. I've never used jest actually. @cpojer or @vjeux might know

emilsjolander:
So the things left to do now are
- Move JavaScript/build/build_js.sh into buck
- Run js tests outside of browser (preferably via buck) on travis

Am I missing anything?

vjeux:
Jest doesn't run in browser, only in node.

Automocking has been disabled by default. Are you using the latest version?

I'm in Christmas mode with my family for the next couple of days so can't guarantee to be able to take a look.

emilsjolander:
@vjeux No rush! thanks. I would prefer to run these tests in node so that's perfect.

emilsjolander:
Also is there a reason for having the API as it is now?

```js
var root = Yoga.nodeNew();
Yoga.nodeStyleSetWidth(root, 100);
Yoga.nodeStyleSetHeight(root, 100);
```

instead of something like this

```js
var root = new YogaNode();
root.setWidth(100);
root.setHeight(100);
```

Second option seems more javascript like.

splhack:
tried to migrate some browser unittest into jest
https://gist.github.com/splhack/4fcfb2b25c07755a9853ef6784804c48
This works well with `jest` cli.

btw apparently we can use Chrome Developer Tools with this http://facebook.github.io/jest/docs/troubleshooting.html#tests-are-failing-and-you-don-t-know-why

vjeux:
Hmm, this seems to be working for me.

screen shot 2016-12-24 at 6 09 49 pm

emilsjolander:
As long as we can run the test suite outside the browser i'm happy. I know
very little about this javascript stuff so i'll trust you on this.

mikearmstrong001:
@emilsjolander to be clear I'm expecting to run the tests outside the browser (jest) as well as within (non Jest).
@vjeux unblocked me with his screen shot, I had an old version of jest. Thank you!

I just need to run through the manually constructed tests and then I'll consider this a non WIP PR.

I can move to something like
```
var root = new Yoga.Node();
root.setWidth(100);
root.setHeight(100);
```

My only concern was writing an api around an api would increase the change requirements as you add new APIs, if this is a non issue for you to maintain then that is cool by me. Whatever we choose does need to hide behind a module.

emilsjolander:
> to be clear I'm expecting to run the tests outside the browser (jest) as well as within (non Jest).

👍 

> I just need to run through the manually constructed tests and then I'll consider this a non WIP PR.

👍 

>My only concern was writing an api around an api would increase the change requirements as you add new APIs, if this is a non issue for you to maintain then that is cool by me. Whatever we choose does need to hide behind a module.

This ads a small burden to API changes but nothing which we don't already have for other language bindings. We also try not to make API breaking changes too often. It makes the API much nicer to use in my opinion.

mikearmstrong001:
> This ads a small burden to API changes but nothing which we don't already have for other language bindings. We also try not to make API breaking changes too often. It makes the API much nicer to use in my opinion.

Excellent, I agree and as you're not concerned I'm very happy with this direction

cpojer:
yay jest 🎉

Issue #296, opened 2016-12-22
Math.floor operation on MeasureOutput for Java and C#


rozele:
I just wanted to check if this behavior was by-design before I made any PR.  In react-native-windows, we have to do Math.Ceiling on all of our measure function outputs because casting from float/double to int always does a floor operation:

https://github.com/facebook/yoga/blob/master/csharp/Facebook.Yoga/MeasureOutput.cs#L16
https://github.com/facebook/yoga/blob/master/java/com/facebook/yoga/YogaMeasureOutput.java#L18

I imagine this isn't a problem for react-native on Android because they don't encounter fractional pixel dimensions?

woehrl01:
Nice catch! ButI think you should not use any rounding logic as those values are treated as float in the c part.

For Java you should instead use:
```java
  public static long make(float width, float height) {
    return make(Float.floatToIntBits(width), Float.floatToIntBits(height));
  }
```

In C#:
```csharp
        public static long Make(double width, double height)
        {
            int widthInt = BitConverter.ToInt32(BitConverter.GetBytes((float)width), 0);
            int heightInt = BitConverter.ToInt32(BitConverter.GetBytes((float)height), 0);
            return Make((widthInt , heightInt);
        }
```

As this is a little cumbersome on C# maybe its better to use directly a struct with two floats (see [YGSize](https://github.com/facebook/yoga/blob/master/yoga/Yoga.h#L36))

This additionally prevents us from a bug if we calculate a size bigger than 16777216 as this would overflow into the exponent part.

woehrl01:
@emilsjolander Java still needs to be changed/fixed.

emilsjolander:
@woehrl01 putting up a diff now (internally). Thanks for noticing!

emilsjolander:
@woehrl01 putting up an internal diff for java now. thanks!

emilsjolander:
https://github.com/facebook/yoga/commit/352f592767b148747da2d5822f0f4de17eb7b9c2

PR #300, opened 2016-12-23
C# use YogaSize throughout the measure call chain

woehrl01:
#296 hint me to taking a deeper look. We already have ```YogaSize``` struct in c# an convert into a ```long``` into the meantime. We loose some precission here and could get a wrong result with very hugh values.

Java still has the same problem (see #296)

Also there is no need to have double in the method signature as we only use float internally, which could confuse some.

### DIFF
--- a/csharp/Facebook.Yoga/MeasureFunction.cs
+++ b/csharp/Facebook.Yoga/MeasureFunction.cs
@@ -9,7 +9,7 @@
 
 namespace Facebook.Yoga
 {
-    public delegate long MeasureFunction(
+    public delegate YogaSize MeasureFunction(
         YogaNode node,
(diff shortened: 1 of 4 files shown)


emilsjolander:
I don't know much about P/Invoke performance. For java I did the bit fiddling for performance reasons as passing primitives is much more performant than creating and passing objects. If you don't think this will be an issue then I am happy to merge this once tests are passing.

woehrl01:
The interop already uses a struct, the method I changed gets called purely inside c#. It creates the struct directly instead of going through the long like in c#. This won't degrade the performance here.

emilsjolander:
Awesome! Merging :) thanks!

Issue #215, opened 2016-08-18
JavaScript variant planned?


queequac:
Just some question: In the past css-layout was written in JavaScript and transpiled to C.
Now C seems to be the primary language the library is written in.

Are there any plans to do the same vice-versa again, transpiling this time to JavaScript?
Otherwise I am stuck to the deprecated version, while it was the big advantange of css-layout having really identical behavior in both languages.

emilsjolander:
Let me start with some backstory on the change. The js version was not used by us internally and working with the regex transpilation process slowed us down. The libraries focus was to bring flexbox to mobile and in reaching that goal js is not important. That said I know of many reasons why a js version is important and would be useful so if we could solve that with little to no development overhead I would be all for it. One way I can think of doing this is via emscripten (http://kripken.github.io/emscripten-site/index.html). 

I won't personally be working on this in the near future as there are more pressing issues (for us) to solve. However I would love community contribution here.

PR #304, opened 2016-12-26
Adds Javascript Support

arcanis:
- As mentioned in the title, this PR adds Javascript support to Yoga. Two different builds are included in this PR thanks to [nbind](https://github.com/charto/nbind), which conveniently allow to target both Node.js' native addons and browser environments via asmjs with approximately the same codebase. That should solve #215.

- All tests successfully pass on both codepaths. You can run `yarn test:all` inside the `javascript` directory to test it.

- Because of a bug in nbind, the [following PR](https://github.com/charto/nbind/pull/57) needs to be merged and a new version released before this one can be safely merged as well.

- I had to use `double` types instead of `float` in the C++ bindings because of an Emscripten [bug](https://github.com/kripken/emscripten/issues/3592) where symbols aren't correctly exported when using floats.

- There's some tweaks to do before this PR is 100% ready to merge, but I wanted to have your opinion first. What do you think of this?

--- 

To do:

- [x] Ensure that Travis run
- [x] Add missing functions
- [x] Check measurement functions
- [x] Recursive frees
- [x] Make a decision about the memory management


emilsjolander:
https://github.com/facebook/yoga/pull/299 Also adds javascript support via emscrimten and is nearly ready to be merged. Could you collaborate with @mikearmstrong001 on adding anything that might be missing from that PR?

arcanis:
Aww I completely missed it, sorry @mikearmstrong001 :(

Looking at it, I'm not sure anything is missing per-se, but there's two major differences between our PRs:

- #299 doesn't export a native Node.js module. I think it's the main difference, and maybe the most difficult one to implement since it requires either a lot of glue code, or a dedicated build chain. That's why I chose nbind, which nicely abstract the interfacing of both Node.js addons and asmjs modules under the same umbrella.

- This PR uses an OOP API which might be a bit more idiomatic than free functions, at least in Javascript. That's kind of a philosophical choice, so there's really nothing to fix on #299.

Fixing both of these issues would probably require a PR very similar to mine, so I guess you'll have to choose between those two approaches.

emilsjolander:
> #299 doesn't export a native Node.js module. I think it's the main difference, and maybe the most difficult one to implement since it requires either a lot of glue code, or a dedicated build chain. That's why I chose nbind, which nicely abstract the interfacing of both Node.js addons and asmjs modules under the same umbrella.

Sounds cool, I don't know anything about this. @mattpodwysocki was looking at building node.js bindings. Wonder if he has any comments.

> This PR uses an OOP API which might be a bit more idiomatic than free functions, at least in Javascript. That's kind of a philosophical choice, so there's really nothing to "fix" on #299.

If you read the comments in the other pull request you can see that this is the way we were going with it 👍 

> "Fixing" both of these issues to match the behaviour in mine would probably end up in a PR very similar to mine, so I guess you'll have to choose which one of those two approaches you prefer.

Agreed, we will most likely end up just merging one of these. Would love some input from people who know more than me about this. cc: @mikearmstrong001 @vjeux

vjeux:
I'm not really familiar with the c -> js options.

The only thing I know is that it is very annoying when a node dependency has a native module as any time you npm install it it takes forever to actually build it. But, on the other hand, it's probably unfortunate to lose the performance when you can actually run C code.

Anyway, I don't have much to add here.

emilsjolander:
@arcanis I'll await comments @mikearmstrong001 on using nbind before conducting a code review.

mikearmstrong001:
As @emilsjolander and I discussed: I actually have moved to and arrived at the same interface through a different process so the lack of a native implementation is the difference.

We should go with the solution that solves the most problems with the minimum effort! which seems to be this PR, as long as it works for day to day maintenance, handles running on a browser as is not ill performant I'm happy.

I'll check my requirements and get back very soon. Nice work @arcanis I've not used or looked at nbind before.

emilsjolander:
@mikearmstrong001 awesome! Would be super helpful if you can help check that this meets all the requirements you have.

@arcanis Can you start by making this PR run the JS tests in travis?

arcanis:
@mikearmstrong001 great ! thanks !

@emilsjolander sure - native tests should be easy enough to run, asmjs not that much (the main issue I had when I tried this task previously is that building the asmjs version requires emscripten, which in turn requires to build llvm, which is too heavy for Travis CI). Would that work?

I'm currently working on implementing measureFunc and it should be mostly done. However, it seems there is no automatic measurement test for non-cpp languages, is that correct?

emilsjolander:
@arcanis I would like us to run the javascript tests in CI. In #299 the output of emscripten was checked in. Can you do similar here? Or can we pre-build an emscripten binary for travis?

> I'm currently working on implementing measureFunc and it should be mostly done. However, it seems there is no automatic measurement test for non-cpp languages, is that correct?

Not sure what you mean, could you explain?

mikearmstrong001:
measureFun: @arcanis that is correct, a

arcanis:
> @arcanis I would like us to run the javascript tests in CI. In #299 the output of emscripten was checked in. Can you do similar here? Or can we pre-build an emscripten binary for travis?

Hm, I think you're right, commiting the pre-built asmjs file to the repository might be good enough. Only issue I see is that it will not correctly detect modifications if a contributor forget to rebuild it before pushing.

@mikearmstrong001 Good idea, I'll use your benchmark !

emilsjolander:
@arcanis Yeah that is the downside. Can you look into getting a pre-build emscripten binary running on travis. That would be awesome.

mikearmstrong001:
@arcanis how are you handling the free functions? they seem to be missing from the generated tests, we should confirm these work as expected. My use case will certainly be hitting these.

I'd also recommend YogaNodeTest in c# this is the final test I was working through.

arcanis:
@mikearmstrong001 Which functions do you specifically need? I've currently exported `setExperimentalFeatureEnabled` and `isExperimentalFeatureEnabled` ([here](https://github.com/facebook/yoga/pull/304/files#diff-19ff424c8a54c1d5951306a85d6e3656R11)), and exporting the others shouldn't take too long.

mikearmstrong001:
The reason I was asking about free is that I need to ensure the the `YGNodeRef m_node` is correctly deallocated. I'm not familiar on how nbind will handle destructors within JS based code.

Having something in the PR I can run in a browser would be very helpful.

These will need to be exposed and there may be others I've missed.

```
WIN_EXPORT void YGNodeFree(const YGNodeRef node);
WIN_EXPORT void YGNodeFreeRecursive(const YGNodeRef node);
WIN_EXPORT void YGNodeReset(const YGNodeRef node);
WIN_EXPORT void YGNodeMarkDirty(const YGNodeRef node);
WIN_EXPORT bool YGNodeIsDirty(const YGNodeRef node);
WIN_EXPORT bool YGNodeCanUseCachedMeasurement(const YGMeasureMode widthMode,
WIN_EXPORT void YGNodeCopyStyle(const YGNodeRef dstNode, const YGNodeRef srcNode);
YG_NODE_STYLE_PROPERTY(float, AspectRatio, aspectRatio);
WIN_EXPORT void YGNodeStyleSetFlex(const YGNodeRef node, const float flex);
YG_NODE_PROPERTY(bool, HasNewLayout, hasNewLayout);
```

By looking at the handwritten test functions eg YGStyleTest these would be actively hit.

arcanis:
I'll check this. According to https://github.com/charto/nbind/issues/34, destructors should be called by the garbage collector in native modules, and by calling the `node.free()` function inside browsers.

> Having something in the PR I can run in a browser would be very helpful.

If you have `emcc` in your PATH, you should be able to run `yarn build:asmjs` and use the module with bundlers such as webpack (you can use a symlink from the `node_modules/yoga-layout` to the `javascript` directory to make it easier).

mikearmstrong001:
@emilsjolander and @arcanis for easy of adoption I think prebuilt JS library should be included. Building a JS solution via emcc is going to be a chore for anyone just looking to use the lib.

Can we run tests on a travis generated build and also against the prebuilt solution which we should actively encourage to be unloaded as part of a PR. I'm sure we could look at automating the generation.

arcanis:
The option I went for with https://github.com/manaflair/text-layout was to ship the asmjs file but only on npm, not github, simply by adding a `.npmignore` with different rules than the `.gitignore` (the asmjs file is ignored by git, but not by npm), so that there's only the contributor who types `npm publish` who needs to have emscripten in its PATH.

mikearmstrong001:
@arcanis the current build seems to be ignoring the 

```
            "cflags": [
                "-s", "PRECISE_F32=1"
            ],

            "ldflags": [
                "--memory-init-file", "0",
                "-s", "TOTAL_MEMORY=134217728",
                "-s", "PRECISE_F32=1"
            ]
```

found via `--verbose` on the build command line 

I definitely would like to get like to get a build step in place to prepare a bundle for use in browser as part of this PR or a quick follow up.

mikearmstrong001:
also to confirm, I have a built the js and run some of the tests in the browser. I will be able to spend a bit more time on this tomorrow.

arcanis:
> I definitely would like to get like to get a build step in place to prepare a bundle for use in browser as part of this PR or a quick follow up.

Just to be sure, do you mean exporting a ready-to-use script that would expose a global, or would an npm module (with both main & browser) fields be enough? This PR should already cover the second case, but there might be more work required to support the second. It would probably require something like browserify.

mikearmstrong001:
It's both cases, the second is definitely covered. First shouldn't block this PR, I'm currently using webpack so can contribute this or another method. Is your focus node ?

Also happy to contribute some of the manually written tests.

Is the main blocker now the asmjs changes required in nbind, i.e. Your PR you mentioned in your first comment. Any thoughts on the missing ldflags? An additional nbind issue? FYI these missing flags mean the external mem init file is written out from emscripten when optimizations are enabled - not ideal for this case as the file is very small and the additional network request would probably be more expensive

arcanis:
> Is the main blocker now the asmjs changes required in nbind, i.e. Your PR you mentioned in your first comment?

Yep. The maintainer checked them, and should release a new version by tomorrow, so it shouldn't block for long. Travis testing has been a hard beast to tame, but once nbind is updated it should only be a matter of time.

However, the issue you have with ldflags is strange as I can't reproduce them locally: my build doesn't generate mem files, and every flag appears in the command line. What's the exact output of running `npm run build:asmjs && (cd build && rm -rf Release && V=1 make)` ?

mikearmstrong001:
Great news on nbind.

Hope this helps.

```
  COPY Release/obj.target/nbind/geni/pre.js
  COPY Release/obj.target/nbind/geni/post.js
  COPY Release/obj.target/nbind/geni/em-api.js
  CC(target) Release/obj.target/nbind/sources/yoga/YGNodeList.o
  CC(target) Release/obj.target/nbind/sources/yoga/Yoga.o
  CXX(target) Release/obj.target/nbind/sources/Node.o
  CXX(target) Release/obj.target/nbind/sources/global.o
  CXX(target) Release/obj.target/nbind/sources/nbind.o
  CXX(target) Release/obj.target/nbind/node_modules/nbind/src/common.o
  CXX(target) Release/obj.target/nbind/node_modules/nbind/src/reflect.o
  CXX(target) Release/obj.target/nbind/node_modules/nbind/src/em/Binding.o
  LINK(target) Release/nbind.js
  rm -rf "Release/obj.target/nbind/geni/pre.js" && cp -af "../node_modules/nbind/src/pre.js" "Release/obj.target/nbind/geni/pre.js"
  rm -rf "Release/obj.target/nbind/geni/post.js" && cp -af "../node_modules/nbind/src/post.js" "Release/obj.target/nbind/geni/post.js"
  rm -rf "Release/obj.target/nbind/geni/em-api.js" && cp -af "../node_modules/nbind/dist/em-api.js" "Release/obj.target/nbind/geni/em-api.js"
  $HOME/git/yoga/javascript/node_modules/nbind/bin/emcc '-DNODE_GYP_MODULE_NAME=nbind' '-D_DARWIN_USE_64_BIT_INODE=1' '-D_LARGEFILE_SOURCE' '-D_FILE_OFFSET_BITS=64' -I$HOME/.node-gyp/5.3.0/include/node -I$HOME/.node-gyp/5.3.0/src -I$HOME/.node-gyp/5.3.0/deps/uv/include -I$HOME/.node-gyp/5.3.0/deps/v8/include -I../sources -I../node_modules/nan -I../node_modules/nbind/include  -Os -mmacosx-version-min=10.5 -arch x86_64 -Wall -Wendif-labels -W -Wno-unused-parameter -fno-strict-aliasing -O3 --pre-js ./Release/obj.target/nbind/geni/pre.js --post-js ./Release/obj.target/nbind/geni/post.js --js-library ./Release/obj.target/nbind/geni/em-api.js -s NO_FILESYSTEM=1 -s "EXPORTED_FUNCTIONS=[\"_nbind_init\",\"_nbind_value\"]" -s "DEFAULT_LIBRARY_FUNCS_TO_INCLUDE=[\"nbind_value\",\"\$Browser\"]" -MMD -MF ./Release/.deps/Release/obj.target/nbind/sources/yoga/YGNodeList.o.d.raw   -c -o Release/obj.target/nbind/sources/yoga/YGNodeList.o ../sources/yoga/YGNodeList.c
  $HOME/git/yoga/javascript/node_modules/nbind/bin/emcc '-DNODE_GYP_MODULE_NAME=nbind' '-D_DARWIN_USE_64_BIT_INODE=1' '-D_LARGEFILE_SOURCE' '-D_FILE_OFFSET_BITS=64' -I$HOME/.node-gyp/5.3.0/include/node -I$HOME/.node-gyp/5.3.0/src -I$HOME/.node-gyp/5.3.0/deps/uv/include -I$HOME/.node-gyp/5.3.0/deps/v8/include -I../sources -I../node_modules/nan -I../node_modules/nbind/include  -Os -mmacosx-version-min=10.5 -arch x86_64 -Wall -Wendif-labels -W -Wno-unused-parameter -fno-strict-aliasing -O3 --pre-js ./Release/obj.target/nbind/geni/pre.js --post-js ./Release/obj.target/nbind/geni/post.js --js-library ./Release/obj.target/nbind/geni/em-api.js -s NO_FILESYSTEM=1 -s "EXPORTED_FUNCTIONS=[\"_nbind_init\",\"_nbind_value\"]" -s "DEFAULT_LIBRARY_FUNCS_TO_INCLUDE=[\"nbind_value\",\"\$Browser\"]" -MMD -MF ./Release/.deps/Release/obj.target/nbind/sources/yoga/Yoga.o.d.raw   -c -o Release/obj.target/nbind/sources/yoga/Yoga.o ../sources/yoga/Yoga.c
  $HOME/git/yoga/javascript/node_modules/nbind/bin/emcc '-DNODE_GYP_MODULE_NAME=nbind' '-D_DARWIN_USE_64_BIT_INODE=1' '-D_LARGEFILE_SOURCE' '-D_FILE_OFFSET_BITS=64' -I$HOME/.node-gyp/5.3.0/include/node -I$HOME/.node-gyp/5.3.0/src -I$HOME/.node-gyp/5.3.0/deps/uv/include -I$HOME/.node-gyp/5.3.0/deps/v8/include -I../sources -I../node_modules/nan -I../node_modules/nbind/include  -Os -mmacosx-version-min=10.5 -arch x86_64 -Wall -Wendif-labels -W -Wno-unused-parameter -std=gnu++0x -fno-rtti -fno-exceptions -fno-threadsafe-statics -fno-strict-aliasing -O3 --pre-js ./Release/obj.target/nbind/geni/pre.js --post-js ./Release/obj.target/nbind/geni/post.js --js-library ./Release/obj.target/nbind/geni/em-api.js -s NO_FILESYSTEM=1 -s "EXPORTED_FUNCTIONS=[\"_nbind_init\",\"_nbind_value\"]" -s "DEFAULT_LIBRARY_FUNCS_TO_INCLUDE=[\"nbind_value\",\"\$Browser\"]" -MMD -MF ./Release/.deps/Release/obj.target/nbind/sources/Node.o.d.raw   -c -o Release/obj.target/nbind/sources/Node.o ../sources/Node.cc
  $HOME/git/yoga/javascript/node_modules/nbind/bin/emcc '-DNODE_GYP_MODULE_NAME=nbind' '-D_DARWIN_USE_64_BIT_INODE=1' '-D_LARGEFILE_SOURCE' '-D_FILE_OFFSET_BITS=64' -I$HOME/.node-gyp/5.3.0/include/node -I$HOME/.node-gyp/5.3.0/src -I$HOME/.node-gyp/5.3.0/deps/uv/include -I$HOME/.node-gyp/5.3.0/deps/v8/include -I../sources -I../node_modules/nan -I../node_modules/nbind/include  -Os -mmacosx-version-min=10.5 -arch x86_64 -Wall -Wendif-labels -W -Wno-unused-parameter -std=gnu++0x -fno-rtti -fno-exceptions -fno-threadsafe-statics -fno-strict-aliasing -O3 --pre-js ./Release/obj.target/nbind/geni/pre.js --post-js ./Release/obj.target/nbind/geni/post.js --js-library ./Release/obj.target/nbind/geni/em-api.js -s NO_FILESYSTEM=1 -s "EXPORTED_FUNCTIONS=[\"_nbind_init\",\"_nbind_value\"]" -s "DEFAULT_LIBRARY_FUNCS_TO_INCLUDE=[\"nbind_value\",\"\$Browser\"]" -MMD -MF ./Release/.deps/Release/obj.target/nbind/sources/global.o.d.raw   -c -o Release/obj.target/nbind/sources/global.o ../sources/global.cc
  $HOME/git/yoga/javascript/node_modules/nbind/bin/emcc '-DNODE_GYP_MODULE_NAME=nbind' '-D_DARWIN_USE_64_BIT_INODE=1' '-D_LARGEFILE_SOURCE' '-D_FILE_OFFSET_BITS=64' -I$HOME/.node-gyp/5.3.0/include/node -I$HOME/.node-gyp/5.3.0/src -I$HOME/.node-gyp/5.3.0/deps/uv/include -I$HOME/.node-gyp/5.3.0/deps/v8/include -I../sources -I../node_modules/nan -I../node_modules/nbind/include  -Os -mmacosx-version-min=10.5 -arch x86_64 -Wall -Wendif-labels -W -Wno-unused-parameter -std=gnu++0x -fno-rtti -fno-exceptions -fno-threadsafe-statics -fno-strict-aliasing -O3 --pre-js ./Release/obj.target/nbind/geni/pre.js --post-js ./Release/obj.target/nbind/geni/post.js --js-library ./Release/obj.target/nbind/geni/em-api.js -s NO_FILESYSTEM=1 -s "EXPORTED_FUNCTIONS=[\"_nbind_init\",\"_nbind_value\"]" -s "DEFAULT_LIBRARY_FUNCS_TO_INCLUDE=[\"nbind_value\",\"\$Browser\"]" -MMD -MF ./Release/.deps/Release/obj.target/nbind/sources/nbind.o.d.raw   -c -o Release/obj.target/nbind/sources/nbind.o ../sources/nbind.cc
  $HOME/git/yoga/javascript/node_modules/nbind/bin/emcc '-DNODE_GYP_MODULE_NAME=nbind' '-D_DARWIN_USE_64_BIT_INODE=1' '-D_LARGEFILE_SOURCE' '-D_FILE_OFFSET_BITS=64' -I$HOME/.node-gyp/5.3.0/include/node -I$HOME/.node-gyp/5.3.0/src -I$HOME/.node-gyp/5.3.0/deps/uv/include -I$HOME/.node-gyp/5.3.0/deps/v8/include -I../sources -I../node_modules/nan -I../node_modules/nbind/include  -Os -mmacosx-version-min=10.5 -arch x86_64 -Wall -Wendif-labels -W -Wno-unused-parameter -std=gnu++0x -fno-rtti -fno-exceptions -fno-threadsafe-statics -fno-strict-aliasing -O3 --pre-js ./Release/obj.target/nbind/geni/pre.js --post-js ./Release/obj.target/nbind/geni/post.js --js-library ./Release/obj.target/nbind/geni/em-api.js -s NO_FILESYSTEM=1 -s "EXPORTED_FUNCTIONS=[\"_nbind_init\",\"_nbind_value\"]" -s "DEFAULT_LIBRARY_FUNCS_TO_INCLUDE=[\"nbind_value\",\"\$Browser\"]" -MMD -MF ./Release/.deps/Release/obj.target/nbind/node_modules/nbind/src/common.o.d.raw   -c -o Release/obj.target/nbind/node_modules/nbind/src/common.o ../node_modules/nbind/src/common.cc
  $HOME/git/yoga/javascript/node_modules/nbind/bin/emcc '-DNODE_GYP_MODULE_NAME=nbind' '-D_DARWIN_USE_64_BIT_INODE=1' '-D_LARGEFILE_SOURCE' '-D_FILE_OFFSET_BITS=64' -I$HOME/.node-gyp/5.3.0/include/node -I$HOME/.node-gyp/5.3.0/src -I$HOME/.node-gyp/5.3.0/deps/uv/include -I$HOME/.node-gyp/5.3.0/deps/v8/include -I../sources -I../node_modules/nan -I../node_modules/nbind/include  -Os -mmacosx-version-min=10.5 -arch x86_64 -Wall -Wendif-labels -W -Wno-unused-parameter -std=gnu++0x -fno-rtti -fno-exceptions -fno-threadsafe-statics -fno-strict-aliasing -O3 --pre-js ./Release/obj.target/nbind/geni/pre.js --post-js ./Release/obj.target/nbind/geni/post.js --js-library ./Release/obj.target/nbind/geni/em-api.js -s NO_FILESYSTEM=1 -s "EXPORTED_FUNCTIONS=[\"_nbind_init\",\"_nbind_value\"]" -s "DEFAULT_LIBRARY_FUNCS_TO_INCLUDE=[\"nbind_value\",\"\$Browser\"]" -MMD -MF ./Release/.deps/Release/obj.target/nbind/node_modules/nbind/src/reflect.o.d.raw   -c -o Release/obj.target/nbind/node_modules/nbind/src/reflect.o ../node_modules/nbind/src/reflect.cc
  $HOME/git/yoga/javascript/node_modules/nbind/bin/emcc '-DNODE_GYP_MODULE_NAME=nbind' '-D_DARWIN_USE_64_BIT_INODE=1' '-D_LARGEFILE_SOURCE' '-D_FILE_OFFSET_BITS=64' -I$HOME/.node-gyp/5.3.0/include/node -I$HOME/.node-gyp/5.3.0/src -I$HOME/.node-gyp/5.3.0/deps/uv/include -I$HOME/.node-gyp/5.3.0/deps/v8/include -I../sources -I../node_modules/nan -I../node_modules/nbind/include  -Os -mmacosx-version-min=10.5 -arch x86_64 -Wall -Wendif-labels -W -Wno-unused-parameter -std=gnu++0x -fno-rtti -fno-exceptions -fno-threadsafe-statics -fno-strict-aliasing -O3 --pre-js ./Release/obj.target/nbind/geni/pre.js --post-js ./Release/obj.target/nbind/geni/post.js --js-library ./Release/obj.target/nbind/geni/em-api.js -s NO_FILESYSTEM=1 -s "EXPORTED_FUNCTIONS=[\"_nbind_init\",\"_nbind_value\"]" -s "DEFAULT_LIBRARY_FUNCS_TO_INCLUDE=[\"nbind_value\",\"\$Browser\"]" -MMD -MF ./Release/.deps/Release/obj.target/nbind/node_modules/nbind/src/em/Binding.o.d.raw   -c -o Release/obj.target/nbind/node_modules/nbind/src/em/Binding.o ../node_modules/nbind/src/em/Binding.cc
  $HOME/git/yoga/javascript/node_modules/nbind/bin/emcc -O3 --pre-js ./Release/obj.target/nbind/geni/pre.js --post-js ./Release/obj.target/nbind/geni/post.js --js-library ./Release/obj.target/nbind/geni/em-api.js -s NO_FILESYSTEM=1 -s "EXPORTED_FUNCTIONS=[\"_nbind_init\",\"_nbind_value\"]" -s "DEFAULT_LIBRARY_FUNCS_TO_INCLUDE=[\"nbind_value\",\"\$Browser\"]" -Wl,-search_paths_first -mmacosx-version-min=10.5 -arch x86_64 -L./Release  -o "Release/nbind.js" ./Release/obj.target/nbind/sources/yoga/YGNodeList.o ./Release/obj.target/nbind/sources/yoga/Yoga.o ./Release/obj.target/nbind/sources/Node.o ./Release/obj.target/nbind/sources/global.o ./Release/obj.target/nbind/sources/nbind.o ./Release/obj.target/nbind/node_modules/nbind/src/common.o ./Release/obj.target/nbind/node_modules/nbind/src/reflect.o ./Release/obj.target/nbind/node_modules/nbind/src/em/Binding.o 
```

mikearmstrong001:
I've put up this  [commit](https://github.com/mikearmstrong001/yoga/commit/a385d6d63561dc1c4af6cffe488909a76aeef205) on my branch which shows use and npm build for browser.

Feel free to pull the relevant and useful parts into your PR, my aim is purely to make sure we get the coverage required and take some pressure of you :)

Please note the measure funcs are not working yet, I've had to expose the measureFuncs to nbind and there is some issues with the attached function, maybe you've resolved this but at least there is now a test.

arcanis:
Hm that's strange, I'm not sure why are those flags disappearing. My guess is that it has something to do with you being on OSX (I build on Linux), but I'm unfortunately not familiar enough with OSX to be affirmative :( Perhaps I count find a VM but it would take some time.

In the meantime, I've pushed a commit to implement measure funcs and ship a webpack config that should help you run the tests in the browsers. It should be compiled with `npm run build:asmjs && npm run build:browser`, after then you can use `http-server` or alike from the current directory to access the `test/` directory.

mikearmstrong001:
pushed a bunch a few other changes to my branch, feel feel to grab: https://github.com/mikearmstrong001/yoga/commit/a3e4640d136c73f57d4cb62a1c8ee0036b6a9a87

Adds obtaining parent, additional apis and fixes a concern I have with dereferencing nullptrs in the getChild functions.

arcanis:
I'm gonna pull it. In the meantime, can you check if using the `develop` branch from nbind fix your flag issue?

arcanis:
Btw, what is `hasNewLayout` used for? Is it part of the public interface? I don't see it documented

mikearmstrong001:
> I'm gonna pull it. In the meantime, can you check if using the develop branch from nbind fix your flag issue?

the develop branch does not resolve the missing flags and the .mem files is still produced

> Btw, what is hasNewLayout used for? Is it part of the public interface?

@emilsjolander would have a better idea but my philosophy is that in providing a port we should expose all the current functionality there is a strong reason not to: eg the use of the YGNode Context is used internally in this port and JS provides a better way.

mikearmstrong001:
@arcanis would you attach your generated `javascript/build/Makefile` and `javascript/build/nbind.target.mk` from a linux build, this might help track down the issue.

arcanis:
Here they are: [Makefile](http://sprunge.us/QgZb) & [nbind.target.mk](http://sprunge.us/XOSW)

mikearmstrong001:
Looks like I've found the changes needed to build correctly on mac, please see https://github.com/mikearmstrong001/yoga/commit/ad7ae99b3b321b0a6504869a9f24da7d0514b0ea 

A useful reference point https://github.com/nodejs/node-gyp/wiki/%22binding.gyp%22-files-out-in-the-wild

arcanis:
Thanks, I've just pushed a commit, can you try? I've tried to avoid duplicating the flags twice, but if it doesn't work I'll just take your file as is. I've also removed the `-s PRECISE_F32` option, which was a relicate of an experiment I made and had no real effect except maybe worse perfs.

emilsjolander:
> Btw, what is hasNewLayout used for? Is it part of the public interface?

I hope to remove this soon, don't bother exposing it. It does not effect the algorithm and is just there as a flag to be set and read by users of Yoga (should not be part of yoga but is for historical reasons).

mikearmstrong001:
OSX build seems to be happy but when I cloned your version I did need to change the binding.gyp to:

```
{
    "targets": [ {

    	"includes": [
    	    "auto.gypi",
                "final-flags.gypi"
    	],

    	"sources": [
            "../yoga/YGNodeList.c",
            "../yoga/Yoga.c",
            "sources/Node.cc",
            "sources/global.cc",
    	    "sources/nbind.cc"
    	],

        "cflags": [
            "-DNBIND"
        ],

        "include_dirs": [
            "sources",
            "../",
        ]

    } ],

    "includes": [
	   "auto-top.gypi"
    ]
}
```

not the additional include and changes for `../yoga/YGNodeList.c` and `../yoga/Yoga.c`. Do you have a yoga directory under source? I seem to remember a copy happening, maybe a throwback ?

arcanis:
Npm should automatically cp the yoga source files when available, when running the prepublish hook ([here](https://github.com/facebook/yoga/pull/304/files#diff-a5a7369df8d76f082ec86f71c0e2188bR26)) (this hook should be also executed during `npm install` until this behaviour is completely [deprecated](http://blog.lholmquist.org/npm-prepublish-changes/)). We can't use relative paths that go outside of the `javascript/` directory, otherwise these files won't be accessible when.

Currently a bit stuck on compiling the asmjs build on Travis ... a nice mix of node-gyp, osx, Travis itself, and a 6-minutes delay between each trial don't make it easy :smirk:

mikearmstrong001:
hmm, I've just cleared out and `node_modules`,` source/yoga` and `build` and run npm install as well as yarn. Build fails on both. if I then run `npm run prepublish` then `npm install` all is ok. I'm not convinced this is reliable and believe this needs to be looked at

Additionally, if you must do the copy surely it should be before each build as what happens if you update the source in between an npm install - you'd be building against old source.

FYI the example `binding.gyp` did build for both targets on OSX but I can't state for other platforms.

arcanis:
I've pushed a commit than should improve the copy, let me know if that works for you. I'm not sure I'll have the time to configure Travis well enough to compile the emscripten build, would the native tests be enough for the initial release?

emilsjolander:
@arcanis Where did we land on bundling a compiled version of the amjs code to test on travis so we don't need to run emscripten on travis?

mikearmstrong001:
Latest is building straight from your branch, testing is successful on node, asmjs and browser :)

arcanis:
Nice ! We're close ! 👍

I've just pushed a new commit after noticing some weird things with the memory on Node (the pointers were garbage collected as soon as the variables were leaving the scope, even though they had been inserted inside other nodes). This commit adds an internal list of shared_ptr to keep tracks of which nodes need to stay alive (with the unfortunate side effect that cyclic dependencies will never get garbage collected :/). If you or @emilsjolander can review it that would be great, just in case.

Now, back to Travis ! Why does it say it doesn't find the package.json file ? ...

emilsjolander:
> with the unfortunate side effect that cyclic dependencies will never get garbage collected

This seems pretty bad. Could you use weak pointers instead?

arcanis:
I'm not sure but I don't think so. In the following code, if A only contains a weak reference to B, it will not increase the refcount of B, which will reach 0 as soon as we leave the scope of the function (while still being referenced inside the Yoga node structure). Am I mistaken?

```js
let a = (function () {
    let a = new Yoga.Node();
    let b = new Yoga.Node();
    a.insertChild(b, 0);
    return a;
})();
```

emilsjolander:
Correct. When would the reference cycle occur which you mentioned? if a was child of b which was child of a? That should be be able to happen as we have asserts for that sort of thing in the C code.

arcanis:
If you omit the return statement in the code above, you end up with a cycling reference as far as the garbage collector is concerned:

```js
(function () {
    let a = new Yoga.Node();
    let b = new Yoga.Node();
    a.insertChild(b, 0);
})();
```

arcanis:
That being said, since garbage collection doesn't exist in asmjs, users will have to take care to free their objects themselves anyway, so ... :worried:

emilsjolander:
@mikearmstrong001 what do you think?

mikearmstrong001:
I'm of the opinion we should have one solution for asmjs, node and browser. An explicit free is required for asmjs and I would suggest we make that true for node as well and thus make the code more portable.

I ran some tests by exposing the `instanceCount` and assert it to be 0 at the end each test, even with an explicit free on the node it was not.

I've attached a gist which contains a recommended diff that passes tests on all platforms.

This adds a `dispose` and a `disposeRecursive` as well as a global to return the allocated `instanceCount`

https://gist.github.com/mikearmstrong001/9f5d87ed622119c97cebc6ebd1534bd6

arcanis:
Hm I'm not sure this `dispose` method is enough - if `free` is not correctly called, then even if we do free the YGNode instance, its wrapper class won't be freed and we will end up with a memory leak.

That being said, I think the behaviour you've witnessed might be caused because the garbage collector doesn't run consistently unless you specifically ask for it (with `--expose-gc` & `global.gc()`) (it's the same reason why I haven't noticed the segfaults that were happening in the destructors before my refactoring). I'll check it.

arcanis:
I digged a bit (spoiler alert, it's not super-great):

- Calling the nbind-providen `node.free()` function only decrease the reference counter, and doesn't actually free the resource (which won't be released as long as it is used inside an other node).

- I've overloaded this function so that it will also automatically detach the node from its parent and its children (so that it will get properly garbage collected as soon as the garbage collector will make a pass).

- I've also added a `.freeRecursive()` which also free its children.

- On asmjs (and on native module too), it is **required to manually release each node variable at some point**. Let me explain, but first just as a remainder: the `a = new Yoga.Node()` expression creates a new `shared_ptr` containing a new `Node`, and return it.

  When working on the C++ side, as soon as the scope of the `a` variable becomes unreachable, the underlying object will become candidate to be garbage collected. That's why we need `a` to hold a reference to a shared_ptr: so that the `Node` inside will *not* get garbage collected when the `a` variable get destroyed. Unfortunately, it has a side effect on asmjs!

  When working on asmjs, there is no such thing as garbage collection. The memory mapped to a variable will live forever, or at least until we call its `free` function, even if the `a` variable get destroyed. It means that as long as we don't call the `free` function, the underlying `shared_ptr` will still exist and will in turn keep the `Node` object alive.

  So what does it mean for us? Let's consider the example below:

  ```js
  // Note: new Yoga() and Yoga.makeSharedNode() are the same thing - remember that we need to
  // use a shared_ptr rather than actual object pointers, so "new Yoga()" delegates the
  // shared_ptr creation to makeSharedNode.

  (function () {
      let a = Yoga.makeSharedNode();
      let b = Yoga.makeSharedNode();
  }());

  (typeof gc !== "undefined") && gc();
  console.log(Yoga.getInstanceCount());
  ```

  In Node.js, once we evaluate the `gc()` function, both `a` and `b` will be garbage collected, their destructors will be called, the refcount of their internal `shared_ptr` will go down, and in turn the `Node` instances will be destroyed. The instance count returned will be 0.

  In asmjs, since there's no garbage collection, `a` and `b` will not get garbaged collected, their destructors won't be called, etc. The instance count returned will be 2. We will need to manually free those nodes:

  ```js
  (function () {
      let a = Yoga.makeSharedNode();
      let b = Yoga.makeSharedNode();
      a.free(); // <---
      b.free(); // <---
  }());

  (typeof gc !== "undefined") && gc();
  console.log(Yoga.getInstanceCount());
  ```

  But it gets tricker. Consider the following code:

  ```js
  (function () {
      let a = Yoga.makeSharedNode();
      let b = Yoga.makeSharedNode();

      a.insertChild(b, 0);
      let c = a.getChild(0);

      a.free();
      b.free();
      c.release();
  }());

  (typeof gc !== "undefined") && gc();
  console.log(Yoga.getInstanceCount());
  ```

  Noticed how we needed to manually release `c`? It's because `getChild` (and `getParent`) both return new `shared_ptr` instances rather than the existing ones (*nbind actually has different behaviours depending on whether we run on Node.js or asm.js. I've opened [an issue](https://github.com/charto/nbind/issues/59) on their repository*). If we hadn't released it manually, then the refcount of the `Node` object pointed by the `b` element would still be above 0, because of the `c` ptr that would never be garbage collected.

  Finally, it means that when recursively freeing a node, you still need to manually release the dangling references:

  ```js
  (function () {
      let a = Yoga.makeSharedNode();
      let b = Yoga.makeSharedNode();

      a.insertChild(b, 0);
      let c = a.getChild(0);

      a.freeRecursive();
      b.release();
      c.release();
  }());

  (typeof gc !== "undefined") && gc();
  console.log(Yoga.getInstanceCount());
  ```

This post is quite long but I just wanted to be clear about the memory implications of asm.js, and how to use it efficiently in things like ReactVR.

emilsjolander:
@arcanis Great investigation thanks! I'll let @mikearmstrong001 make the call here but wanted to give my input. I think this seems pretty complex and would rather (if possible) have all javascript platforms use the same memory management model (for now) even though it makes the node.js API slightly less intuitive. Can we have javascript manually manage memory like the C version? not use ref counting. This way nodes could be created and as long as all the nodes are attached in the same tree calling `freeRecursive()` on the root will free all the memory needed, nothing more to worry about.

arcanis:
I'm not 100% positive, but according to https://github.com/charto/nbind/issues/26 there's currently no other option to prevent an instance from being garbage collected than to use shared pointer :confused:

emilsjolander:
Maybe nbind is not the best option right now then?

jjrv:
`nbind` has a planned feature to make the Node.js platform work identically to the asm.js platform (you can free things manually for an immediate effect, but on Node.js it'll still garbage collect lost references that you could no longer manually free anyway).

It should be ready at the moment, only needs testing to land in the next release. If tests go well, I can release tomorrow.

emilsjolander:
@jjrv that would be awesome! Thanks for the update.

arcanis:
I've updated the code to use the solution @jjrv suggested in https://github.com/charto/nbind/issues/59, and I'm happy to say that everything seems to work fine :) I've also added a few tests to specifically check these cases, to avoid regressions in the future (https://github.com/facebook/yoga/blob/1cb22325da743342dd2bd5bc05f46ec48cd5a72d/javascript/tests/Facebook.Yoga/YGMemory.js).

@emilsjolander My opinion is that even if nbind currently has some shortcomings, I feel like the pro eventually outweight the cons. The benefits of sharing a single codebase are huge (for this project, but also for the numerous other OSS projects that could benefit from a JS port), and the missing parts are bound to eventually fade away as the project becomes more mature.

As for using the C memory model, I'm not completely sold. If we go this way, it will be much harder to ask users to update their code to correctly release references in the future. Maybe other function names would be enough (`releaseReference` / `detachAndRelease` / `recursivelyDetachAndRelease` ...)?

emilsjolander:
@arcanis Agreed, sharing JS implementation is a good thing. Having both `free` and `release` though is very confusing. With the C memory model all you would ever have to do is free the root to not leak any memory, this seems much simpler.

Again i'll let @mikearmstrong001 make the call on this as he will be the main internal user of this at the moment.

jjrv:
It's possible to force the C memory model by removing bindings for the C++ constructor.

A C++ factory function or static method can then create an object using `new` and return a pointer. When JavaScript no longer owns the object and only holds a pointer, nbind won't use `shared_ptr` any more.

A JavaScript wrapper class can then define a constructor that actually calls the C++ factory function.

mikearmstrong001:
@arcanis and @emilsjolander also agree a single solution is best and the memory model should be the same for both asmjs and node.

I don't think having a free and a release is a viable api choice, it has a strong code smell and it becomes very difficult to explain and justify from a user point of view to others, it would be clearer and less error prone to use the c api in JS directly despite the lack of OO.

For React VR I have tight control over the life time of the `YGNode` so really I'm thinking about clarity for other users coming to use this.

I can't look at a code level (other a cursory glance at the example usage) tonight but will be able to first thing tomorrow.

thanks @jjrv and @arcanis for the iteration!

arcanis:
As you've requested I've just pushed a commit that remove the entire `shared_ptr` logic and delegates to the user the responsibility to create and destroy the references:

- `Yoga.Node.create()` creates a new node
- `node.free()` destroy the node
- `node.freeRecursive()` destroy the node and all of its children

What do you think?

mikearmstrong001:
Thanks,

I suggest changing `Yoga.Node.create()` to `Yoga.createNode` and hiding the `Node` from export.

ie
```
    return Object.assign({

        createNode: function() {
            return lib.Node.create()
        },

        Layout,
        Size,

        setExperimentalFeatureEnabled,
        isExperimentalFeatureEnabled,

        getInstanceCount

    }, constants);
```

This will prevent accidental incorrect usage via `new Yoga.Node`. What are your thoughts ?

Memory retainment for both instance and JS objects seems good :) For testing I ran the Benchmark tests (once updated with the new function and freeRecursive) as a requestAnimationFrame and took Memory snapshots and there was no accruing of `Bound` nodes, to follow up I forced retaining of the JS object and got the expected behaviour, alongside attached properties.

Again, thanks so much for the iteration.

@emilsjolander how do you wanting documentation of apis for individual ports ?

arcanis:
> This will prevent accidental incorrect usage via new Yoga.Node. What are your thoughts ?

I've left out the constructor from the nbind bindings and declared the C++ constructor as private, so that should prevent nodes from being ever created via `new` (however the error isn't very clear. Maybe that could be fixed in nbind itself). The main interest is that `node instanceof Yoga.Node` yields the right result.

emilsjolander:
>  how do you wanting documentation of apis for individual ports ?

I moved the docs folder onto the master branch recently so it can be part of this PR (or a follow up). Add a javascript.md here https://github.com/facebook/yoga/tree/master/docs/_docs/api and then add a reference to it here https://github.com/facebook/yoga/blob/master/docs/_data/nav_docs.yml. I can later update the home page to feature js as well.

arcanis:
@emilsjolander @mikearmstrong001 I've added a `npm run bench` script that gives some insight about the perfs of the node module vs the asmjs one, renamed a few other scripts to make everything consistent, rebased the branch on top of master, and added some documentation. I haven't changed yet the gist links, since I wasn't sure if you wanted to do it yourself or not (also, have you considered using something like [gist-it](https://gist-it.appspot.com/) to directly print the relevant source code in the documentation?).

emilsjolander:
@arcanis This looks good to go from my end. @mikearmstrong001 what do you think?

Regarding using gists, I did it mostly because I like the styling of them and did not want to spend time replicating it 👍 I would rather have the code in the docs file :) I can update them and update the main landing page as well once this is landed.

mikearmstrong001:
@arcanis Thanks for the doc entries! I wasn't expecting you to take that on as well, really appreciated. I too would like to pull docs straight from code, it works well for React VR :)

I think your approach and justification for Node is good and agree that the error message is somewhat obtuse at
Browser:
```
Yoga.bundle.js:269 Uncaught TypeError: Cannot read property 'apply' of undefined
    at new Bound (Yoga.bundle.js:269)
   ...
```
Node
```
Error: Wrong number of arguments
    at evalmachine.<anonymous>:15:14
    at YGBENCHMARK (/Users/mikea/git/yoga-arc/javascript/tests/run-bench.js:29:21)
```
But given docs on how to create nodes already exist, this will be fine.

I made changes to your bench test and those are
1. allow the JS to warm up 
1. run the function several times and average the result

We should apply these to make measurements more valid

Before Change:
```
Stack with flex
  - node: 23ms (100%)
  - browser: 181ms (786.96%)

Align stretch in undefined axis
  - node: 50ms (100%)
  - browser: 193ms (386%)

Nested flex
  - node: 53ms (100%)
  - browser: 77ms (145.28%)

Huge nested layout
  - node: 18ms (100%)
  - browser: 86ms (477.78%)
```

After change:
```
Stack with flex
  - node: 19.4ms (100%)
  - browser: 33.8ms (174.23%)

Align stretch in undefined axis
  - node: 47.4ms (145.4%)
  - browser: 32.6ms (100%)

Nested flex
  - node: 38ms (125.41%)
  - browser: 30.3ms (100%)

Huge nested layout
  - node: 15ms (100%)
  - browser: 27.5ms (183.33%)
```


```
                let testEntry = testResults.get(name);

                if (testEntry === undefined)
                    testResults.set(name, testEntry = new Map());

                let WARMUP_IT = 3;
                for (let i=0; i<WARMUP_IT; i++) {
                    fn();
                }

                let BENCHMARK_IT = 10;
                let start = Date.now();
                for (let i=0; i<BENCHMARK_IT; i++) {
                    fn();
                }
                let end = Date.now();

                testEntry.set(type, (end - start) / BENCHMARK_IT);
```

Great stuff!

emilsjolander:
Thanks! This is an awesome addition. I'll import the PR now and start the merge process

javascript/sources/entry-common.js
```diff
@@ -0,0 +1,189 @@
...
+
+        toString() {
+
+            return `<Size#${this.width}x${this.height}>`;
+
+        }
+
+    }
+
+    patch(lib.Node.prototype, `free`, function () {
```
mikearmstrong001: For future maintainers: perhaps some comments on why these functions are patched.

arcanis:
Oh right ! Just pushed your changes. Funny how the asmjs build is noticeably faster than the native one on some tests ... maybe some optim flags will need to be tweaked.

emilsjolander:
@arcanis Just noticed that the added files are missing our open source license header. Could you please make sure all the added files have it?

arcanis:
@emilsjolander Yup, here they are !

javascript/sources/entry-common.js
arcanis: Just added some comments

PR #305, opened 2016-12-26
Carthage Support

hartbit:
This is a pull request to support the Carthage dependency manager for Objective-C and Swift development. Unfortunately, doing this is causing a few issues:

1. **The buck build is now broken:** I have the impression we can't `#import <yoga/Yoga.h>` from a public header of the framework because the importing module does not know the `yoga` module. This was the case in `UIView+YogaKit.h`. Fixing this requires adding the C headers as standard headers of the framework and `#import <YogaKit/Yoga.h>` instead. Unfortunately, this has broken the buck build. I don't know enough about buck to fix this. Any ideas?

2. **The tests fail in Xcode**: Weirdly enough, two tests which run fine when run with `buck test //YogaKit:YogaKitTests --config cxx.default_platform=iphonesimulator-x86_64 --config cxx.cflags=-DTRAVIS_CI` fail when run inside an Xcode test target. Those are `testSizeThatFitsSmoke` and
`testFrameAndOriginPlacement`. There's a test target in the Xcode projet of this Pull Request to play around with. Any ideas?


dshahidehpour:
Yeah, we were running into a similar problem internally. I'm going to take another look as soon as I have time, thanks for the PR!

hartbit:
@guidomb Yep, I saw I had forgotten that. But haven't fixed this yet because of the other roadblocks. And by the time they are cleared, I think I might as well do a PR to support Cocoapods at the same time. But that requires getting a good idea how to handle semver for the project.

guidomb:
@hartbit @dshahidehpour  Yeah, what do you guys think is the best approach moving forward? I am happy to help, I can add CocoaPods support but I think it would be better for now to wait, as @hartbit said, for a semver policy and ideally for #289 to be merged. Regarding buck I don't have any experience with it.

emilsjolander:
What is the status of this? 1. should be able to be fixed using header maps in xcode i think?

guidomb:
@hartbit Yes it worked. I must have done something wrong. I also added support for CocoaPods in #309 but for some reason I couldn't pick your fork as the PR's target.

hartbit:
Closing so it can be implemented in #309 instead.

PR #308, opened 2016-12-27
[YogaKit] Set nodes dirty when sizeThatFits changes

hartbit:
In YogaKit, nodes are not dirtied when `sizeThatFits` of leaf nodes change. This makes it difficult to reuse views like `UILabel` when their content (text) changes. I wrote a failing test to prove it and wrote a fix by caching the latest value of `sizeThatFits` in `YGNodeBridge` and checking for changes in `YGAttachNodesFromViewHierachy` (renamed to `YGUpdateNodesFromViewHierachy` to make it clear it's not simply reattaching nodes anymore). When a new size is detected, I call `YGNodeMarkDirty` on the leaf node.

**Unfortunately**, this makes another test fail for no logical reason (`testThatViewNotIncludedInFirstLayoutPassAreIncludedInSecond`). Either there's something I don't understand or I've un-advertedly unearthed a bug in the C implementation, but in what case would setting a node dirty make the layout return a different size?

### DIFF
--- a/YogaKit/Tests/YogaKitTests.m
+++ b/YogaKit/Tests/YogaKitTests.m
@@ -96,6 +96,27 @@
   XCTAssertTrue(CGSizeEqualToSize(CGSizeMake(514,21), containerSize), @"Size is actually %@", NSStringFromCGSize(containerSize));
 }
 
+- (void)testThatModificationsInSizeThatFitsUpdatesLayout
+{
+  UIView *container = [[UIView alloc] initWithFrame:CGRectMake(0, 0, 500, 50)];
+  [container yg_setUsesYoga:YES];
+  [container yg_setFlexDirection:YGFlexDirectionRow];
+  [container yg_setAlignItems:YGAlignFlexStart];
+
+  UILabel *label = [[UILabel alloc] initWithFrame:CGRectZero];
+  label.text = @"This is a short text.";
+  label.numberOfLines = 1;
+  [label yg_setUsesYoga:YES];
+  [container addSubview:label];
+
+  [container yg_applyLayout];
+  XCTAssertTrue(CGSizeEqualToSize(CGSizeMake(146,21), label.bounds.size), @"Size is actually %@", NSStringFromCGSize(label.bounds.size));
+
+  label.text = @"This is a slightly longer text.";
+  [container yg_applyLayout];
+  XCTAssertTrue(CGSizeEqualToSize(CGSizeMake(213,21), label.bounds.size), @"Size is actually %@", NSStringFromCGSize(label.bounds.size));
+}
+
 - (void)testFrameAndOriginPlacement
 {
   const CGSize containerSize = CGSizeMake(320, 50);
--- a/YogaKit/UIView+Yoga.m
+++ b/YogaKit/UIView+Yoga.m
@@ -13,6 +13,7 @@
 
 @interface YGNodeBridge : NSObject
 @property (nonatomic, assign, readonly) YGNodeRef cnode;
+@property (nonatomic, assign) YGSize cachedSize;
 @end
 
 @implementation YGNodeBridge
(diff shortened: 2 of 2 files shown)


hartbit:
@emilsjolander Any ideas?

dshahidehpour:
This seems to me like it is a problem at the C-Level. Calling `sizeThatFits:` is just the method that C-Layer is using for measuring the size of leaf nodes. I **don't think** that YogaKit should be responsible for dirtying nodes when we are building the YGNodeRef hierarchy.

hartbit:
The Yoga header says:

```
// Mark a node as dirty. Only valid for nodes with a custom measure function
// set.
// YG knows when to mark all other nodes as dirty but because nodes with
// measure functions
// depends on information not known to YG they must perform this dirty
// marking manually.
WIN_EXPORT void YGNodeMarkDirty(const YGNodeRef node);
```

In YogaKit, leaf nodes have a measure function, so Yoga does not know when that value has changed. I don't see how it can work without YogaKit dirtying them.

emilsjolander:
@hartbit Correct 👍 Thanks. Some tests are failing though

emilsjolander:
I think the more correct fix here would be to swizzle `-(void)setNeedDisplay` on leaves to call `YGNodeMarkDirty()` after calling the original `-(void)setNeedDisplay`. What do you think?

emilsjolander:
@hartbit yeah `setNeedsLayout` is better! Good catch. Not sure why the test is failing. Make the changes and if it is still failing i'll have a look 👍

dshahidehpour:
+1 to what @emilsjolander said. 

@hartbit, per your question:

>  By the way, do you have any ideas on why that test is failing? It makes no sense to me.

I'm pretty sure this is caused by part of your new diff (which I expect will probably disappear, once we add swizzling). Specifically, when you are passing size {0, 0} to remeasure dirty nodes.

dshahidehpour:
@hartbit, we started to see a bug internally that requires this fix today. I'm going to go ahead and update your diff and land. You'll still be credited with the work! :)

hartbit:
@dshahidehpour what about the solution with `setNeedsLayout` swizzling? What about the failing tests? I'm just curious on what the updated diff will do about those. 👍

dshahidehpour:
@hartbit Yes, I'll be fixing both! :)

dshahidehpour:
Okay, so the original plan was to swizzle `setNeedsLayout`. I had to take the PR in a slightly different direction, and here's why:

## Swizzling, `method_exchangeImplementations()`
I implemented the swizzling, but I didn't feel good about. I was also greeted with some warnings from our internal linter, so, I did some digging into whether this is *actually* a good idea, and found some interesting info.

First. In the most common cases of swizzling you need to call [`method_exchangeImplementions()`](https://github.com/opensource-apple/objc4/blob/cd5e62a5597ea7a31dccef089317abb3a661c154/runtime/objc-runtime-new.mm#L2992-L3014). Unfortunately, this calls `flushCaches()`. If `flushCaches()` is passed a nil argument (like it is `method_exchangeImplementations`) the **entire runtime method cache is nuked**. Which would happen every single time we called `+(void)load` on any subclass of `UIView`.

## Startup, `+(void)load`
The next problem I found is that on cold startup, `+(void)load` is called before the app has time to register any crash handlers, so, in the event of problem that occurred because of this diff our crash logging wouldn't be able to properly attribute it to the swizzling that was happening.

## Moving on
Despite all of this, I carried on. I reran the tests and they still failed...why?
When you change the text of a `UILabel`, it does not necessarily call `setNeedsLayout` because the text will simply truncate inside of the currently set frame based of the settings of the label. So, if we wanted to mark that a UILabel needed to be resized outside of `YogaKit`-land, we would have to explicitly call `setNeedsLayout`.

So at this point I thought to myself, "if we have to mark this dirty anyway, why bother with swizzling at all?". So, I just exposed the API for marking leafs dirty. The next step will be to have `YogaKit` automatically detect when leaves are dirty, which I think will be much easier to do once we land #289.

I definitely think we should keep iterating on this, but simply exposing a flag to mark leafs dirty is solid and stable first step.

PR #314, opened 2017-01-03
Improve JS bindings

arcanis:
- Fix a few things related to how npm should publish the package (sources were in the gitignore file, so I think it would have break the build during install - I fixed this by adding a npmignore that should override the gitignore rules)

- The enumerations values are now generated directly from `enums.py`

- I added percent unit support (#258) to the bindings (`.setWidthPercent` is currently exposed, but I've also added a very very little parsing to also support `.setWidth("100%")` and `.setWidth(.getWidth())`), added the missing tests, and fixed Travis.


emilsjolander:
Awesome! thanks

javascript/sources/entry-common.js
```diff
@@ -19,57 +19,11 @@ function patch(prototype, name, fn) {
...
-    constants.POSITION_TYPE_ABSOLUTE = 1;
-    constants.POSITION_TYPE_RELATIVE = 0;
-
-    constants.WRAP_NO_WRAP = 0;
-    constants.WRAP_WRAP = 1;
-
-    constants.UNDEFINED = NaN;
+    let constants = Object.assign({
+
+        UNDEFINED: NaN
```
mikearmstrong001: Any reason not to move this to YGEnums.js as well ? 
It seems like YGEnums would be the logical place to contain the value as it would mean you'd only need to require the one file if you needed all the constants.

javascript/sources/entry-common.js
```diff
@@ -129,6 +83,77 @@ module.exports = function (bind, lib) {
...
+
+            return this.value;
+
+        }
+
+    }
+
+    for (let fnName of [ `setPosition`, `setMargin`, `setFlexBasis`, `setWidth`, `setHeight`, `setMinWidth`, `setMinHeight`, `setMaxWidth`, `setMaxHeight`, `setPadding` ]) {
+
+        let methods = { [constants.UNIT_PIXEL]: lib.Node.prototype[fnName], [constants.UNIT_PERCENT]: lib.Node.prototype[`${fnName}Percent`] };
```
mikearmstrong001: nit: could we throw an exception here if either are undefined. It would mean that you would know early on if you'd not fully implements support for Percent on a new function and all tests would fail

javascript/sources/entry-common.js
arcanis: `UNDEFINED` isn't defined by `enums.py`, so it would get removed after each run.

javascript/sources/entry-common.js
arcanis: Sure. I've just pushed a commit to add this check.

Issue #132, opened 2015-09-25
Labels and image views baseline alignment support


zackhsuan:
Hi all,

Flexbox natively supports baseline alignment, but it seems like react-native has not implemented this yet. 
I suppose that it would be awesome if css-layout could offer controls like labels or image views to work with baseline mode. Plus, the native auto layout tech supports baseline alignment too.

Many thanks. I appreciate all you excellent work.

zackhsuan:
Is anyone here could help?

lucasr:
I thinks this ticket is more about React Native than css-layout itself as this requires some platform-specific bits. Unless I'm missing something?

zackhsuan:
Thanks for your reply. @lucasr but unfortunately, I was redirected by React Native to post the issue here. https://github.com/facebook/react-native/issues/3020 Any idea that I can at least get the right place to post this issue?

emilsjolander:
Baseline support is definitely something which would be great if we could support. We are not actively working on this but I would be happy to work with you or anyone else in the community on a pull request implementing it.

mpost:
`baseline` support sounds like a must have. Currently it is not possible to accurately align a "label" and a "text input" to be optical in line across platforms.

emilsjolander:
@mpost totally agreed. Just not a top priority yet.

mpost:
@emilsjolander thanks for the info. Speaking of top priorities, is there a roadmap to features we could expect?

emilsjolander:
No roadmap just yet. I am thinking about posting something soon.

woehrl01:
@emilsjolander @zackhsuan this can be closed now, can't it?

PR #317, opened 2017-01-04
Baseline support

woehrl01:
Added baseline support (see #132)

You have the ability for a custom baseline function (```float(*YGBaselineFunc)(YGNodeRef node);```) to return whatever baseline you want.


woehrl01:
Not sure if we should cache the baseline, as we get the baseline twice. First to get the maximum and second to position it. 

I'm thinking about something like ```computedFlexBasisGeneration``` just named ```baselineCalculationGeneration``` and a cached value property.

yoga/Yoga.c
```diff
@@ -334,6 +335,20 @@ YGMeasureFunc YGNodeGetMeasureFunc(const YGNodeRef node) {
   return node->measure;
 }
 
+
+void YGNodeSetBaselineFunc(const YGNodeRef node, YGBaselineFunc baselineFunc) {
+  if (baselineFunc == NULL) {
```
emilsjolander: `node->baseline = baselineFunc;` is enough

yoga/Yoga.c
```diff
@@ -941,6 +956,61 @@ static inline YGDirection YGNodeResolveDirection(const YGNodeRef node,
...
+
+static float YGBaselineOfFirstLine(const YGNodeRef node, const YGFlexDirection mainAxis, const float parentWidth)
+{
+  if(node->baseline != NULL)
+  {
+    return node->baseline(node);
+  }
+
+  YGNodeRef baselineChild = NULL;
+  for(unsigned int i = 0; i < YGNodeGetChildCount(node); ++i)
```
emilsjolander: uint32_t

emilsjolander: i++

woehrl01:
> Please run format.sh in the mean time.

I'd love to do this, but the effort to create the environment to run this on windows, is currently too much work for me. As it seams I need a full linux runtime to run this.

woehrl01:
@emilsjolander clang to the rescue! 👍 strangely clang-format on windows doesn't allow wildcard. So I hope I didn't miss a file.

woehrl01:
> most likely your shell is not doing the expansion.

jep 😄

yoga/Yoga.c
```diff
@@ -2452,19 +2494,30 @@ static void YGNodelayoutImpl(const YGNodeRef node,
 
       // compute the line's height and find the endIndex
       float lineHeight = 0;
+      float maxAscentForCurrentLine = 0;
```
emilsjolander: Declare these further down where they are used.

yoga/Yoga.c
```diff
@@ -941,6 +950,39 @@ static inline YGDirection YGNodeResolveDirection(const YGNodeRef node,
   }
 }
 
+static float YGBaselineOfFirstLine(const YGNodeRef node, const YGFlexDirection crossAxis) {
```
emilsjolander: `YGNodeBaseline`

yoga/Yoga.c
```diff
@@ -2452,19 +2494,30 @@ static void YGNodelayoutImpl(const YGNodeRef node,
...
           if (child->lineIndex != i) {
             break;
           }
-
           if (YGNodeIsLayoutDimDefined(child, crossAxis)) {
             lineHeight = fmaxf(lineHeight,
                                child->layout.measuredDimensions[dim[crossAxis]] +
                                    YGNodeMarginForAxis(child, crossAxis, availableInnerWidth));
           }
+          if (performLayout && YGNodeAlignItem(node, child) == YGAlignBaseline) {
```
woehrl01: isn't performlayout checked at line 2526 a little bit down below?

gentest/fixtures/YGAlignBaseline.html
```diff
@@ -0,0 +1,87 @@
...
+    <div style="width: 50px; height: 20px;"></div>
+  </div>
+  <div style="width: 50px; height: 20px;">
+    <div style="width: 50px; height: 15px;"></div>
+  </div>
+</div>
+
+
+<div id="align_baseline_column" style="width: 100px; height: 100px; flex-direction:column; align-items: baseline;">
+  <div style="width: 50px; height: 50px;"></div>
```
emilsjolander: please test with different widths on the two items as width is the axis being aligned here

gentest/fixtures/YGAlignBaseline.html
```diff
@@ -0,0 +1,87 @@
...
+
+<div id="align_baseline_multiline_column" style="width: 100px; height: 100px; flex-direction:column; align-items: baseline;flex-wrap:wrap;">
+  <div style="width: 50px; height: 50px;"></div>
+  <div style="width: 50px; height: 50px;">
+    <div style="width: 50px; height: 10px;"></div>
+  </div>
+  <div style="width: 50px; height: 20px;">
+    <div style="width: 50px; height: 10px;"></div>
+  </div>
+    <div style="width: 50px; height: 20px;"></div>
```
emilsjolander: indentation

gentest/fixtures/YGAlignBaseline.html
```diff
@@ -0,0 +1,87 @@
...
+
+<div id="align_baseline_multiline_row_and_column" style="width: 100px; height: 100px; flex-direction:row; align-items: baseline;flex-wrap:wrap;">
+  <div style="width: 50px; height: 50px;"></div>
+  <div style="width: 50px; height: 50px;flex-direction:column;">
+    <div style="width: 50px; height: 10px;"></div>
+  </div>
+  <div style="width: 50px; height: 20px;flex-direction:column;">
+    <div style="width: 50px; height: 10px;"></div>
+  </div>
+    <div style="width: 50px; height: 20px;"></div>
```
emilsjolander: indendation

gentest/fixtures/YGAlignBaseline.html
```diff
@@ -0,0 +1,87 @@
...
+
+<div id="align_baseline_self" style="width: 100px; height: 100px; flex-direction:row;">
+  <div style="width: 50px; height: 50px;align-self: baseline;"></div>
+  <div style="width: 50px; height: 20px;align-self: baseline;">
+    <div style="width: 50px; height: 10px;"></div>
+  </div>
+</div>
+
+<div id="align_baseline_multiline_column" style="width: 100px; height: 100px; flex-direction:column; align-items: baseline;flex-wrap:wrap;">
+  <div style="width: 50px; height: 50px;"></div>
```
emilsjolander: please vary the widths as this is a column layout

gentest/fixtures/YGAlignBaseline.html
```diff
@@ -0,0 +1,87 @@
+<div id="align_baseline" style="width: 100px; height: 100px; flex-direction:row; align-items: baseline;">
```
emilsjolander: This whole file would probably make sense to have as part of the YGAlignItems.html + YGAlignSelf.html files which already exist

yoga/Yoga.c
```diff
@@ -929,7 +938,12 @@ static inline float YGNodePaddingAndBorderForAxis(const YGNodeRef node,
 }
 
 static inline YGAlign YGNodeAlignItem(const YGNodeRef node, const YGNodeRef child) {
-  return child->style.alignSelf == YGAlignAuto ? node->style.alignItems : child->style.alignSelf;
+  YGAlign align =
```
emilsjolander: const

yoga/Yoga.c
```diff
@@ -941,6 +955,41 @@ static inline YGDirection YGNodeResolveDirection(const YGNodeRef node,
   }
 }
 
+static float YGBaseline(const YGNodeRef node,
+                        const YGFlexDirection crossAxis) {
+  if (node->baseline != NULL) {
+    const float baseline = node->baseline(node);
+    if (YGFloatIsUndefined(baseline)) {
+      return node->layout.measuredDimensions[dim[crossAxis]];
```
emilsjolander: This will always be height as baseline is undefined for column flex direction. right? in that case I think it is better to be clear and use `YGDimensionHeight` instead of `dim[crossAxis]`

yoga/Yoga.c
```diff
@@ -941,6 +955,41 @@ static inline YGDirection YGNodeResolveDirection(const YGNodeRef node,
   }
 }
 
+static float YGBaseline(const YGNodeRef node,
+                        const YGFlexDirection crossAxis) {
+  if (node->baseline != NULL) {
+    const float baseline = node->baseline(node);
+    if (YGFloatIsUndefined(baseline)) {
```
emilsjolander: Is there a good reason to allow undefined return value for baseline? I would prefer asserting and crashing as it would probably indicate a bug. What do you think?

yoga/Yoga.c
```diff
@@ -941,6 +955,41 @@ static inline YGDirection YGNodeResolveDirection(const YGNodeRef node,
...
+      return node->layout.measuredDimensions[dim[crossAxis]];
+    }
+    return baseline;
+  }
+
+  YGNodeRef baselineChild = NULL;
+  for (uint32_t i = 0; i < YGNodeGetChildCount(node); i++) {
+    const YGNodeRef child = YGNodeGetChild(node, i);
+    if (child->style.positionType == YGPositionTypeAbsolute || child->lineIndex > 0) {
```
emilsjolander: Why `child->lineIndex > 0`? I understand that it skips multiline children but i'm not sure why.

yoga/Yoga.c
```diff
@@ -941,6 +955,41 @@ static inline YGDirection YGNodeResolveDirection(const YGNodeRef node,
...
+      break;
+    }
+
+    if (baselineChild == NULL) {
+      baselineChild = child;
+    }
+  }
+
+  if (baselineChild == NULL) {
+    return node->layout.measuredDimensions[dim[crossAxis]];
```
emilsjolander: same as above regarding `dim[crossAxis]`

yoga/Yoga.c
```diff
@@ -941,6 +955,41 @@ static inline YGDirection YGNodeResolveDirection(const YGNodeRef node,
...
+  }
+
+  YGNodeRef baselineChild = NULL;
+  for (uint32_t i = 0; i < YGNodeGetChildCount(node); i++) {
+    const YGNodeRef child = YGNodeGetChild(node, i);
+    if (child->style.positionType == YGPositionTypeAbsolute || child->lineIndex > 0) {
+      continue;
+    }
+
+    if (YGNodeAlignItem(node, child) == YGAlignBaseline) {
```
emilsjolander: The baseline of a container is defined by its first baseline aligned child, but when there is no baseline aligned child then the container's baseline is defined by it's last child. This seems odd. Are you sure we have test cases covering edge cases here to make sure this is how it works on the web?

yoga/Yoga.c
```diff
@@ -941,6 +955,41 @@ static inline YGDirection YGNodeResolveDirection(const YGNodeRef node,
...
+      baselineChild = child;
+    }
+  }
+
+  if (baselineChild == NULL) {
+    return node->layout.measuredDimensions[dim[crossAxis]];
+  }
+
+  const float baseline = YGBaseline(baselineChild, crossAxis);
+  return baseline + baselineChild->layout.position[pos[crossAxis]];
```
emilsjolander: same as above regarding `pos[crossAxis]`

yoga/Yoga.c
```diff
@@ -966,6 +1015,21 @@ static inline bool YGNodeIsFlex(const YGNodeRef node) {
...
+  }
+
+  for (uint32_t i = 0; i < YGNodeGetChildCount(node); i++) {
+    const YGNodeRef child = YGNodeGetChild(node, i);
+    if (child->style.positionType == YGPositionTypeRelative &&
+        child->style.alignSelf == YGAlignBaseline) {
+      return true;
+    }
+  }
+  return false;
```
emilsjolander: nit: empty line above

yoga/Yoga.h
```diff
@@ -49,6 +49,9 @@ typedef YGSize (*YGMeasureFunc)(YGNodeRef node,
                                 YGMeasureMode widthMode,
                                 float height,
                                 YGMeasureMode heightMode);
+
+typedef float (*YGBaselineFunc)(YGNodeRef node);
```
emilsjolander: nit: remove added empty lines

yoga/Yoga.c
```diff
@@ -2452,19 +2516,30 @@ static void YGNodelayoutImpl(const YGNodeRef node,
...
           if (child->lineIndex != i) {
             break;
           }
-
           if (YGNodeIsLayoutDimDefined(child, crossAxis)) {
             lineHeight = fmaxf(lineHeight,
                                child->layout.measuredDimensions[dim[crossAxis]] +
                                    YGNodeMarginForAxis(child, crossAxis, availableInnerWidth));
           }
+          if (YGNodeAlignItem(node, child) == YGAlignBaseline) {
```
emilsjolander: same as above. Don't reference `crossAxis` as it confuses the fact that baseline is only defined for row layouts

yoga/Yoga.c
woehrl01: I'm not sure my self about this. I thought this could be a kind of "feature", so if you return undefined, you simply use the nodes height. But crashing would be fine too for me. What I'm not sure about is if we explicitly add the padding-top here or if the implementation of the custom function needs to consider this.

yoga/Yoga.c
woehrl01: we use only the first line of the children for base layout alignment. At least this is how chrome handles it.

yoga/Yoga.c
emilsjolander: The custom function should not take padding into account. We don't expect this for the measure function so I would like to preserve that here if possible.

Let's crash for now. If we find a valid reason to have this feature we can implement it later.

yoga/Yoga.c
emilsjolander: yes, but this looks like we are skipping over children with multiple lines instead of just looking at their first line?

yoga/Yoga.c
emilsjolander: So the current code is wrong i think. Right? As it will pick the last child in the case no child is baseline aligned?

yoga/Yoga.c
woehrl01: oh, yep. We should break instead of continue here!

yoga/Yoga.c
woehrl01: no as we only use the child if ```baselineChild``` is still ```NULL```.

emilsjolander:
@woehrl01 If you push all updates in one go (Ok if they are different commits) that would make it more clear to me and others when you are finished addressing comments so we can have another look 👍

yoga/Yoga.c
woehrl01: should I use ```YGFlexDirectionColumn``` here on the ```YGNodeLeadingMargin``` and ```YGNodeMarginForAxis``` calls?

woehrl01:
@emilsjolander sorry, will do next time!

yoga/Yoga.c
emilsjolander: yes :) and replace `dim[crossAxis]`

woehrl01:
I think I got all your points, please have a look 😄

yoga/Yoga.h
```diff
@@ -49,6 +49,8 @@ typedef YGSize (*YGMeasureFunc)(YGNodeRef node,
                                 YGMeasureMode widthMode,
                                 float height,
                                 YGMeasureMode heightMode);
+typedef float (*YGBaselineFunc)(YGNodeRef node);
+
```
emilsjolander: nit: remove added empty line

yoga/Yoga.c
```diff
@@ -2501,6 +2574,12 @@ static void YGNodelayoutImpl(const YGNodeRef node,
                 //                (auto) crossAxis dimension.
                 break;
               }
+              case YGAlignBaseline: {
+                child->layout.position[pos[crossAxis]] =
```
emilsjolander: Could `pos[crossAxis]` be hardcoded? Same with `crossAxis` a couple lines below

yoga/Yoga.c
emilsjolander: Please add this 👍

yoga/Yoga.c
```diff
@@ -941,6 +955,40 @@ static inline YGDirection YGNodeResolveDirection(const YGNodeRef node,
...
+  if (node->baseline != NULL) {
+    const float baseline = node->baseline(node);
+    YG_ASSERT(!YGFloatIsUndefined(baseline), "Expect custom baseline function to not return NaN")
+    return baseline;
+  }
+
+  YGNodeRef baselineChild = NULL;
+  for (uint32_t i = 0; i < YGNodeGetChildCount(node); i++) {
+    const YGNodeRef child = YGNodeGetChild(node, i);
+    if (child->lineIndex > 0) {
```
emilsjolander: not quite sure why we break if a child has more than one line. That would mean if the first child of this node has two lines then the baseline would default to the height of the node. I'm not sure why that would be correct.

woehrl01: we break as only childs on the first line which have align baseline are taken into account. I already added a testcase for this.

yoga/Yoga.c
woehrl01: not sure what you mean, should I add a empty line above ```return false```?

yoga/Yoga.c
woehrl01: sure, I just wanted to be the same like the others, to not cause confusion. But I'll replace them, to be explicit that it is only used on ```YGEdgeTop```

yoga/Yoga.c
woehrl01: oh, I see the confusion, ```child->lineIndex``` does not mean that it has more than one line. It means it is itself on the seconds (or later) line.

emilsjolander: @woehrl01 yeah that was my confusion, i'm not super familiar with the multi line code and read it too quickly 🐱

woehrl01:
@emilsjolander I additionally modified ```YGIsBaselineLayout``` to shortcut to false if it's a column layout.

anql:
@woehrl01 The baseline feature is really awesome,  but I'm not sure how to implement the function correctly. Would you give me some suggestions?  Thanks.

woehrl01:
@anql: Sure! You must return the vertical position inside the node you want to align. 

If you do not have a baseline function set, it uses the nodes height. That means that all the nodes are align at their bottom line. 

If you have a text node where you want to align your nodes across each others, you need to return the position of the baseline of the firstline inside the node.

I unpacked my awesome paint skills, to make this a bit clearer: 

image

If the black numbers are the height of the nodes, you need to return the red values, to align the nodes across. If you have multiline text you need to return the value of the first line.

anql:
The paint is so clearer, it's very helpful for me.  👍 @woehrl01

PR #320, opened 2017-01-05
[YogaKit] Fix crash when a non-leaf node becomes a leaf node

hartbit:
When a non-leaf node becomes a leaf node, `YGAttachNodesFromViewHierachy` calls `YGNodeSetMeasureFunc` before `YGRemoveAllChildren`, which causes a crash because that node has children. Fixed by swapping the order of the calls and have a unit-test to make sure it works.

### DIFF
--- a/YogaKit/Tests/YogaKitTests.m
+++ b/YogaKit/Tests/YogaKitTests.m
@@ -343,4 +343,22 @@
     }
 }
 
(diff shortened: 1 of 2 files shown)


dshahidehpour:
Nice find! I'll pull this in once the tests finish running :)

PR #335, opened 2017-01-12
Add YGLayoutGetMargin

### LINKED ISSUE #326: Content box
Now that percent unit support has been added to the library, I was wondering if you've considered exposing the content box on top of the element box. What do you think? It currently seems hard to know the exact size of the padding inside the box.

arcanis:
Fix #326. I'll open another PR once this one gets accepted to add support for `YGLayoutGetBorder` 👌


emilsjolander:
Could you add the c# and java bindings for this as well? Should be fairly straight forward as it will look exactly like the padding ones.

arcanis:
Eh eh, I was working on them, they've just been pushed :) I haven't been able to test, will Travis run the tests for those interfaces as well ?

emilsjolander:
Yup it will :) thanks!

arcanis:
Seems good, tests are now passing

arcanis:
@emilsjolander I started looking at how I can implement `YGNodeLayoutGetBorder`, and actually there will be a bit more work than I thought, since such a function would need to go through the same `EDGE_START` / `EDGE_END` check than its two siblings. Would you mind if I implemented it in this PR, since the code will be quite similar?

emilsjolander:
@arcanis I would rather it was a PR based on top of this one. Trying to move to smaller pull requests where possible as it makes it much easier to grok the code. Thanks 👍

arcanis:
Ok 👍

csharp/Facebook.Yoga/YogaNode.cs
```diff
@@ -333,6 +333,12 @@ public void SetPosition(YogaEdge edge, YogaValue value)
             }
         }
 
+        [Obsolete("use LayoutMargin properties")]
```
emilsjolander: Don't add an Obsolete property :p all these will be removed soon enough.

emilsjolander:
@arcanis Looks good now, thanks!

PR #336, opened 2017-01-12
[chsarp] Add YogaKit Shared and iOS version

rmarinho:
Using shared code for reuse in other platforms based on iOS native implementation.

Adds YogaKit sample.

Adds YogaKit tests (same as objc). (3 tests failing)

```
YogaKitTest : 80 ms
Facebook.YogaKit.iOS.Tests.exe : 81 ms
Tests run: 11 Passed: 8 Inconclusive: 0 Failed: 3 Ignored: 1
```

Since we don't have extension properties we need to go with a extension method to get access to the YogaLayout . 

I m also not sure this is leak free yet, would love some help with testing and feedback about view/node lifecycle


emilsjolander:
I'm worried about this copying a lot of logic from YogaKit. Can Xamarin interop with the obj-c code in YogaKit instead?

rmarinho:
This is so we can share YogaKit implementation to other platforms like android and uwp.

I could do binding for the the YogaKit lib ... but that will only work for iOS.

emilsjolander:
Aren't you tying yourself to iOS by doing `[DllImport("/usr/lib/libobjc.dylib")]` though?

rmarinho:
That file only lives in the YogaKit iOS project, it's a partial class , so the methods are only added when you are running in that platform.

emilsjolander:
Ok, makes sense. Thanks for taking the time to explain 👍 I'll let @splhack give the final ok on the c# code but it looks good to me.

splhack:
I feel maybe we can have another approach for this. For example, this line from test.

    root.Yoga().AlignItems = YogaAlign.Center;

It would be like next line if we can use `YogaNode` class directly.

    root.Yoga.AlignItems = YogaAlign.Center;

I'll try to hack around YogaKit based on this PR later.

rmarinho:
@splhack yeah maybe C# 7 or 8 will give us [Extension Properties ](https://github.com/dotnet/roslyn/issues/112)

rmarinho:
Added tests

PR #338, opened 2017-01-13
[YogaKit] Fix test bug

rmarinho:
Simple fix on a possible copy paste error. The test actually passes the same way by coincidence.


dshahidehpour:
Great catch, @rmarinho!

PR #341, opened 2017-01-17
fix a build error in YogaKitSample

laiso:
> yoga/yoga/YGValue.h: No such file or directory

Remove YGValue.h reference in the file tree


dshahidehpour:
Thanks for fixing this @laiso!

emilsjolander:
This has already been fixed by the look of it.

PR #347, opened 2017-01-20
Fix unnecessary measure calls

### LINKED ISSUE #334: measure function called multi times unnecessarily
the measure function:
```
static YGSize _measure(YGNodeRef node,
                       float width,
                       YGMeasureMode widthMode,
                       float height,
                       YGMeasureMode heightMode) {
  printf("_measure: %f(%s), %f(%s)\n",
         width,
         (widthMode==YGMeasureModeAtMost?"AtMost":widthMode==YGMeasureModeExactly?"Exactly":"Undefined"),
         height,
         (heightMode==YGMeasureModeAtMost?"AtMost":heightMode==YGMeasureModeExactly?"Exactly":"Undefined"));
  return YGSize {
    .width = 84.f,
    .height = 49.f,
  };
}
```

the tree:
```
 YGNodeRef root = YGNodeNew();
  YGNodeStyleSetWidth(root, 288.f);
  YGNodeStyleSetHeight(root, 288.f);
  YGNodeStyleSetFlexDirection(root, YGFlexDirectionRow);
  
  YGNodeRef root_child0 = YGNodeNew();
  YGNodeStyleSetPadding(root_child0, YGEdgeAll, 2.88f);
  YGNodeStyleSetFlexDirection(root_child0, YGFlexDirectionRow);
  YGNodeInsertChild(root, root_child0, 0);
  
  YGNodeRef root_child0_child0 = YGNodeNew();
  YGNodeSetMeasureFunc(root_child0_child0, _measure);
  YGNodeInsertChild(root_child0, root_child0_child0, 0);
  
  YGNodeCalculateLayout(root, YGUndefined, YGUndefined, YGDirectionLTR);
  
  YGNodeFreeRecursive(root);
```

and the result:
```
 _measure: 282.239990(AtMost), 282.239990(Exactly)
_measure: 83.999992(AtMost), 282.239990(Exactly)
```

I test this in iPhone and android mobiles.

Apparently, the float value `83.99999` and `84.f` are equal, so the second call is unnecessary, I guess.

woehrl01:
Fix #334

### DIFF
--- a/tests/YGMeasureTest.cpp
+++ b/tests/YGMeasureTest.cpp
@@ -26,6 +26,22 @@ static YGSize _measure(YGNodeRef node,
   };
 }
 
+static YGSize _measure_84_49(YGNodeRef node,
+                        float width,
+                        YGMeasureMode widthMode,
+                        float height,
+                        YGMeasureMode heightMode) {
+  int *measureCount = (int*) YGNodeGetContext(node);
+  if (measureCount) {
+    (*measureCount)++;
+  }
+
+  return YGSize {
+      .width = 84.f,
+      .height = 49.f,
+  };
+}
+
 TEST(YogaTest, dont_measure_single_grow_shrink_child) {
   const YGNodeRef root = YGNodeNew();
   YGNodeStyleSetWidth(root, 100);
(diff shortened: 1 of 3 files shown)


tests/YGMeasureTest.cpp
```diff
@@ -21,8 +21,24 @@ static YGSize _measure(YGNodeRef node,
   }
 
   return YGSize {
-      .width = 10,
-      .height = 10,
+      width = 10,
```
emilsjolander: .width/.height

tests/YGMeasureTest.cpp
```diff
@@ -21,8 +21,24 @@ static YGSize _measure(YGNodeRef node,
...
+                        YGMeasureMode widthMode,
+                        float height,
+                        YGMeasureMode heightMode) {
+  int *measureCount = (int*) YGNodeGetContext(node);
+  if (measureCount) {
+    (*measureCount)++;
+  }
+
+  return YGSize {
+      width = 84.f,
```
emilsjolander: .width/.height

tests/YGMeasureTest.cpp
```diff
@@ -47,6 +63,31 @@ TEST(YogaTest, dont_measure_single_grow_shrink_child) {
   YGNodeFreeRecursive(root);
 }
 
+TEST(YogaTest, measure_only_single_time) {
```
emilsjolander: please use a more descriptive name for the test

emilsjolander: Also could you move the test to YGMeasureCacheTest?

woehrl01: will do

tests/YGMeasureTest.cpp
woehrl01: it would be great if I could leave the ```.``` as this is the only way to let it compile on VS15. As it seems there is no need for it on clang, too, this shouldn't be a problem. What do you think?

emilsjolander: Well the syntax is not part of the c99 designated initializer syntax as far as I know. I would rather stick to the standard for now.

woehrl01: It seems the problem here is, that we compile with c++ (```*.cpp```) and there is no c99 compliance for c++ compilation on VS15. 😞

PR #350, opened 2017-01-22
[YogaKit] Implemented auto-detection of dirtying

hartbit:
I've come back with my earlier proposal. As we saw on the previous PR, it seems that the only reliable way to do it is to cache the previous size and set the leaf dirty if the measure function returns a different value.

On the previous PR, my modifications caused the `testThatViewNotIncludedInFirstLayoutPassAreIncludedInSecond` test to fail. I finally found out why: the auto-dirtying was actually performing so well that it surfaced a bug in the test :) Because UIKit's `sizeThatFits` returns the current size by default, and because the leafs were not set with a `flexBasis` of `0`, the previous calculated size is taken into account. I fixed the test.

### DIFF
--- a/YogaKit/YGLayout.m
+++ b/YogaKit/YGLayout.m
@@ -102,6 +102,7 @@ YG_VALUE_SHORTHAND_EDGE_PROPERTY(lowercased_name, capitalized_name, capitalized_
 @interface YGLayout ()
 
 @property (nonatomic, weak, readonly) UIView *view;
+@property (nonatomic, assign) YGSize cachedSize;
 
 @end
 
@@ -132,13 +133,6 @@ YG_VALUE_SHORTHAND_EDGE_PROPERTY(lowercased_name, capitalized_name, capitalized_
 - (void)dealloc
 {
   YGNodeFree(self.node);
-}
-
-- (void)markDirty
-{
-  if (self.isLeaf) {
-    YGNodeMarkDirty(self.node);
-  }
 }
 
 - (NSUInteger)numberOfChildren
@@ -318,6 +312,14 @@ static void YGAttachNodesFromViewHierachy(UIView *const view)
   if (yoga.isLeaf) {
     YGRemoveAllChildren(node);
     YGNodeSetMeasureFunc(node, YGMeasureView);
+
+    YGSize oldSize = yoga.cachedSize;
+    YGSize newSize = YGMeasureView(node, 0, YGMeasureModeUndefined, 0, YGMeasureModeUndefined);
+    yoga.cachedSize = newSize;
+
+    if (!YGNodeIsDirty(node) && (oldSize.width != newSize.width || oldSize.height != newSize.height)) {
+        YGNodeMarkDirty(node);
+    }
   } else {
     YGNodeSetMeasureFunc(node, NULL);
 
--- a/YogaKit/Tests/YogaKitTests.m
+++ b/YogaKit/Tests/YogaKitTests.m
@@ -284,16 +284,15 @@
 
   [container.yoga applyLayout];
 
-  XCTAssertEqual(subview1.bounds.size.width, 150);
-  XCTAssertEqual(subview2.bounds.size.width, 150);
-  XCTAssertEqual(subview3.bounds.size.width, 0);
+  XCTAssertTrue(CGSizeEqualToSize(CGSizeMake(150, 50), subview1.bounds.size), @"Actual size is %@", NSStringFromCGSize(subview1.bounds.size));
+  XCTAssertTrue(CGSizeEqualToSize(CGSizeMake(150, 50), subview2.bounds.size), @"Actual size is %@", NSStringFromCGSize(subview2.bounds.size));
+  XCTAssertTrue(CGSizeEqualToSize(CGSizeZero, subview3.bounds.size), @"Actual size %@", NSStringFromCGSize(subview3.bounds.size));
 
   subview3.yoga.isIncludedInLayout = YES;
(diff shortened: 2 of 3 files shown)


YogaKit/YGLayout.m
```diff
@@ -318,6 +312,14 @@ static void YGAttachNodesFromViewHierachy(UIView *const view)
   if (yoga.isLeaf) {
     YGRemoveAllChildren(node);
     YGNodeSetMeasureFunc(node, YGMeasureView);
+
+    YGSize oldSize = yoga.cachedSize;
+    YGSize newSize = YGMeasureView(node, 0, YGMeasureModeUndefined, 0, YGMeasureModeUndefined);
```
dshahidehpour: I'm going to be honest, I'm not sure this is something we would want to do.

1. We are now guaranteeing that every leaf node is going to be measure during every single layout call, that's going to get pretty expensive, especially if we have to size text.

2. Are viewed measured with zero size supposed to return a non-zero value? I don't think that's an assumption we should make. As a result, I'm not sure that the result is something that we should be comparing to an *actual* measurement.

hartbit: A few points:

1. My benchmarks show me that this should not be a concern. `UIImageView` `sizeThatFits` seems instantaneous because it simply returns the size of the image. `UILabel`'s `sizeThatFits` takes 0.0000567 seconds (5% STD) on my iPhone 6S.

2. I'm not quite sure I understand your point. Could you explain again? :)

I understand the drawbacks you are putting forward. But, having used this auto-detection in the app I'm currently working, I can tell you it's a lifesaver. I tried for a while to mark views as dirty when necessary, but it's very bug-prone. If we are really worried about performance, perhaps we could make it an optional feature?

```
[YGYoga setDirtyNodeDetectionEnabled:YES];
```

emilsjolander: @hartbit `UILabel` and `UITextView` can take substantially longer than that to measure text. It depends on the complexity of the text being shown (for example multiline bidirectional text). Also the 6s is a very fast phone.

hartbit: @emilsjolander @dshahidehpour Point taken. But in that case, I'd still argue that this feature is worthwhile for quite a few projects where the performance impact won't be felt. I use it myself on a project and I'm still at 60FPS. Could we introduce the feature as an **off-by-default** feature as suggested?

hartbit: @emilsjolander @dshahidehpour Sorry to bother you about this, but have you had time to think about my last message and the idea of introducing the feature as an "experimental" feature?

emilsjolander: @hartbit Sorry for the delay in responding. I would rather not implement these kinds of flags for now. Is it easy enough to implement outside of Yoga within applications which would like to use this? I think we should continue to think about this irregardless of whether we end up merging this PR, I know @dshahidehpour is very interested in solving this as well.

hartbit: It's not easy necessarily easy to implement within applications outside of yoga: it requires wrapping a call to `applyLayout` with a function which visits the view hierarchy, finding the leaf nodes, caching the view's size and dirty it only if the size changes. I know iOS fairly well, and I don't yet see a better way to do it. Swizzling `setNeedsLayout` was a good idea, but it didn't work in practice.

dshahidehpour:
Hey @hartbit @emilsjolander. I think I have solution that will work for all of us. It involves us taking advantage of another FB open-source project: `KVOController`.

I put up a sample diff to monitor property changes on a `UILabel`.

https://github.com/dshahidehpour/yoga/commit/55d38a22b0b6225b5a1441258412c91dda389c92

hartbit:
@dshahidehpour Looks promising! A few comments below:

## Performance

I have a project where I use `KVOController` and had to find alternative solutions in certain scenarios where I heavily used it (several per cell in a `UICollectionView`). It might be worth checking this out.

My original solution of caching the layout size of leaf nodes and comparing it to `sizeThatFits` on each layout calculation has a performance impact relative to the number of leaf nodes on each layout. The `KVOController` solution has the advantage of only having a performance impact relative to the number of property changes. So I think it should be better in most cases :) But it's worth a check.

If we go the KVO route and performance is a problem, we have the option of using the KVO APIs directly instead of going through `KVOController`.

# Maintenance & Extensibility

One drawback of the `KVOController` solution is that we have to maintain observers for all properties that affect the size of all `UIView` subclasses, which is both error-prone and needs updating when Apple introduces new properties which affect size calculation. And third-party custom `UIView`s don't automatically work: custom code/observers have to be written to call `makeDirty`.

dshahidehpour:
@hartbit You are totally correct. Here's what I propose.

Let's build a sample view hierarchy and test size caching against KVO. I agree that using KVO is extremely error prone, but (in a perfect world) it would have less performance implications in comparison to remeasuring UILabel.

I'll make a sample layout, and we can give it a shot!

emilsjolander:
@dshahidehpour what's the status on this? Do we have a sample to test perf on?
@hartbit Could you rebase this PR?

hartbit:
@emilsjolander I can rebase this PR, but I thought that you were not interested in that solution (caching the size). Correct?

emilsjolander:
Not sure, i haven't been following this very closely. I'll let @dshahidehpour decide

dshahidehpour:
@hartbit @emilsjolander I've also been really busy, and I haven't had time to profile the changes yet. I'll try and come back to it this week.

dshahidehpour:
I don't have the time to keep investigating this. I'm going to open a task and close this PR for inactivity.

PR #357, opened 2017-01-27
Feature auto margin

woehrl01:
Even so I know there are some opinions against ```margin: 0 auto``` it's still part of the spec: https://www.w3.org/TR/css-flexbox-1/#auto-margins and pretty usefull if you have to position via ```justify-content```. 

This PR adds an implementation for that.

It adds an additonal ```YGUnitAuto``` and margins got ```YGNodeStyleSetMarginAuto``` functions as well.


emilsjolander:
@woehrl01 We don't aim to implement the full css spec. Could you add some examples of layouts which are made significantly easier / possible with margin auto support compared to implementing those layouts today?

woehrl01:
@emilsjolander I didn't came across any layout yet, which wasn't able to implement with yoga today. But I had to add spacer views. I want to get rid of spacer views as much as I can and the last "thing" can be only solved with auto margins. 

I'm totally aware that there is no need for the full css spec, but as auto margins are explicitly mentioned and described in flex-box I would still account it to the flex-box model layouting logic.

emilsjolander:
I would count needing spacer views as a limitation. Could you give some examples of where margin auto would remove the need for spacer views?

woehrl01:
Sure, see this simple [JSFiddle](https://jsfiddle.net/uojyuspt/).

It's pretty much any layout were you want to have elements on one side and some additonal on the other side e.g for menus.

I also use this e.g. for centered notification messages where I have additional text at the bottom (can be boiled down to the layout above but in column direction)

emilsjolander:
Thanks, that helps!

gentest/fixtures/YGMarginTest.html
```diff
@@ -15,27 +15,73 @@
 </div>
 
 <div id="margin_and_flex_row" style="width: 100px; height: 100px; flex-direction: row;">
-  <div style="margin-start: 10px; margin-end; 10px; flex-grow: 1;"></div>
```
emilsjolander: This should be a fix in a seperate pull request.

gentest/gentest-cpp.js
```diff
@@ -15,6 +15,8 @@ function toValueCpp(value) {
 function toFunctionName(value) {
   if (value.indexOf('%') >= 0){
     return 'Percent';
+  }else if(value.indexOf('Auto') >= 0){
```
emilsjolander: space before `else`

gentest/gentest-cpp.js
```diff
@@ -189,7 +193,13 @@ CPPEmitter.prototype = Object.create(Emitter.prototype, {
   }},
 
   YGNodeStyleSetMargin:{value:function(nodeName, edge, value) {
-    this.push('YGNodeStyleSetMargin' + toFunctionName(value) + '(' + nodeName + ', ' + edge + ', ' + toValueCpp(value) + ');');
+    var valueStr = toValueCpp(value);
+    if(valueStr != 'YGAuto'){
```
emilsjolander: space before `(` and after `)`

gentest/gentest-cpp.js
```diff
@@ -189,7 +193,13 @@ CPPEmitter.prototype = Object.create(Emitter.prototype, {
...
+      valueStr = ', ' + valueStr;
+    }else{
```
emilsjolander: space before and after `else`

emilsjolander:
We currently use `YGUndefined` to mean `auto` for `width`, `height`, and `flex-basis`. I'm concerned with this being confusing after this PR introduced an explicit `auto` value for margin. If we introduce a specific `auto` value then I think this should be respected for other properties as well.

woehrl01:
@emilsjolander I thought about this too, should I add/change those? I would handle undefined the same like auto would on those values, so this wouldn't brake existing users.

emilsjolander:
@woehrl01 I'm not sure about the best way to handle it, feel free to experiment. But yes, whatever we do we should not break existing usages.

woehrl01:
@emilsjolander I changed ```width```, ```height``` and ```flex-basis``` to have ```auto``` as default value. undefined on the setter is silently converted to auto.

emilsjolander:
@woehrl01 looks pretty good. I'll get back to you later this week regarding this.

gentest/fixtures/YGMarginTest.html
```diff
@@ -39,3 +39,54 @@
...
+  <div style="width: 50px; height: 50px;margin-top:auto;margin-bottom:auto"></div>
+  <div style="width: 50px; height: 50px;"></div>
+</div>
+
+<div id="margin_auto_bottom_and_top_justify_center" style="width: 200px; height: 200px; justify-content: center;">
+  <div style="width: 50px; height: 50px;margin-top:auto;margin-bottom:auto"></div>
+  <div style="width: 50px; height: 50px;"></div>
+</div>
+
+<div id="margin_auto_top_double" style="width: 200px; height: 200px; align-items: center;">
```
emilsjolander: add a similar test for row direction. Also please change name to something like `margin_auto_multiple_children` for a `_row` and `_column` suffix.

woehrl01:
I added most of the language bindings except objc and javascript. As I don't know objc good enough and for javascript I think it would be good to extend ```entry-common``` to also support ```"auto"``` as an additional short cut (the tests are generated for this) but I'm not familiar enough with this change.

emilsjolander:
@woehrl01 that's fine! I'll fix up javascript and objc when importing 👍 Thanks for this awesome addition 💯

emilsjolander:
@woehrl01 could you rebase this? Github says there are no conflicts but i'm having some issues importing it.

woehrl01:
@emilsjolander merged the master for this and the other branches you didn't imported yet.

emilsjolander:
@woehrl01 Could you rebase this on top of master instead of merging master into this branch?

woehrl01:
@emilsjolander I'll try my best, I'm more of a mercurial than a git user. 😊

emilsjolander:
@woehrl01 lol me to. Another option is to squash the commit and the push it, that should work as well when I import.

woehrl01:
@emilsjolander let me know if it works, otherwise I'll squash them all. 💪

emilsjolander:
@woehrl01 it didn't work :/ can you try squashing?

woehrl01:
@emilsjolander try this, please 🤘

emilsjolander:
@woehrl01 that worked! Thanks! sorry for the hassle.

woehrl01:
@emilsjolander 👏 awesome! No worries, thanks to this I learn new things about git today 👍

emilsjolander:
I'm still working on this. It breaks some tests internally and I have a couple other things that are higher priority. hang tight 👍

woehrl01:
@emilsjolander If there is anything were I can help you out. Let me know it! No rush 👍

emilsjolander:
@woehrl01 will do!

woehrl01:
@emilsjolander I did a rebase and fixed the auto margin as some of the behaviour this implicitly used has changed, now its explicit.

PR #365, opened 2017-01-30
Updates the js build

arcanis:
I've updated the js build. Let's see what Travis thinks of it!


emilsjolander:
thanks @arcanis this seems to look good. Not sure why I was not able to get it running

PR #366, opened 2017-01-31
Add support for showing the YogaKit README on CocoaPods.org

orta:
Hey there, congrats on the launch - we use React Native at Artsy, and I've been wondering about what Yoga + Jest Snapshots for RN looks like but haven't had the time to explore that yet. 

With https://github.com/CocoaPods/cocoadocs.org/pull/501 I added the ability to customize the location of the README for the front-page of the CocoaPods site. That is being deployed as I speak, so it should be available the next time you decide to ship a version of Yoga.

I also fixed the capitalisation of CocoaPods, and removed a redundant CocoaPods badge on your initial README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Yoga [![Build Status](https://travis-ci.org/facebook/yoga.svg?branch=master)](https://travis-ci.org/facebook/yoga) [![Cocoapods](https://img.shields.io/cocoapods/v/Yoga.svg)](http://cocoapods.org/pods/Yoga)[![Cocoapods](https://img.shields.io/cocoapods/v/YogaKit.svg)](http://cocoapods.org/pods/YogaKit)[![npm](https://img.shields.io/npm/v/yoga-layout.svg)](https://www.npmjs.com/package/yoga-layout)
+# Yoga [![Build Status](https://travis-ci.org/facebook/yoga.svg?branch=master)](https://travis-ci.org/facebook/yoga) [![CocoaPods](https://img.shields.io/cocoapods/v/YogaKit.svg)](http://cocoapods.org/pods/YogaKit) [![npm](https://img.shields.io/npm/v/yoga-layout.svg)](https://www.npmjs.com/package/yoga-layout)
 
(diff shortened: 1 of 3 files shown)


dshahidehpour:
Thanks @orta! We have one small problem, this repo actually contains *two* Pods ([yoga](http://cocoapods.org/pods/Yoga) and [YogaKit](http://cocoapods.org/pods/YogaKit)). 

As you can see, both of the pods reference the same github repo, however, we need a way to differentiate the README files that are used for each pod. I think that adding the `.cocoapods.yml` file is going to continue to point to the same README for both specs.

yoga is in the root directory, it's podspec and README is in the root directory.
YogaKit is in the `YogaKit` directory. It's podspec and README in that directory.

orta:
Hrm, we keep Podspecs free of any non-build related metadata, which makes that tough - I guess I do know of a sneaky work-around, and I'm not sure if I'm opening up pandora's box of encouraging more of this behavior https://github.com/CocoaPods/CocoaPods/issues/6405#issuecomment-272289620

orta:
Alright, I'm going to have CD handle it, but I'm keeping it undocumented outside of this PR, and it's corresponding PRs.

orta:
Alright, PR updated.

dshahidehpour:
Thanks @orta! I'm importing the PR now.

Just out of curiosity, why did you choose to got he route of `.cocoadocs.yml` instead of specifying a location in the `.podspec` (similar to the way you can reference the path to the LICENSE file)? Thanks again for helping out! :)

orta:
The Podspec has strictly always been about just build metadata, with the exception being the social media url. I ended up creating the `.cocoadocs.yml` to handle things like documentation styling and website related metadata

orta:
Just wasn't worth breaking some reasonable scope boundaries this late in the game

dshahidehpour:
@orta I've updated the podspec, but now that I have pushed the update to cocoapods.org, I'm still seeing the wrong README :/

http://cocoapods.org/pods/YogaKit

PR #369, opened 2017-01-31
Added property display: flex and none

### LINKED ISSUE #241: Feature: possibility to hide nodes
I know this is currently by design, but it would be great if we could hide / show nodes. Any roadmap on this? Maybe we can have a chat and I can implement this?

woehrl01:
Fix #241 and successor for #302 

Added new property ```display``` with ```YGDisplayFlex``` and ```YGDisplayNone```. Allows to hide nodes from the layout without the need to remove it from the DOM.


emilsjolander:
@joce does this solve it in a way that would work for you?
@woehrl01 could you add API bindings for java/csharp/objc/js?

I'll come back tomorrow and comment on the implementation. API looks good though. Thanks!

woehrl01:
@emilsjolander I could give it a try, but I only have c++ set up at the moment (new laptop) so I have to rely on travis, will take me "some" time. Not sure about js/nbind, maybe @arcanis could do a follow up 😊 please?

emilsjolander:
@woehrl01 java/csharp/objc should be trivial as it is just a simple property setter, just follow what is done for overflow already. I could probably do the javascript stuff unless @arcanis has some time 👍

woehrl01:
@emilsjolander "some" time 😉

arcanis:
I'm afraid I won't have time to work on this this week, but it shouldn't be much more than copying what is done for `setPositionType` - same prototype, same concept (cf https://github.com/facebook/yoga/blob/master/javascript/sources/Node.hh#L55, https://github.com/facebook/yoga/blob/master/javascript/sources/Node.cc#L67-L70 and https://github.com/facebook/yoga/blob/master/javascript/sources/nbind.cc#L55). One run of `enums.py`, another to generate the tests, and that should be it! ... should :)

emilsjolander:
@arcanis thanks! That should be enough for me to do this. Don't worry about it

woehrl01:
@emilsjolander @arcanis I updated those files, too, let's see if those are enough 👍

joce:
@woehrl01 You are a god among men. I can't thank you enough.

woehrl01:
@joce I'm happy that it fits your needs!

gentest/fixtures/YGDisplayTest.html
```diff
@@ -0,0 +1,20 @@
+<div id="display_none" style="width: 100px; height: 100px; flex-direction: row;">
```
emilsjolander: All these tests are a bit more complicated than they need to be. For example this first test could be done with just 2 children and have the same effect.

gentest/fixtures/YGDisplayTest.html
```diff
@@ -0,0 +1,20 @@
...
+ <div style="flex: 1;"></div>
+ <div style="flex: 1; display:none;">
+   <div style="flex: 1; width: 20px;"></div>
+ </div>
+ <div style="flex: 1;"></div>
+</div>
+
+<div id="display_none_with_position" style="width: 100px; height: 100px; flex-direction: row;">
+ <div style="flex-grow: 1;"></div>
+ <div style="flex-grow: 1; display:none; top: 10px;"></div>
```
emilsjolander: what are you trying to test here?

java/com/facebook/yoga/YogaNode.java
```diff
@@ -310,6 +310,18 @@ public void setOverflow(YogaOverflow overflow) {
     jni_YGNodeStyleSetOverflow(mNativePointer, overflow.intValue());
   }
 
+  private native int jni_YGNodeStyleGetDisplay(long nativePointer);
+  @Override
+  public YogaDisplay getDisplay() {
+    return YogaDisplay.values()[jni_YGNodeStyleGetDisplay(mNativePointer)];
```
emilsjolander: please use the generated fromInt() method on `YogaDisplay`. We should be doing the same in other places (i'll push a commit for that now, don't worry about it)

woehrl01: took the changes from GetOverflow and just renamed the name. If you push a commit I'll leave this for now.

gentest/fixtures/YGDisplayTest.html
woehrl01: If the position over hidden nodes is 0 or 10. It is 10, so we need to set the postion to 0 in ``` ZeroOutLayoutRecursivly```, not sure if this is actually a good idea, that's the reason I didn't implemented it yet. I'm curious what you think about that.

gentest/fixtures/YGDisplayTest.html
woehrl01: True, I'll make them simpler/split them up.

emilsjolander:
Looks good! Thanks for this 👍 👍 👍 👍 👍

PR #370, opened 2017-02-01
Generate enums in sorted order

woehrl01:
This PR forces the order of the generated enums in alphabetically order. This is needed to have a predictable order of the enums across different python versions. Which reduces the pain for merging multiple enum additions.


emilsjolander:
Cool stuff! thanks

Issue #378, opened 2017-02-07
YogaKit doesn't remember the origin of the root view.


dshahidehpour:
# Report

When you apply layout to a root view of layout with an origin that is not `CGPointZero`, in some scenarios it seems that this would be the incorrect behavior.

- [x] I have searched [existing issues](https://github.com/facebook/yoga/issues) and this is not a duplicate

# Issues and Steps to Reproduce
1. Clone or download [this](https://github.com/dshahidehpour/yoga/tree/space_repro)
2. Go into `YogaKit` directory and run `pod install`.
3. Open the workspace and run the app.
4. Click the last cell in table view.
5. Click the button at the bottom of the screen.

# Expected Behavior
The bordered view should stay on screen (ideally).

# Actual Behavior
The bordered view's .frame.origin is set to (0,0). 

![](https://media.giphy.com/media/l3q2CaaUj6CvvY9d6/giphy.gif)

dshahidehpour:
cc @hartbit Have you run into this? I'm trying to decide if there is a better way to fix this, other than requiring users to manually set the inset on a yoga view.

woehrl01:
does this problem persist if you use some thing different than```space-around``` on the root node? My theorie is that on the second click the root height is diffierent/bigger. And so you position the blue square at a lower position/centered.

dshahidehpour:
@woehrl01 let me spit out some debug info for you.

Here's the output when the screen loads (before I hit the button):
```
 1.{*wm: LAY_EXACTLY, hm: LAY_EXACTLY, aw: 414.000000 ah: 736.000000 initial
  2.{*wm: AT_MOST, hm: AT_MOST, aw: 414.000000 ah: 736.000000 measure
  2.}*wm: AT_MOST, hm: AT_MOST, d: (106.000000, 30.000000) measure
  2.{*wm: LAY_EXACTLY, hm: LAY_EXACTLY, aw: 414.000000 ah: 300.000000 flex
  2.}*wm: LAY_EXACTLY, hm: LAY_EXACTLY, d: (414.000000, 300.000000) flex
  2.{[skipped] wm: LAY_AT_MOST, hm: LAY_EXACTLY, aw: 414.000000 ah: 30.000000 => d: (106.000000, 30.000000) flex
 1.}*wm: LAY_EXACTLY, hm: LAY_EXACTLY, d: (414.000000, 736.000000) initial
[...]

PR #379, opened 2017-02-07
Fix resolving bug on changing layout and optimized memory usage

woehrl01:
Due to the changes in 46817a38. We have a bug if the layout changes. 

This PR fixes this bug and adds a test for it. 

Additionally it correctly marks negative percentage values as undefined dim. 

It also changes the ```resolvedDimensions``` to use a reference instead of the full value in order to minimize the memory requirement of a ```YGNode``` and reduces the copying of the ```YGValue```.

### DIFF
--- a/tests/YGRelayoutTest.cpp
+++ b/tests/YGRelayoutTest.cpp
@@ -29,3 +29,22 @@ TEST(YogaTest, dont_cache_computed_flex_basis_between_layouts) {
 
   YGSetExperimentalFeatureEnabled(YGExperimentalFeatureWebFlexBasis, false);
 }
+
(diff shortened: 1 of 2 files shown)


dshahidehpour:
Do you think this will fix the issue I'm having in #378

woehrl01:
@dshahidehpour I think the problem you encounter is a different one. Did your problem existes before the change in 46817a3 ? If it is then this is not the same. 

The problem I fixed here is, if you swap from defined height to undefined height, then it takes the previous definied height.

dshahidehpour:
@woehrl01 ahh, sounds like I still have a YogaKit bug

emilsjolander:
Could you squash & rebase. The PR failed to import

woehrl01:
Will do

PR #380, opened 2017-02-07
Exclude YGEnums.h from format.sh

woehrl01:
This PR excludes YGEnums.h from formatting via format.sh. Reduces the clutter due to missed format.sh calls after enums.py generation.


emilsjolander:
@woehrl01 Could you instead update format.sh to ignore YGEnums.h ? The reason format.sh does not handle it nicely is due to the macros. Because this file is generated we don't need to format it.

woehrl01:
@emilsjolander sure 👍 I wasn't sure how you handle the formating of those files internally.

emilsjolander:
thanks!

PR #386, opened 2017-02-10
[csharp,iOS] Fix callbacks on AOT

rmarinho:
When using AOT mode on Mono one can't use instance methods for callbacks, the compiler needs the MonoPInvokeCallback attribute on a static method to know how to get back to the managed world. 
This worked fine without the change in JIT mode. 

I not sure if we should use a __MONO__ flag for this stuff as this could be needed for other usage that not only on iOS.  

~~The adicional change is that one should as good practice call Dispose on the YogaNode when using callbacks to make sure we remove the handler from our dictionary, i was trying to write a test for this.. but i saw TestMeasureFuncWithDestructor was already there, but it doesn't seem correct to me, can you check @splhack.~~


splhack:
I'll look into this.

rmarinho:
Ok after internal feedback from Alex Corrado i got a new implementation for the fix using the native ref context of Yoga.

rmarinho:
The test for the Parent destructor is failing for me on AOT .. still looking at that one.

splhack:
can you take a look at #387 ?

chkn:
@splhack It looks like a more complicated way of doing the same thing? I do see one improvement: using a weak GCHandle and only allocating one of them.

Also note that neither PR appears to ever free the GCHandle

splhack:
Ah, I haven't seen the updated diff.

splhack:
revised in #388

splhack:
Could you rebase this onto the current master too? I'll merge this and #388

rmarinho:
It's rebased on top of master..

csharp/Facebook.Yoga/YogaLogger.cs
```diff
@@ -8,38 +8,42 @@
...
-                    }
-
-                    if (level == YogaLogLevel.Error)
-                    {
-                        throw new InvalidOperationException(message);
-                    }
-                };
-                Native.YGInteropSetLogger(_managedLogger);
+
+                Native.YGInteropSetLogger(LoggerInternal);
```
rolfbjarne: You need to store the delegate in a static field, otherwise the GC will collect it and the app will crash when the callback is called:

    static Func _logger_internal = LoggerInternal;

    Native.YGInteropSetLogger (_logger_internal);

csharp/Facebook.Yoga/YogaNode.cs
```diff
@@ -546,13 +547,21 @@ public void SetMeasureFunction(MeasureFunction measureFunction)
...
+
             Native.YGNodeSetMeasureFunc(_ygNode, _ygMeasureFunc);
         }
 
         public void SetBaselineFunction(BaselineFunction baselineFunction)
         {
             _baselineFunction = baselineFunction;
             _ygBaselineFunc = baselineFunction != null ? BaselineInternal : (YogaBaselineFunc)null;
+
+            var handle = GCHandle.Alloc(this);
```
rolfbjarne: Your patch never calls `GCHandle.Free`.

Also you allocate a GCHandle both SetMeasureFunction and SetBaselineFunction, if both are called one of the values stored with YGNodeSetContext will be overwritten. The same will happen if either function is called more than once.

csharp/Facebook.Yoga/YogaLogger.cs
rmarinho: @splhack can you take a note on this in your PR ?

csharp/Facebook.Yoga/YogaNode.cs
rmarinho: @splhack pr #388 fixes this.

csharp/Facebook.Yoga/YogaLogger.cs
splhack: sure

PR #387, opened 2017-02-10
[C#][iOS] Fix callbacks on AOT

splhack:
Based on the idea of #386


yoga/Yoga.h
```diff
@@ -143,6 +143,7 @@ WIN_EXPORT void YGNodeCopyStyle(const YGNodeRef dstNode, const YGNodeRef srcNode
   WIN_EXPORT type YGNodeLayoutGet##name(const YGNodeRef node, const YGEdge edge);
 
 YG_NODE_PROPERTY(void *, Context, context);
+YG_NODE_PROPERTY(void *, Managed, managed);
```
splhack: @emilsjolander technically we can use YGNodeSetContext instead of this. what do you think? this will keep a weak reference of YogaNode managed object.

PR #388, opened 2017-02-10
[C#][iOS] Fix callbacks on AOT

splhack:
Based on the idea of #386 by @rmarinho

### DIFF
--- a/csharp/Facebook.Yoga/Native.cs
+++ b/csharp/Facebook.Yoga/Native.cs
@@ -58,6 +58,14 @@ namespace Facebook.Yoga
         [DllImport(DllName, ExactSpelling = true, CallingConvention = CallingConvention.Cdecl)]
         public static extern int YGNodeGetInstanceCount();
 
+#if (UNITY_IOS && !UNITY_EDITOR) || __IOS__
+        [DllImport(DllName, ExactSpelling = true, CallingConvention = CallingConvention.Cdecl)]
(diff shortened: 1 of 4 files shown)


rmarinho:
Don't we need to free the handle? i think we might need IDisposable ,  free the handle on Dispose, and call Dispose(false) on the Destructor , we can clear parent weak reference too. i think it will help fix the test failing on iOS for parent destructor.

splhack:
In my understanding GCHandle.Alloc with GCHandleType.Weak means we don't need to free the handle, right?

> iOS for parent destructor

maybe iOS mono implementation has different GC timing for that certain situation?

rmarinho:
@splhack i m not a expert but from internal feedback i got that _"Yea, you still need to dispose of the GCHandle manually. "_

rmarinho:
About the test , maybe something around this :

https://github.com/rmarinho/yoga/commit/9a68a14699a038043d4176e5d5ce8b7c965e2b3a

splhack:
Or just skip GC test on iOS?

rmarinho:
sure that's an option ..

splhack:
How about this attempt to call GCHandle.Free?

rmarinho:
👍

rmarinho:
Can we merge this? why is the build failing ?

splhack:
Yeah, we're trying to do that with #386.

PR #393, opened 2017-02-12
[YogaKit] Add helper function for bulk updates.

dshahidehpour:
Currently, our configuration of Views looks like this:

```objc
UIView *view = [[UIView alloc] initWithFrame:CGRectZero];
view.yoga.isEnabled = YES;
view.yoga.height = 50;
view.yoga.width = 50;
```

Every time that we access `view.yoga` we have to access the associated object on `UIView` to get the `YGLayout`. This adds an extra `objc_msgSend` which increases binary size, and is slight perf impact.

This diff creates a way to modify the `YGLayout` with only a single `objc_msgSend`. Here's the new syntax:

```objc
UIView *view = [[UIView alloc] initWithFrame:CGRectZero];
[view configureLayoutWithBlock:^void(YGLayout *layout){
  layout.isEnabled = YES
  layout.height = 50;
  layout.width = 50;
}];
```

Here's the Swift version:
```swift
let view = UIView(frame: .zero)
view.configureLayout { (layout) in 
  layout.isEnabled = true
  layout.height = 50
  layout.width = 50
}
```

### DIFF
--- a/YogaKit/CHANGELOG.md
+++ b/YogaKit/CHANGELOG.md
@@ -25,3 +25,20 @@ view.yoga.marginLeft // 0
 ### Enhancements
 
 - Pixel Rounding now uses `roundf()` instead of `round()`.
+
+- There is now a method that allows "bulk" updates to YGLayout.
+```objc
+[view configureLayoutWithBlock:^(YGLayout *layout) {
+  layout.isEnabled = YES;
+  layout.width = 50;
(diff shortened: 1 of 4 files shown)


dshahidehpour:
cc: @hartbit, you'll probably be interested in this!

hartbit:
This is really cool :)

Issue #395, opened 2017-02-13
Simplify/Dedup STEP 5 Line 2271 - 2367


woehrl01:
The code [here](https://github.com/facebook/yoga/blob/master/yoga/Yoga.c#L2271-2367) contains lots of dublicate code, which could be greatly reduced. It switches only if its row or column. The code would be much more readable/maintainable if we calculate all of this named with cross/main-axis and only switch at the end to set either ```childWidth``` or ```childHeight```. 

If you agree, I'd put up a PR for this.

woehrl01:
create PR

PR #396, opened 2017-02-13
Removed duplicate code by calculating with mainSize/crossSize

woehrl01:
This PR removes some duplicate code by calculating with ```mainSize```/```crossSize``` and converting that to ```width``` or ```height``` at the end. See #395 .

### DIFF
--- a/yoga/Yoga.c
+++ b/yoga/Yoga.c
@@ -2345,8 +2345,8 @@ static void YGNodelayoutImpl(const YGNodeRef node,
 
         if (!YGFloatIsUndefined(availableInnerCrossDim) &&
             !YGNodeIsStyleDimDefined(currentRelativeChild, crossAxis, availableInnerCrossDim) &&
-            heightMeasureMode == YGMeasureModeExactly &&
-              !(isNodeFlexWrap && flexBasisOverflows) &&
+            measureModeCrossDim == YGMeasureModeExactly &&
+            !(isNodeFlexWrap && flexBasisOverflows) &&
             YGNodeAlignItem(node, currentRelativeChild) == YGAlignStretch) {
           childCrossSize = availableInnerCrossDim;
           childCrossMeasureMode = YGMeasureModeExactly;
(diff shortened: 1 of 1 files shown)


emilsjolander:
Awesome! Please hold off until tomorrow to rebase and I will import it

emilsjolander:
Could you rebase this to make sure everything still works, lots of changes recently landed? 👍

woehrl01:
@emilsjolander just landed 👍

emilsjolander:
thanks!

emilsjolander:
Some tests regarding aspect ratio are failing with this PR. Could you have a look?

woehrl01:
@emilsjolander 😢 I missed that, because of the [dots](https://github.com/facebook/yoga/blob/master/tests/YGAspectRatioTest.cpp#L19) those aren't included inside my VS project. Will definitely fix this!

emilsjolander:
@woehrl01 Oh that sucks. My understanding was that newer VS versions handled C99?

woehrl01:
@emilsjolander VS15 handels C99 but only on ```.c``` files as the tests are C++, it doesn't work there 😕

emilsjolander:
@woehrl01 Ok thanks for making that clear, I was not aware of that. You are a super important contributor so I would like to make this experience better for you and others using VS. I would be more than happy to accept a PR changes this to something which works better for you.

I would suggest using naming such as `YGValueMake(...)` and `YGSizeMake(...)`. Sorry for the inconvenience!

woehrl01:
@emilsjolander 

#### To failing tests:

I know where the problem is, I'll fix this likely tonight.

#### VS experience:

Simply removing the ```.``` is enough to compile with VS, this should works for clang, too. I'd setup a PR to add a more complete VS solution, including a separate test project, similar to #400. 
I need to take a look where best position for this is inside the repo, as I don't want to put this under ```/csharp```.

emilsjolander:
I would rather not just remove the `.` as that leads to syntax which is not really in any standard. Rather move to a proper API in that case.

woehrl01:
sounds good!

woehrl01:
@emilsjolander fixed the aspect ratio, sorry for the trouble!

emilsjolander:
No worries 👍  That's why we have tests ^^

Issue #394, opened 2017-02-13
Support for flex-wrap: wrap-reverse


emilsjolander:
https://developer.mozilla.org/en/docs/Web/CSS/flex-wrap

PR #398, opened 2017-02-13
flex-wrap: wrap-reverse support

woehrl01:
I couldn't resist to do this 😄 #394

This adds ```flex-wrap: wrap-reverse```

I think we hit a edge case here:
https://stackoverflow.com/questions/33891709/when-flexbox-items-wrap-in-column-mode-container-does-not-grow-its-width

as is differs here from chrome, but I think that yoga is here more correct. 

So I haven't added this test yet as this would fail against chrome, as chrome outputs a width of 30 for root, whereas yoga gets a width of 60 here, which I think is correct. Strangely the output of ```flex-wrap:wrap``` is in jsfiddle also only with a (visual) width of 30 on chrome, while the tests gets generated with 60.
 
```html
<div id="wrap_reverse_column" style="height: 100px; flex-wrap: wrap-reverse">
  <div style="height: 30px; width: 30px;"></div>
  <div style="height: 30px; width: 30px;"></div>
  <div style="height: 30px; width: 30px;"></div>
  <div style="height: 30px; width: 30px;"></div>
</div>
```

Looking forward what you think here @emilsjolander

### DIFF
--- a/gentest/fixtures/YGFlexWrapTest.html
+++ b/gentest/fixtures/YGFlexWrapTest.html
@@ -44,3 +44,17 @@
   <div style="width: 50px;"></div>
   <div style="width: 50px;"></div>
 </div>
+
+<div id="wrap_reverse_row_align_items_center" style="width: 100px; flex-direction: row; flex-wrap: wrap-reverse; align-items: center;">
+  <div style="height: 10px; width: 30px;"></div>
+  <div style="height: 20px; width: 30px;"></div>
+  <div style="height: 30px; width: 30px;"></div>
+  <div style="height: 40px; width: 30px;"></div>
+</div>
+
+<div id="wrap_reverse_row_align_items_flex_end" style="width: 100px; flex-direction: row; flex-wrap: wrap-reverse; align-items: flex-end;">
+  <div style="height: 10px; width: 30px;"></div>
+  <div style="height: 20px; width: 30px;"></div>
+  <div style="height: 30px; width: 30px;"></div>
+  <div style="height: 40px; width: 30px;"></div>
+</div>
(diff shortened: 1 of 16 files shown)


gentest/fixtures/YGFlexWrapTest.html
```diff
@@ -30,3 +30,17 @@
   <div style="flex-basis: 50px; height: 50px; min-width: 55px;"></div>
   <div style="flex-basis: 50px; height: 50px; min-width: 55px;"></div>
 </div>
+
+<div id="wrap_reverse_row_align_items_center" style="width: 100px; flex-direction: row; flex-wrap: wrap-reverse; align-items: center;">
```
emilsjolander: please add a test where 2 items are on the next line so that we ensure order is correct on the wrapped items

gentest/fixtures/YGFlexWrapTest.html
```diff
@@ -30,3 +30,17 @@
...
+  <div style="height: 10px; width: 30px;"></div>
+  <div style="height: 20px; width: 30px;"></div>
+  <div style="height: 30px; width: 30px;"></div>
+  <div style="height: 40px; width: 30px;"></div>
+</div>
+
+<div id="wrap_reverse_row_align_items_flex_end" style="width: 100px; flex-direction: row; flex-wrap: wrap-reverse; align-items: flex-end;">
```
emilsjolander: why are you testing with different `align-items` values?

woehrl01: ups, sorry I wanted to test with different ```align-content``` , thx!

gentest/fixtures/YGFlexWrapTest.html
woehrl01: will do

emilsjolander:
Could you rebase this to make sure everything still works, lots of changes recently landed? 👍

woehrl01:
@emilsjolander I already did the rebase with the latest changes

emilsjolander:
thanks!

PR #400, opened 2017-02-14
Create Xcode project so it's much easier to run tests.

dshahidehpour:
I don't even know how people have been running tests (unless they install BUCK). This creates a workspace to make it much easier to run tests.

### DIFF
--- a/YogaKit/Podfile
+++ b/YogaKit/Podfile
@@ -1,0 +1,10 @@
+use_frameworks!
+
+workspace 'YogaKit'
+
+target 'YogaKit' do
+  pod 'Yoga', :path => '../Yoga.podspec'
+end
+
+target 'YogaKitTests' do
+end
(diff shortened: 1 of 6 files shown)


dshahidehpour:
@hartbit and @guidomb, if one you could try patching this onto your machine and making sure it works the way you expect, that would be great! :)

hartbit:
@dshahidehpour Yeah, I've been using BUCK up to now, to get as close as possible to the environment travis will run the tests on. But the project is going to help for general development! It's working fine on  my machine.

guidomb:
I was about to test the xcode project but realized that it was removed from master. Why?

dshahidehpour:
We were running into a strange issue with the xcodeproj, and an internal BUCK builds. I'm going to take a look at this during the week and try to resolve it.

PR #401, opened 2017-02-14
Adds UWP build for yoga.dll

rozele:
The existing Yoga.vcxproj works great for Desktop apps, but breaks when you try to use the resulting DLL on mobile, Xbox, etc. This change compiles the DLL for the UWP and the Windows Store.


csharp/Yoga/Yoga.Universal.vcxproj
```diff
@@ -0,0 +1,279 @@
...
+      <Configuration>Release</Configuration>
+      <Platform>x64</Platform>
+    </ProjectConfiguration>
+  </ItemGroup>
+  <PropertyGroup Label="Globals">
+    <ProjectGuid>{2EACF721-73B5-46AE-9775-4A8674D05A9C}</ProjectGuid>
+    <Keyword>Win32Proj</Keyword>
+    <ProjectName>yoga</ProjectName>
+    <RootNamespace>Yoga</RootNamespace>
+    <AppContainerApplication>true</AppContainerApplication>
```
rozele: This is copy paste from Yoga.vcxproj except for these 5 lines and the `<CompileAsWinRT>false</CompileAsWinRT>` directives below.

rozele:
@matthargett still in the process of testing, but I've already fixed up the Facebook.Yoga.sln (master solution), as well as the .nuspec.  I'm going to publish a private package and see if I can get it to work.

rmarinho:
why is everyone doing the same thing? we need support for all platforms.. taking just 1 at the time isn't going to work. please try this package

https://www.myget.org/feed/yoga/package/nuget/Facebook.Yoga/1.1.0.10-pre

rozele:
@rmarinho I actually was not aware of the myget feed for Yoga.  That's great!

rozele:
@rmarinho Do you plan to add UWP support to your NuGet package?  Unfortunately, I cannot use that one unless uap10.0 target framework is added.

rmarinho:
@rozele i m using on  UWP project! you can't install the package ? what kind of Project ?

rmarinho:
It should be picking the netstandard stuff https://github.com/rmarinho/yoga/blob/xamarin-ci/csharp/nuget/Facebook.Yoga.nuspec#L60

rozele:
@rmarinho the netstandard stuff works fine for UWP Desktop, but this does not work for mobile, xbox, hololens, etc.  Try running your NuGet package on a Windows 10 Mobile emulator.

rmarinho:
yap trying it out with your change..

pre10der89:
@rozele The changes to add a Universal yoga.dll look good to me... From the conversation above with @rmarinho it seems there is an official nuspec now... If this is the case, then perhaps we should eventually remove the nuget folder under the csharp folder?

splhack:
can we proceed to merge this?

rmarinho:
@splhack i have talked with @rozele and have pull in the changes to my CI branch that generates the nuget, and it's working ok. I will put this in a PR for finishing up CI integration and nuget. I think we might close or we can merge and i adjust my PR later, since it's just some naming on the nuspec

rozele:
@splhack @rmarinho this is ready to merge if you'd like, it would certainly reduce the diff size for @rmarinho's pending PR

rmarinho:
Yeah @splhack you can merge it now it will make my PR smaller.

PR #402, opened 2017-02-14
Remove no longer valid limitiation hints

woehrl01:
Remove no longer valid limitiation hints, according to percentag, auto margins and flex values.


emilsjolander:
👍 So many out of date comments in this file... I'll try to go through some of them myself and remove more stuff as well.

woehrl01:
@emilsjolander One small step at a time 😉

PR #416, opened 2017-02-21
Fix percentage in flexing parent

woehrl01:
If we don't measure exactly, percentage values aren't exactly either. Fix for #414.

### DIFF
--- a/yoga/Yoga.c
+++ b/yoga/Yoga.c
@@ -2045,7 +2045,7 @@ static void YGNodelayoutImpl(const YGNodeRef node,
   }
 
   const bool flexBasisOverflows =
-    measureModeMainDim == YGMeasureModeUndefined ? false : totalFlexBasis > availableInnerMainDim;
+      measureModeMainDim == YGMeasureModeUndefined ? false : totalFlexBasis > availableInnerMainDim;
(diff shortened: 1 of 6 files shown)


yoga/Yoga.c
```diff
@@ -2339,7 +2339,12 @@ static void YGNodelayoutImpl(const YGNodeRef node,
                                           availableInnerCrossDim) +
                            marginCross;
           childCrossMeasureMode =
-              YGFloatIsUndefined(childCrossSize) ? YGMeasureModeUndefined : YGMeasureModeExactly;
+              YGFloatIsUndefined(childCrossSize) ||
+                      (currentRelativeChild->resolvedDimensions[dim[crossAxis]]->unit ==
```
astreet: Could we extract this condition into a named bool? Or add comments on the two conditions we're trying to check here?

woehrl01: Sure, will do

emilsjolander:
@woehrl01 I think you accidentally added a bunch of `flex-basis` tests into this PR.

woehrl01:
@emilsjolander ups, sorry! here you are! 😄

Issue #413, opened 2017-02-20
align-content:stretch does not work like on web 


anql:
```
<div id="align_content_stretch_different_web_browser" style="align-content:stretch;">
	<div style="width:100px; height: 100px; align-items: center; flex-direction: row; align-content:stretch;">
		<div style="height: 10px; width: 10px; align-content:stretch;"></div>
	</div>
</div>
```

jmmiracle:
we already had this question.seems like align-content make align-items invalid.

emilsjolander:
PR has landed. This should be fixed now 👍

PR #417, opened 2017-02-21
Fix align-content:strech overriding align-item

woehrl01:
Fix for #413. This was a hangover from a previous attept to fix other align-content problems.

### DIFF
--- a/gentest/fixtures/YGAlignContentTest.html
+++ b/gentest/fixtures/YGAlignContentTest.html
@@ -143,7 +143,7 @@
   <div style="height: 50px;"></div>
 </div>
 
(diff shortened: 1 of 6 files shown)


gentest/fixtures/YGAlignContentTest.html
```diff
@@ -142,3 +142,9 @@
   <div style="height: 50px;"></div>
   <div style="height: 50px;"></div>
 </div>
+
+<div id="align_content_stretch_is_not_overriding_align_items" style="align-content:stretch;">
+	<div style="width:100px; height: 100px; align-items: center; flex-direction: row; align-content:stretch;">
```
emilsjolander: please use spacing instead of tabs 👍

PR #430, opened 2017-02-22
Fix a typo in C API documentation

AljoschaMeyer:
Change "Certain nodes need *to* ability to measure themselves" to "Certain nodes need *the* ability to measure themselves".


emilsjolander:
Thanks!

Issue #421, opened 2017-02-22
Getting styles for flex* props should return whatever was passed to the corresponding setter


emilsjolander:
`YGNodeStyleGetFlexShrink()` for example does not always return what was set in `YGNodeStyleSetFlexShrink()`. This is due to these being computed based on the shorthand `flex` property. This should however be more of an implementation detail and a style getter should always return what was passed to the setter.

woehrl01:
I'll give this a go 🔨

PR #431, opened 2017-02-22
Change flex getters return the setted values

woehrl01:
Changed the flex getters to return the values they were actually set. See #421 .

### DIFF
--- a/tests/YGDefaultValuesTest.cpp
+++ b/tests/YGDefaultValuesTest.cpp
@@ -25,9 +25,9 @@ TEST(YogaTest, assert_default_values) {
   ASSERT_EQ(YGPositionTypeRelative, YGNodeStyleGetPositionType(root));
   ASSERT_EQ(YGWrapNoWrap, YGNodeStyleGetFlexWrap(root));
   ASSERT_EQ(YGOverflowVisible, YGNodeStyleGetOverflow(root));
-  ASSERT_FLOAT_EQ(0, YGNodeStyleGetFlexGrow(root));
-  ASSERT_FLOAT_EQ(0, YGNodeStyleGetFlexShrink(root));
-  ASSERT_FALSE(YGNodeStyleGetFlexBasis(root).unit != YGUnitAuto);
+  ASSERT_TRUE(YGFloatIsUndefined(YGNodeStyleGetFlexGrow(root)));
+  ASSERT_TRUE(YGFloatIsUndefined(YGNodeStyleGetFlexShrink(root)));
+  ASSERT_EQ(YGNodeStyleGetFlexBasis(root).unit, YGUnitAuto);
 
   ASSERT_EQ(YGNodeStyleGetPosition(root, YGEdgeLeft).unit, YGUnitUndefined);
   ASSERT_EQ(YGNodeStyleGetPosition(root, YGEdgeTop).unit, YGUnitUndefined);
--- a/yoga/Yoga.c
+++ b/yoga/Yoga.c
@@ -413,7 +413,7 @@ void YGNodeCopyStyle(const YGNodeRef dstNode, const YGNodeRef srcNode) {
   }
 }
 
-inline float YGNodeStyleGetFlexGrow(const YGNodeRef node) {
+static inline float YGResolveFlexGrow(const YGNodeRef node) {
   if (!YGFloatIsUndefined(node->style.flexGrow)) {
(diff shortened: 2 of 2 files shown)


tests/YGDefaultValuesTest.cpp
```diff
@@ -25,9 +25,9 @@ TEST(YogaTest, assert_default_values) {
   ASSERT_EQ(YGPositionTypeRelative, YGNodeStyleGetPositionType(root));
   ASSERT_EQ(YGWrapNoWrap, YGNodeStyleGetFlexWrap(root));
   ASSERT_EQ(YGOverflowVisible, YGNodeStyleGetOverflow(root));
-  ASSERT_FLOAT_EQ(0, YGNodeStyleGetFlexGrow(root));
-  ASSERT_FLOAT_EQ(0, YGNodeStyleGetFlexShrink(root));
-  ASSERT_FALSE(YGNodeStyleGetFlexBasis(root).unit != YGUnitAuto);
+  ASSERT_TRUE(YGFloatIsUndefined(YGNodeStyleGetFlexGrow(root)));
```
emilsjolander: I think these should still return 0 (their current default) if nothing has been set.

yoga/Yoga.c
```diff
@@ -634,9 +630,9 @@ YG_NODE_STYLE_PROPERTY_IMPL(YGWrap, FlexWrap, flexWrap, flexWrap);
 YG_NODE_STYLE_PROPERTY_IMPL(YGOverflow, Overflow, overflow, overflow);
 YG_NODE_STYLE_PROPERTY_IMPL(YGDisplay, Display, display, display);
 
-YG_NODE_STYLE_PROPERTY_SETTER_IMPL(float, FlexGrow, flexGrow, flexGrow);
-YG_NODE_STYLE_PROPERTY_SETTER_IMPL(float, FlexShrink, flexShrink, flexShrink);
-YG_NODE_STYLE_PROPERTY_SETTER_UNIT_AUTO_IMPL(float, FlexBasis, flexBasis, flexBasis);
+YG_NODE_STYLE_PROPERTY_IMPL(float, FlexGrow, flexGrow, flexGrow);
```
emilsjolander: Can you also add a `YG_NODE_STYLE_PROPERTY_IMPL` for the `Flex` shorthand?

tests/YGDefaultValuesTest.cpp
woehrl01: Currently the default is Undefined and we use that value inside ```YGResolveFlexGrow```. What do you think how we should use that with the ```flex``` shorthand? Should we use ```flex``` to set the ``flexGrow``` directly. But so we loose the previous set values.

woehrl01: As ```flex``` 

> is a shorthand for the flex-grow, flex-shrink, and the flex-basis properties.

maybe we should explictily set ```flex-grow```, ```flex-shrink``` if we call it?

woehrl01: An alternative is, to return 0 for these values in a (custom) getter, and keep the undefined as an internal implementation detail?

emilsjolander:
👍

yoga/Yoga.c
```diff
@@ -423,7 +423,15 @@ inline float YGNodeStyleGetFlexGrow(const YGNodeRef node) {
   return 0.0f;
 }
 
-inline float YGNodeStyleGetFlexShrink(const YGNodeRef node) {
+float YGNodeStyleGetFlexGrow(const YGNodeRef node) {
+  return YGFloatIsUndefined(node->style.flexGrow) ? 0.0f : node->style.flexGrow;
```
astreet: Instead of 0.0f, can we have the reference the defaults instead for clarity and to not introduce bugs?

woehrl01:
We don't really have a default constant here. But this return value is at least covered by  the default tests.

emilsjolander:
@astreet Right, I forgot about this. The value encoded in the default struct (gYGNodeDefaults) is `YGUndefined` as it is resolved later using the flex shorthand.

emilsjolander:
@woehrl01 Can you pull out 0.0f into constants kDefaultFlexGrow and kDefaultFlexShrink ? then use these in both the style getter and the resolve function? I think that would make it clearer what the zero is referencing.

woehrl01:
Sounds like a great idea 💡 will do!

PR #433, opened 2017-02-22
Fix YogaKitExample: Change YGUnitPoint to YGUnitPixel

maicki:
Seems like a recent API change happened that was not reflected in the YogaKitExample yet.


maicki:
Ah I see it was actually renamed from `YGUnitPixel` to `YGUnitPoint`. Seems like the example was changed before an official release so the Cocoapods can point to the new API.

Closing therefore.

PR #442, opened 2017-02-24
WIP: Improve caching logic and fixed caching bug with percentage

woehrl01:
This PR adds some improvements to the caching.

- Reuse the ```YGNodeCanUseCachedMeasurement``` for normal nodes
- Introduces a more generic dirty flag to further improve skipping some calculations (skipping, not implemented yet)
- Fixes a bug where the cache is mistakingly used where some values use percentage (thats the main cause why I already publish it)
- Fixes a bug that it always uses the cache if you mistakingly calculate the layout with a negative width/height passed down from upper layers

I only relied on the unittest time execution for measuring performance and the time was reduced from 850ms to 630ms. But I certainly would like to check the time on Travis.


woehrl01:
Travis benchmark busted this pr.

PR #447, opened 2017-03-01
Explicitly set new experimental and a small other improvement

woehrl01:
This sets the new ```YGExperimentalFeatureMinFlexFix``` explicitly to false in the config. 

It also a changes a stretch reason to reflect that it is the strech in a multiline.


emilsjolander:
👍

PR #448, opened 2017-03-01
Mark all children as dirty if display changes

### LINKED ISSUE #443: calculation error of  'display' feature
The leaf node‘s measure result is error while 'YGNodeStyleSetDisplay' be called several times before calculating.

```C
#include <iostream>
#include "Yoga.h"

void toggle(const YGNodeRef tab1, const YGNodeRef tab2, const YGNodeRef root, const YGNodeRef label, bool tab1Display) {
    YGNodeStyleSetDisplay(tab1, tab1Display ? YGDisplayFlex : YGDisplayNone);
    YGNodeStyleSetDisplay(tab2, tab1Display ? YGDisplayNone : YGDisplayFlex);
    YGNodeCalculateLayout(root, YGUndefined, YGUndefined, YGDirectionLTR);
    
    float w = YGNodeLayoutGetWidth(label);
    float h = YGNodeLayoutGetHeight(label);
    printf("label display:%s, w=%f, h=%f\n", tab1Display ? "false": "true",  w, h);
    
    if (tab1Display) {
        assert(w == 0 && h == 0);
    } else {
        assert(w == 8 && h == 16);
    }
}


/*
 <div id="root">
 <div id="tab1" style="flex-grow:1;display:flex">
 
 </div>
 
 <div id="tab2" style="flex-grow:1;display:none">
 <div id="div1">
 <div id="label" style="width:8; height:16;"></div>
 </div>
 </div>
 </div>
 */
int main(int argc, const char * argv[]) {
    const YGNodeRef root = YGNodeNew();
    YGNodeStyleSetFlexDirection(root, YGFlexDirectionRow);
    YGNodeStyleSetHeight(root, 100);
    
    const YGNodeRef tab1 = YGNodeNew();
    YGNodeStyleSetFlexGrow(tab1, 1);
    const YGNodeRef tab2 = YGNodeNew();
    YGNodeStyleSetFlexGrow(tab2, 1);
    
    const YGNodeRef div1 = YGNodeNew();
    const YGNodeRef label = YGNodeNew();
    YGNodeStyleSetWidth(label, 8);
    YGNodeStyleSetHeight(label, 16);
    
    YGNodeInsertChild(div1, label, 0);
    
    YGNodeInsertChild(tab2, div1, 0);
    YGNodeInsertChild(root, tab1, 0);
    YGNodeInsertChild(root, tab2, 0);
    
    toggle(tab1, tab2, root, label, true);//OK
    toggle(tab1, tab2, root, label, false);//OK
    toggle(tab1, tab2, root, label, true);//OK
    toggle(tab1, tab2, root, label, false);//FAILED!
    
    YGNodeFreeRecursive(root);
    return 0;
}
```

woehrl01:
If we set ```display:none``` all children are set with layout sizes/values of 0. We need do mark all those children as dirty to force a relayout if we toggle the display on a higher parent node. This fixes #443

### DIFF
--- a/tests/YGDirtyMarkingTest.cpp
+++ b/tests/YGDirtyMarkingTest.cpp
@@ -70,6 +70,54 @@ TEST(YogaTest, dirty_propagation_only_if_prop_changed) {
   YGNodeFreeRecursive(root);
 }
 
+TEST(Yogatest, dirty_mark_all_children_as_dirty_when_display_changes){
+  const YGNodeRef root = YGNodeNew();
+  YGNodeStyleSetFlexDirection(root, YGFlexDirectionRow);
+  YGNodeStyleSetHeight(root, 100);
+  
+  const YGNodeRef child0 = YGNodeNew();
+  YGNodeStyleSetFlexGrow(child0, 1);
+  const YGNodeRef child1 = YGNodeNew();
+  YGNodeStyleSetFlexGrow(child1, 1);
+  
+  const YGNodeRef child1_child0 = YGNodeNew();
+  const YGNodeRef child1_child0_child0 = YGNodeNew();
+  YGNodeStyleSetWidth(child1_child0_child0, 8);
+  YGNodeStyleSetHeight(child1_child0_child0, 16);
+  
+  YGNodeInsertChild(child1_child0, child1_child0_child0, 0);
+  
+  YGNodeInsertChild(child1, child1_child0, 0);
+  YGNodeInsertChild(root, child0, 0);
+  YGNodeInsertChild(root, child1, 0);
+  
+  YGNodeStyleSetDisplay(child0, YGDisplayFlex);
+  YGNodeStyleSetDisplay(child1, YGDisplayNone);
+  YGNodeCalculateLayout(root, YGUndefined, YGUndefined, YGDirectionLTR);
+  ASSERT_FLOAT_EQ(0, YGNodeLayoutGetWidth(child1_child0_child0));
+  ASSERT_FLOAT_EQ(0, YGNodeLayoutGetHeight(child1_child0_child0));
+
+  YGNodeStyleSetDisplay(child0, YGDisplayNone);
+  YGNodeStyleSetDisplay(child1, YGDisplayFlex);
+  YGNodeCalculateLayout(root, YGUndefined, YGUndefined, YGDirectionLTR);
+  ASSERT_FLOAT_EQ(8, YGNodeLayoutGetWidth(child1_child0_child0));
+  ASSERT_FLOAT_EQ(16, YGNodeLayoutGetHeight(child1_child0_child0));
+
+  YGNodeStyleSetDisplay(child0, YGDisplayFlex);
+  YGNodeStyleSetDisplay(child1, YGDisplayNone);
+  YGNodeCalculateLayout(root, YGUndefined, YGUndefined, YGDirectionLTR);
+  ASSERT_FLOAT_EQ(0, YGNodeLayoutGetWidth(child1_child0_child0));
+  ASSERT_FLOAT_EQ(0, YGNodeLayoutGetHeight(child1_child0_child0));
+
+  YGNodeStyleSetDisplay(child0, YGDisplayNone);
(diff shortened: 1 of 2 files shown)


yoga/Yoga.c
```diff
@@ -379,6 +379,14 @@ static void YGNodeMarkDirtyInternal(const YGNodeRef node) {
   }
 }
 
+static void YGNodeMarkChildrenDirtyInternal(const YGNodeRef node) {
```
emilsjolander: YGNodeMarkDirtyRecursive?

yoga/Yoga.c
```diff
@@ -698,6 +705,18 @@ YG_NODE_LAYOUT_RESOLVED_PROPERTY_IMPL(float, Margin, margin);
 YG_NODE_LAYOUT_RESOLVED_PROPERTY_IMPL(float, Border, border);
 YG_NODE_LAYOUT_RESOLVED_PROPERTY_IMPL(float, Padding, padding);
 
+void YGNodeStyleSetDisplay(const YGNodeRef node, const YGDisplay paramName) {
```
emilsjolander: `paramName` -> `display`

yoga/Yoga.c
```diff
@@ -698,6 +705,18 @@ YG_NODE_LAYOUT_RESOLVED_PROPERTY_IMPL(float, Margin, margin);
...
+  if (node->style.display != paramName) {
+    node->style.display = paramName;
+    YGNodeMarkDirtyInternal(node);
```
emilsjolander: no need to call `YGNodeMarkDirtyInternal` here as `YGNodeMarkChildrenDirtyInternal` also marks the passed in node as dirty

woehrl01: But it doesn't mark the parents as dirty.

tests/YGDirtyMarkingTest.cpp
```diff
@@ -70,6 +70,55 @@ TEST(YogaTest, dirty_propagation_only_if_prop_changed) {
   YGNodeFreeRecursive(root);
 }
 
+TEST(Yogatest, dirty_mark_all_children_as_dirty_when_display_changes)
+{
```
emilsjolander: nit: curly brace on previous line

tests/YGDirtyMarkingTest.cpp
```diff
@@ -70,6 +70,55 @@ TEST(YogaTest, dirty_propagation_only_if_prop_changed) {
...
+  const YGNodeRef root = YGNodeNew();
+  YGNodeStyleSetFlexDirection(root, YGFlexDirectionRow);
+  YGNodeStyleSetHeight(root, 100);
+  
+  const YGNodeRef tab1 = YGNodeNew();
```
emilsjolander: can you use similar naming for nodes as other tests? child0, child1, child1_child0 etc.

yoga/Yoga.c
```diff
@@ -379,6 +379,14 @@ static void YGNodeMarkDirtyInternal(const YGNodeRef node) {
   }
 }
 
+static void YGNodeMarkDirtyRecursive(const YGNodeRef node) {
```
emilsjolander: Sorry for the churn on this. I should have looked closer. `YGNodeMarkDirtyInternal` is already recursive so this actually doesn't make much sense ^^ You previous name is better suited 👍

woehrl01: I renamed it back, and also changed that it only affects the children and not the node itself, so we won't have the confusion with ```YGNodeMarkDirtyInternal```

emilsjolander:
👍 Thanks! Sorry about that

emilsjolander:
This PR failed to import. Could you rebase & squash. That usually fixes things. Thanks!

woehrl01:
here you are!

emilsjolander:
I took a slightly different approach to this when looking at the code again. You will see once this lands in a couple hours. I just made sure to set the cachedLayout property while handling the display:none case during layout.

woehrl01:
Great I idea! I haven't thought of that 👍

woehrl01:
@emilsjolander Much better! Great 👍

Issue #445, opened 2017-03-01
Add bool config to YGConfig to configure using web defaults


emilsjolander:
Historically Yoga has used other defaults then the web. Changing this now is not really an option as it would break all users of Yoga. However new products and frameworks using Yoga should not be effected by this if they would like to use the web defaults.

I suggest adding a `bool useWebDefaults` field to YGConfig.

zcbenz:
@emilsjolander Is there any discussion on why `useWebDefaults` should be used?

woehrl01:
@zcbenz `useWebDefaults` can be used if you need to port an existing html flexbox layout to Yoga. Without the need to watchout for places where the defaults differ from the web. You can even now (un)set `useWebDefaults` at a subtree level.

PR #449, opened 2017-03-01
Added bool config to YGConfig to configure using web defaults

woehrl01:
Added bool config to YGConfig to configure using web defaults. See #445.

### DIFF
--- a/yoga/Yoga.h
+++ b/yoga/Yoga.h
@@ -235,6 +235,10 @@ WIN_EXPORT void YGConfigSetExperimentalFeatureEnabled(const YGConfigRef config,
 WIN_EXPORT bool YGConfigIsExperimentalFeatureEnabled(const YGConfigRef config,
                                                      const YGExperimentalFeature feature);
 
+WIN_EXPORT void YGConfigSetUseWebDefaults(const YGConfigRef config, const bool enabled);
+
+WIN_EXPORT bool YGConfigGetUseWebDefaults(const YGConfigRef config);
+
 WIN_EXPORT void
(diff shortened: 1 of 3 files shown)


emilsjolander:
`align-items` also differs, web default is `flex-start`. This will also need to change the logic in YGNodeStyleGetFlexShrink to return `1` even though `flex` shorthand is set as well as return `auto` from YGNodeResolveFlexBasisPtr when `flex` shorthand is set.

woehrl01:
according to https://developer.mozilla.org/de/docs/Web/CSS/align-items it's ```stretch``` for ```align-items```. I already changed to return ```1``` on ```flex-shrink```. 
I changed ```YGNodeResolveFlexBasisPtr``` is this what you meant?

yoga/Yoga.h
```diff
@@ -235,6 +235,10 @@ WIN_EXPORT void YGConfigSetExperimentalFeatureEnabled(const YGConfigRef config,
 WIN_EXPORT bool YGConfigIsExperimentalFeatureEnabled(const YGConfigRef config,
                                                      const YGExperimentalFeature feature);
 
+WIN_EXPORT void YGConfigSetUseWebDefaults(const YGConfigRef config, const bool enabled);
```
emilsjolander: Can you add a comment here expressing that web defaults is the preferred configuration and non-web defaults is to be considered legacy?

emilsjolander:
👍 Thanks!

woehrl01:
Sounds like a good idea to port the tests to web defaults. But it seems that this could be a bit of work. So I'd suggest creating an issue, and I could have a look if I have a little more spare time.

emilsjolander:
Yeah, I might take a stab at it this week! 👍  Tracking in https://github.com/facebook/yoga/issues/451

PR #450, opened 2017-03-01
Use floats to prevent double calculation + float casting on scale

woehrl01:
Use ```float``` for calculation, as we would calculate with ```double``` and cast to float afterwards otherwise. So this removes the warning.


emilsjolander:
👍

Issue #454, opened 2017-03-02
Left, Right, Top and Bottom with percentages don't work when same element has flexDirection: 'row'


henrikra:
### Description

Here is example:
```javascript
import React from 'react';
import {AppRegistry, StyleSheet, View} from 'react-native';

const leftPercentBug = () => (
  <View style={styles.container}>
    <View style={styles.child}>
      <View style={styles.grandChild} />
      <View style={styles.grandChild} />
    </View>
  </View>
);

const styles = StyleSheet.create({
  container: {
    flex: 1,
    backgroundColor: 'red',
    borderWidth: 2,
    borderColor: 'black',
  },
  child: {
    flex: 1,
    backgroundColor: 'rgba(0, 0, 255, 0.5)',
    left: '50%',
    flexDirection: 'row',
  },
  grandChild: {
    backgroundColor: 'green',
    borderWidth: 1,
    borderColor: 'pink',
    flex: 1,
  },
});

AppRegistry.registerComponent('leftPercentBug', () => leftPercentBug);
```

With this code the layout **should** look something like this:
![image](https://cloud.githubusercontent.com/assets/12229968/23516753/c72f7756-ff76-11e6-848e-2ef6727c28ee.png)

Instead it looks like this:
![image](https://cloud.githubusercontent.com/assets/12229968/23516776/d83f7794-ff76-11e6-9b66-ee82badeb1a5.png)

So the `left` is not calculated right when the same element has `flexDirection: 'row'`
This same happened also when using `left`, `right`, `bottom` and `top` properties.

### Reproduction

Run the code above by yourself. Also you can see how the same should look in the web: https://jsfiddle.net/L8ps52sp/3/

### Solution

I think @emilsjolander should check this out since he added percentages in latest version ;)

### Additional Information

* React Native version: [0.42]
* Platform: [iOS]
* Operating System: [MacOS]

woehrl01:
@henrikra the first issue you have is that you assume that yoga uses the same defaults for layouting as in the web. [Sadly it's not.](https://facebook.github.io/yoga/docs/learn-more/#default-values) We are currently adding an option for using the defaults. If you apply the yoga-defaults to your jsfiddle, it certainly looks different from your expectation. 

Nevertheless it looks like that there could be a bug somewhere, too.

woehrl01:
@henrikra I extracted the following test case:

```html
<div id="percent_absolute_position" style="width: 60px; height: 50px; position: relative;">
[...]

PR #456, opened 2017-03-02
Add unittest for percentage width inside absolute layout

woehrl01:
Added unittest to constraint layout of percentage width inside absolute parent. See #454.

### DIFF
--- a/gentest/fixtures/YGPercentageTest.html
+++ b/gentest/fixtures/YGPercentageTest.html
@@ -101,7 +101,7 @@
   </div>
 </div>
 
-<div id="percent_absolute_position" style="width: 60px; height: 50px; position: relative;">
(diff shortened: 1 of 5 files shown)


gentest/fixtures/YGPercentageTest.html
```diff
@@ -90,4 +90,11 @@
         <div style="width:100%;"></div>
     </div>
     <div style="width: 100px;"></div>
+</div>
+
+<div id="percent_absolute_position" style="width: 60px; height: 50px; position: relative;">
```
emilsjolander: Why do you need `position: relative` here? this should be the default

woehrl01: Just wanted to make this explicit compared to the absolute node

Issue #452, opened 2017-03-01
Move PointScaleFactor to YGConfig


woehrl01:
As the configuration of rounding is already moved to the new ```YGConfig```, we should also move ```YGSetPointScaleFactor``` there. 

We also should need to take into account that the config could change across the nodes inside the tree. That means that we have to skip all the nodes where rounding is disabled or take different scale factors into account. The later should be an edge case, though.

PR #457, opened 2017-03-02
Move pointscalefactor to config

woehrl01:
This adds some improvements to the new ```YGConfig```, it tackles #452 and moves the scalefactor into the config.

### DIFF
--- a/yoga/Yoga.h
+++ b/yoga/Yoga.h
@@ -223,7 +223,7 @@ WIN_EXPORT void YGLog(YGLogLevel level, const char *message, ...);
 
 // Set this to number of pixels in 1 point to round calculation results
 // If you want to avoid rounding - set PointScaleFactor to 0
-WIN_EXPORT void YGSetPointScaleFactor(const YGConfigRef config, const float pixelsInPoint);
+WIN_EXPORT void YGConfigSetPointScaleFactor(const YGConfigRef config, const float pixelsInPoint);
 
 // YGConfig
 WIN_EXPORT YGConfigRef YGConfigNew(void);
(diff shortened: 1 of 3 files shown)


emilsjolander:
> On the other hand it removes the config from the method calls and use the nodes config instead. Which fixes the experiment if it's set on differently on nodes.

My bad I should have added a comment to the code. The whole idea was to adopt the root node's config for certain global flags when calculating layout. point scale factor is something I would consider a global flag. So are certain experiments. This was done to remove the need to pass a config into both the node constructor and the calculate layout function.

For changes like these I would appreciate splitting them up into separate pull requests. Makes for easier reviews and quicker merges 👍

woehrl01:
Will do in the next days!

woehrl01:
@emilsjolander I re-added the config argument, and moved ```pointScaleFactor``` to an argument for ```YGRoundToPixelGrid``` to use the value from the root node.

woehrl01:
I kept ```YGExperimentalFeatureWebFlexBasis``` and ```YGExperimentalFeatureMinFlexFix``` to use the value from each node, as I think those are values could be useful to set differently in a dom (e.g. for migrating some layouts)

yoga/Yoga.h
```diff
@@ -223,7 +223,7 @@ WIN_EXPORT void YGLog(YGLogLevel level, const char *message, ...);
 
 // Set this to number of pixels in 1 point to round calculation results
 // If you want to avoid rounding - set PointScaleFactor to 0
-WIN_EXPORT void YGSetPointScaleFactor(float pixelsInPoint);
+WIN_EXPORT void YGSetPointScaleFactor(const YGConfigRef config, const float pixelsInPoint);
```
emilsjolander: `YGConfigSetPointScaleFactor`

PR #463, opened 2017-03-06
[Nuget] Update nuget and sln

rmarinho:
Update nuget CI


rmarinho:
@splhack can we get this pull in so i can get 1.3.0 out ? would be awesome if we could recreate1.3.0 with this commit @emilsjolander so i can be in sync going forward, we will ship public nuget when you create a tag.

emilsjolander:
I can make a 1.3.1 release tomorrow which includes this. Ping me incase i forget :p

PR #471, opened 2017-03-09
Align resolve function names to have similiar namings

woehrl01:
We have some resolve functions with ```YG**Resolve``` and others named ```YGResolve**```. This changes both to be named like the later one, as I think this is the grammatically better readable one.


emilsjolander:
Thanks for improving the code quality!

PR #472, opened 2017-03-09
Update .travis.yml to build iOS sample project

maxoll:


dshahidehpour:
https://travis-ci.org/facebook/yoga/jobs/209746466

seems like this is still erroring the build

Issue #476, opened 2017-03-11
Web Inconsistency: Sometimes horizontally centered node has incorrect left


rigdern:
### Repro

The innermost node is being horizontally centered but it is given an incorrect left position. Here's a small repro which I extracted from my team's app:

```
<div style="height: 52px; width: 52px; position: absolute; align-items: center; justify-content: center; left: 72px;">
  <div style="align-items: center;">
    <div style="width: 52px; height: 52px; margin-left: 10px; margin-right: 10px;"></div>
  </div>
</div>
```

### Expected Output

```
{layout: {width: 52, height: 52, top: 0, left: 72}, children: [
  {layout: {width: 72, height: 52, top: 0, left: -10}, children: [
    {layout: {width: 52, height: 52, top: 0, left: 10}, },
  ]},
]},
```

### Actual Output

The innermost node's `left` is `0` but it should be `10`.

```
{layout: {width: 52, height: 52, top: 0, left: 72}, children: [
  {layout: {width: 72, height: 52, top: 0, left: -10}, children: [
    {layout: {width: 52, height: 52, top: 0, left: 0}, },
  ]},
]},
```

rigdern:
I confirmed this doesn't repro in master (bcc36cc).

PR #477, opened 2017-03-12
Fix align-content: center, flex-end alignment with margin

woehrl01:
This fixes ```align-content: center``` and ```align-content: flex-end``` when the child exceeds the parents size. See #476. It also fixes those layouts if the child has ```margin: auto``` set.

### DIFF
--- a/csharp/tests/Facebook.Yoga/YGAlignItemsTest.cs
+++ b/csharp/tests/Facebook.Yoga/YGAlignItemsTest.cs
@@ -1587,5 +1587,64 @@ namespace Facebook.Yoga
             Assert.AreEqual(20f, root_child3.LayoutHeight);
         }
 
+        [Test]
+        public void Test_align_items_center_nested()
+        {
+            YogaConfig config = new YogaConfig();
+
+            YogaNode root = new YogaNode(config);
+            root.JustifyContent = YogaJustify.Center;
+            root.AlignItems = YogaAlign.Center;
+            root.PositionType = YogaPositionType.Absolute;
+            root.Width = 52;
+            root.Height = 52;
+
+            YogaNode root_child0 = new YogaNode(config);
+            root_child0.AlignItems = YogaAlign.Center;
+            root.Insert(0, root_child0);
+
+            YogaNode root_child0_child0 = new YogaNode(config);
+            root_child0_child0.MarginLeft = 10;
+            root_child0_child0.MarginRight = 10;
+            root_child0_child0.Width = 52;
+            root_child0_child0.Height = 52;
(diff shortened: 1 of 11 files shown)


woehrl01:
> I am not 100% certain with this change, as there isn't any failing test after removing those lines. According to git those lines have been added by @rigdern . Do you have an example layout where this line are required?

After looking deeper into this, and trying to add failing tests, I found that it fails even without margin set, and also has an addional bug there with ```margin: auto```. I'm now a lot more certain that this is a valid fix.

rigdern:
> I am not 100% certain with this change, as there isn't any failing test after removing those lines. According to git those lines have been added by @rigdern . Do you have an example layout where this line are required?

@woehrl01 When you say "removing those lines", I believe you are referring to:

```
if (measureModeCrossDim == YGMeasureModeAtMost) {
  containerCrossAxis = fminf(containerCrossAxis, availableInnerCrossDim);
}
```

I don't have a specific example available. It looks like those lines were added in #185 which involved significant changes to the layout engine in order to make it better match the [W3C flexbox spec](https://www.w3.org/TR/css-flexbox-1/).

#476 didn't repro in the version of the layout engine that is included in React Native 0.32.0-rc.0. So it seems those lines didn't break this scenario back then.

If you delete those lines, what prevents the calculated cross size from exceeding the available cross size when running in the at-most measure mode?

woehrl01:
> #476 didn't repro in the version of the layout engine that is included in React Native 0.32.0-rc.0. So it seems those lines didn't break this scenario back then.

That's strange. 

> If you delete those lines, what prevents the calculated cross size from exceeding the available cross size when running in the at-most measure mode?

Nothing prevents it. But as the resulting value is only used for the ```alignItem```/```margin: auto``` step, this value shouldn't be clipped, as the child is still bigger then ```AT_MOST``` wants it.

I just want to share my findings, as I'm still not that convinced. But I also don't understand why that value should be clipped either.

rigdern:
@woehrl01 The test case I listed in #476 is still laid out incorrectly in master.

woehrl01:
@rigdern that's strange, as I added [a test case with exactly your values](https://github.com/facebook/yoga/blob/master/gentest/fixtures/YGAlignItemsTest.html#L147-L151), which doesn't fail in the master. Could you please verify if I correctly adapted your failing layout?

rigdern:
@woehrl01 The test case I reported used a `position: absolute` element on the root with a non-zero value for `left`. Here's the diff: https://www.diffchecker.com/BD9BgS4m

woehrl01:
@rigdern Oh, thanks, I see! It's not the root node which isn't laid out correctly, I'll have a look. 🔨

Issue #453, opened 2017-03-02
Clean up printing code


emilsjolander:
We have some functions for debugging to print the node hierarchy but the code is quite a mess and is hard to keep up to date (it doesn't support auto units for example). Not sure how to improve it but the current state is not great.

PR #479, opened 2017-03-13
Optimize log print by using html format

woehrl01:
See facebook/yoga#453. Optimizes the node log print by generating some enum text via ```enum.py``` and moving printing to new functions to reduce boilerplate code. 

Changes the log output to format the nodes in html to be able to copy paste it  into browsers for quick debugging. 

Hides all default values.

### DIFF
--- a/enums.py
+++ b/enums.py
@@ -125,6 +125,16 @@ def to_java_upper(symbol):
         out += c.upper()
     return out
 
+def to_log_lower(symbol):
+    symbol = str(symbol)
+    out = ''
+    for i in range(0, len(symbol)):
+        c = symbol[i]
+        if str.istitle(c) and i is not 0 and not str.istitle(symbol[i - 1]):
+            out += '-'
+        out += c.lower()
+    return out
+
 
 root = os.path.dirname(os.path.abspath(__file__))
 
@@ -143,8 +153,27 @@ with open(root + '/yoga/YGEnums.h', 'w') as f:
             else:
                 f.write('  YG%s%s,\n' % (name, value))
         f.write('} YG_ENUM_END(YG%s);\n' % name)
+        f.write('WIN_EXPORT const char *YG%sToString(const YG%s value);\n' % (name, name))
         f.write('\n')
     f.write('YG_EXTERN_C_END\n')
+
+# write out C body for printing
+with open(root + '/yoga/YGEnums.c', 'w') as f:
+    f.write(LICENSE)
+    f.write('#include "YGEnums.h"\n\n')
+    for name, values in sorted(ENUMS.items()):
+        f.write('const char *YG%sToString(const YG%s value){\n' % (name, name))
+        f.write('  switch(value){\n')
+        for value in values:
+            if isinstance(value, tuple):
+                f.write('    case YG%s%s:\n' % (name, value[0]))
+                f.write('      return "%s";\n' % to_log_lower(value[0]))
+            else:
+                f.write('    case YG%s%s:\n' % (name, value))
+                f.write('      return "%s";\n' % to_log_lower(value))
+        f.write('  }\n')
+        f.write('  return "unknown";\n')
+        f.write('}\n\n')
 
 # write out java files
 for name, values in sorted(ENUMS.items()):
--- a/tests/YGLoggerTest.cpp
+++ b/tests/YGLoggerTest.cpp
@@ -1,0 +1,68 @@
+/**
+ * Copyright (c) 2014-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#include <gtest/gtest.h>
+#include <yoga/Yoga.h>
+#include <stdarg.h>
+
+namespace {
+  char writeBuffer[4096];
+  int _unmanagedLogger(YGLogLevel level, const char *format, va_list args) {
+    return vsnprintf(writeBuffer + strlen(writeBuffer), sizeof(writeBuffer) - strlen(writeBuffer), format, args);
+  }
+}
+
+TEST(YogaTest, logger_default_node_should_print_no_style_info) {
+  writeBuffer[0] = '\0';
+  YGSetLogger(_unmanagedLogger);
+  const YGNodeRef root = YGNodeNew();
+  YGNodeCalculateLayout(root, YGUnitUndefined, YGUnitUndefined, YGDirectionLTR);
+  YGNodePrint(root, (YGPrintOptions)(YGPrintOptionsLayout | YGPrintOptionsChildren | YGPrintOptionsStyle));
+  YGSetLogger(NULL);
+  YGNodeFree(root);
+  
+  const char * expected = "<div layout=\"width: 0; height: 0; top: 0; left: 0;\" style=\"\" ></div>";
+  ASSERT_STREQ(expected, writeBuffer);
+}
+
+TEST(YogaTest, logger_node_with_percentage_absolute_position_and_margin) {
+  writeBuffer[0] = '\0';
+  YGSetLogger(_unmanagedLogger);
+  const YGNodeRef root = YGNodeNew();
+  YGNodeStyleSetPositionType(root, YGPositionTypeAbsolute);
+  YGNodeStyleSetWidthPercent(root, 50);
+  YGNodeStyleSetHeightPercent(root, 75);
+  YGNodeStyleSetFlex(root, 1);
+  YGNodeStyleSetMargin(root, YGEdgeRight, 10);
+  YGNodeStyleSetMarginAuto(root, YGEdgeLeft);
+  YGNodeCalculateLayout(root, YGUnitUndefined, YGUnitUndefined, YGDirectionLTR);
+  YGNodePrint(root, (YGPrintOptions)(YGPrintOptionsLayout | YGPrintOptionsChildren | YGPrintOptionsStyle));
+  YGSetLogger(NULL);
+  YGNodeFree(root);
+  
+  const char * expected = "<div layout=\"width: 0; height: 0; top: 0; left: 0;\" style=\"flex: 1; margin-left: auto; margin-right: 10px; width: 50%; height: 75%; position: absolute; \" ></div>";
+  ASSERT_STREQ(expected, writeBuffer);
+}
+
+TEST(YogaTest, logger_node_with_children_should_print_indented) {
+  writeBuffer[0] = '\0';
+  YGSetLogger(_unmanagedLogger);
+  const YGNodeRef root = YGNodeNew();
+  const YGNodeRef child0 = YGNodeNew();
+  const YGNodeRef child1 = YGNodeNew();
+  YGNodeInsertChild(root, child0, 0);
+  YGNodeInsertChild(root, child1, 1);
+  YGNodeCalculateLayout(root, YGUnitUndefined, YGUnitUndefined, YGDirectionLTR);
+  YGNodePrint(root, (YGPrintOptions)(YGPrintOptionsLayout | YGPrintOptionsChildren | YGPrintOptionsStyle));
+  YGSetLogger(NULL);
+  YGNodeFreeRecursive(root);
+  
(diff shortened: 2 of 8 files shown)


yoga/YGEnumsPrint.h
```diff
@@ -0,0 +1,359 @@
...
+ */
+
+#pragma once
+
+#include "YGMacros.h"
+#include "Yoga.h"
+
+YG_EXTERN_C_BEGIN
+
+void YGLogAlign(YGLogLevel logLevel, const char * param, YGAlign value){
```
emilsjolander: What is `param` ? Could you choose a more descriptive name?

emilsjolander: Also could you mark all parameters as const?

yoga/YGEnumsPrint.h
```diff
@@ -0,0 +1,359 @@
...
+      break;
+    case YGAlignSpaceBetween:
+      text = "space-between";
+      break;
+    case YGAlignSpaceAround:
+      text = "space-around";
+      break;
+  }
+  if(param == NULL) {
+    YGLog(logLevel, text);
```
emilsjolander: I would rather the functions in this file not return log but instead return a statically allocated string which is logged by the caller.

yoga/YGEnumsPrint.h
emilsjolander: File is names `YGEnumsPrint` but functions are named `YGLogX`. The mix of `Print` and `Log` is a bit odd.

enums.py
```diff
@@ -146,6 +156,34 @@ def to_java_upper(symbol):
         f.write('\n')
     f.write('YG_EXTERN_C_END\n')
 
+# write out C headers for printing
+with open(root + '/yoga/YGEnumsPrint.h', 'w') as f:
```
emilsjolander: Could this just be part of YGEnums.h? Collocating this stuff is usually pretty nice

yoga/Yoga.c
```diff
@@ -781,14 +772,46 @@ static void YGPrintNumberIfNotUndefinedf(const char *str, const float number) {
...
+    YG_LOG_PRINT_NONZERO_EDGE(type, Right);                                                  \
+    YG_LOG_PRINT_NONZERO_EDGE(type, Left);                                                   \
+    YG_LOG_PRINT_NONZERO_EDGE(type, Top);                                                    \
+    YG_LOG_PRINT_NONZERO_EDGE(type, Bottom);                                                 \
+    YG_LOG_PRINT_NONZERO_EDGE(type, Start);                                                  \
+    YG_LOG_PRINT_NONZERO_EDGE(type, End);                                                    \
+  \
+}
+
 static bool YGFourValuesEqual(const YGValue four[4]) {
```
emilsjolander: This function doesn't even handle start & end or vertical & horizontal so would love it fixes for that were included here as well 👍

yoga/Yoga.c
```diff
@@ -781,14 +772,46 @@ static void YGPrintNumberIfNotUndefinedf(const char *str, const float number) {
...
 }
 
+#define YG_STRING(s) #s
+
+#define YG_LOG_PRINT_NONZERO_EDGE(type, edge)           \
+  \
+YGPrintNumberIfNotZero(YG_STRING(type) YG_STRING(edge), \
+                       YGComputedEdgeValue(node->style.type, YGEdge##edge, &YGValueZero));
+
+#define YG_LOG_PRINT_NONZERO_EDGES(type)                                                     \
```
emilsjolander: This could probably be a function instead of a macro which takes a array of YGValues

woehrl01:
Thanks for the comments, I'll try to work these in.

Additonally I'm thinking if it would be more useful if we output those values as html instead of JSONlike, to be able do copy paste possible wrong layouts more easily. 

But I don't know how the output is used currently (e.g. ReactNative).

yoga/YGEnumsPrint.h
woehrl01: I also thought of this, but this makes the callside a bit more boilerplate, not sure if we want this.

yoga/Yoga.c
woehrl01: this only works nicely if we write the position in lowercase, example: ```margin-left``` instead of current ```marginLeft```. Could be nicer if we write html instead (see suggestion below)

emilsjolander:
Outputting a html layout would be amazing. This isn't really used a lot so I think outputting html will make it much more useful. Love the idea.

yoga/Yoga.c
emilsjolander: Not sure I fully understand

yoga/YGEnumsPrint.h
emilsjolander: Could you give it a try? I think it might be worth it. You can always abstract if if(param == null) part into another function.

yoga/Yoga.c
woehrl01: If we loop over the array we could use ```YGLogEdge``` to stringify ```YGEdge``` but that returns only the edge in lowercase. So we need some fiddeling to uppercase if we want to stick to something like ```marginLeft```. 

But as you think that an html layout would be a great improvment, this is more easily doable (no need to uppercase). I'll update the PR with your suggestions and we can discuss on top of that, what do you think?

emilsjolander: sounds good!

woehrl01:
I added html output and removed the usage for a macro, there are some small things which aren't that nice yet (e.g. ```""``` usage). Those will be removed later, when I find a better way for this.

> This function doesn't even vertical & horizontal ...

I can't add those as accessing this values lead to an ```YG_ASSERT``` error.

> Could this just be part of YGEnums.h? Collocating this stuff is usually pretty nice

not really as we get some linker errors later, we would need to add the function templates to the header and provide an additional ```*.c``` file. Should we go this way?

> File is names YGEnumsPrint but functions are named YGLogX

I change this if we want to keep it that way, depends on the previous.

woehrl01:
I'm not sure if we should print ```pt```, if we want to use the output as input for gentest, we should print ```px```, do not get different outputs based on those different scale factors. What do you think?

emilsjolander:
> we would need to add the function templates to the header and provide an additional *.c file. Should we go this way?

Yeah i'm totally ok with a `YGEnums.c` file.

> I'm not sure if we should print pt, if we want to use the output as input for gentest, we should print px, to not get different outputs based on those different scale factors. What do you think?

perhaps use px if scale factor is 1 otherwise pt? 

> I can't add those as accessing this values lead to an YG_ASSERT error.

Actually we probably should not be calling `YGComputedEdgeValue()` here are we want to print the value which was set on the style, not the computed value.

woehrl01:
> perhaps use px if scale factor is 1 otherwise pt?

isn't css using different factors for pt and px? I'd prefere sticking with just px.

> print the value which was set on the style, not the computed value.

correct, I'll change it accordingly.

emilsjolander:
> isn't css using different factors for pt and px? I'd prefere sticking with just px.

 Ah yeah that sounds good.

emilsjolander:
@woehrl01 not sure. Also not sure why there is no output in the logs. Can you run the tests locally?

tests/YGLoggerTest.cpp
```diff
@@ -0,0 +1,63 @@
...
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#include <gtest/gtest.h>
+#include <yoga/Yoga.h>
+
+static char writeBuffer[4096];
+static int _unmanagedLogger(YGLogLevel level, const char *format, va_list args) {
```
emilsjolander: va_list is defined in stdarg.h right?

yoga/YGEnums.c
```diff
@@ -0,0 +1,234 @@
...
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#include "YGEnums.h"
+
+const char * YGStringifyAlign(const YGAlign value){
```
emilsjolander: `YGAlignToString`

emilsjolander: Also `char *YGStringifyAlign` instead of `char * YGStringifyAlign` (asterix placement)

yoga/YGEnums.c
```diff
@@ -0,0 +1,234 @@
...
+      return "flex-end";
+    case YGAlignStretch:
+      return "stretch";
+    case YGAlignBaseline:
+      return "baseline";
+    case YGAlignSpaceBetween:
+      return "space-between";
+    case YGAlignSpaceAround:
+      return "space-around";
+    default:
```
emilsjolander: Is the default needed? I think this switch should be exhaustive. Not supplying default will allow the compiler to warn when not all cases are specified.

yoga/YGEnums.h
```diff
@@ -132,4 +132,20 @@ typedef YG_ENUM_BEGIN(YGWrap) {
   YGWrapWrapReverse,
 } YG_ENUM_END(YGWrap);
 
+
+WIN_EXPORT const char * YGStringifyAlign(const YGAlign value);
```
emilsjolander: `char *YG...`

emilsjolander: Can you put each toString function bellow the corresponding enum declaration?

yoga/YGEnums.c
woehrl01: not exactly as a default. but at least as a "fall-through". Will remove the default and put the return "unknown" after the switch

yoga/YGEnums.h
woehrl01: will do

tests/YGLoggerTest.cpp
woehrl01: should I include this header?

woehrl01:
Current output looks like:

```html
<div layout="width: 0; height: 0; top: 0; left: 0;" style="flex-direction: column; justify-content: flex-start; align-items: stretch; align-content: flex-start; flex-grow: 0; flex-shrink: 0; " >
  <div layout="width: 0; height: 0; top: 0; left: 0;" style="flex-direction: column; justify-content: flex-start; align-items: stretch; align-content: flex-start; flex-grow: 0; flex-shrink: 0; " ></div>
  <div layout="width: 0; height: 0; top: 0; left: 0;" style="flex-direction: column; justify-content: flex-start; align-items: stretch; align-content: flex-start; flex-grow: 0; flex-shrink: 0; " ></div>
</div>
```

tests/YGLoggerTest.cpp
emilsjolander: yeah 👍

yoga/YGEnums.c
emilsjolander: What is the benefit of this?

emilsjolander:
> Current output looks like:

Lovely!

yoga/YGEnums.c
woehrl01: the compiler needs it, the function doesn't have a return statement otherwise for any call case. (e.g. calling it an invalid argument)

woehrl01:
should I hide all default values e.g. for ```flex-direction```, ```justify-content```, too?

What about ```flex``` ?

emilsjolander:
> should I hide all default values

Good point. I think hiding defaults is good.

> What about flex

Let's print it if it is set. negative or not.

woehrl01:
Found the root cause of the failing travis test. All DEATH tests should be named like ```*DeathTest``` so they are executed at first, as they are run l later, they output seems to be overridden by our new logger tests. See: https://github.com/google/googletest/blob/master/googletest/docs/AdvancedGuide.md

arcanis:
The JSON format had nice properties (for example, extracting a tree from an application and recreating it standalone via the JS bindings in order to better debug the library). Is there plans to support multiple logging outputs?

woehrl01:
Multiple logging outputs could be doable. Did you use the JSONlike output? As the output wasn't exactly JSON ```'``` instead of ```"``` for properties. Even so it could be nice to use the output directly for javascript. I think the HTML output is more generic as you could use ```gentest.py``` to generate your DOM easily for any supported language.

emilsjolander:
@woehrl01 I agree we should stick with html output for now.
@arcanis I'm definetly not against multiple output formats but for now html is more useful

PR #480, opened 2017-03-13
Build iOS sample project with Travis

maxoll:
- Updated `.travis.yml` to install xcpretty, cocoapods, project dependencies, and build.
- Removed the sample project's `Podfile.lock` because it should always pull in the latest Cocoapods dependencies. This way, when we update YogaKit's version, Travis will automatically build the sample project with the new version. Example: [RestKit](https://github.com/RestKit/RestKit/blob/development/Examples/RKTwitter/.gitignore#L1)
- To make sure this works, please don't merge until Travis runs this PR. Thanks!


emilsjolander:
Could you rebase?

Issue #483, opened 2017-03-16
Don't transfer layout outputs to java if they didn't change


emilsjolander:
When setting layout outputs on a node also set a flag indicating if the layout changed. Read this flag before transferring outputs to java. JNI calls are very expensive so the less of them we need to do the better. Possibly re-use the hasNewLayout flag.

ptomasroos:
Is this the issue for the reverted commit of https://github.com/facebook/yoga/commit/b283572453c588df8a4b0ced5ebe5a50a1bc067f @emilsjolander ? Just trying to track progress waiting for an update in order to be able to ship a release.

emilsjolander:
Not exactly. But it is part of it. Reverted commit will reland this week.

ptomasroos:
Thanks for the info @emilsjolander ! Tracking commits into master as well then :)

PR #484, opened 2017-03-16
Don't transfer layout outputs to java for unset edges

woehrl01:
See facebook/yoga#483. We should not transfer the layout if the layout didn't change. (using ```hasNewLayout```). This also changes that the lock on the java node is only aquired if needed, and it holds the lock only for the time the values are set and not for the time all it's children are set.

### DIFF
--- a/java/com/facebook/yoga/YogaNode.java
+++ b/java/com/facebook/yoga/YogaNode.java
@@ -42,9 +42,14 @@ public class YogaNode implements YogaNodeAPI<YogaNode> {
   private long mNativePointer;
   private Object mData;
 
-  private boolean mHasSetPadding = false;
-  private boolean mHasSetMargin = false;
-  private boolean mHasSetBorder = false;
+  /* Those flags needs be in sync with YGJNI.cpp */
+  private final static int MARGIN = 1;
+  private final static int PADDING = 2;
+  private final static int BORDER = 4;
+
+  @DoNotStrip
+  private int mEdgeSetFlag = 0;
+
   private boolean mHasSetPosition = false;
 
   @DoNotStrip
@@ -113,9 +118,7 @@ public class YogaNode implements YogaNodeAPI<YogaNode> {
   private native void jni_YGNodeReset(long nativePointer);
   @Override
   public void reset() {
-    mHasSetPadding = false;
-    mHasSetMargin = false;
-    mHasSetBorder = false;
+    mEdgeSetFlag = 0;
     mHasSetPosition = false;
     mHasNewLayout = true;
 
@@ -382,7 +385,7 @@ public class YogaNode implements YogaNodeAPI<YogaNode> {
   private native Object jni_YGNodeStyleGetMargin(long nativePointer, int edge);
   @Override
   public YogaValue getMargin(YogaEdge edge) {
-    if (!mHasSetMargin) {
+    if (!((mEdgeSetFlag & MARGIN) == MARGIN)) {
       return YogaValue.UNDEFINED;
     }
     return (YogaValue) jni_YGNodeStyleGetMargin(mNativePointer, edge.intValue());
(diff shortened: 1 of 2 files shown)


emilsjolander:
I'm not sure we can directly use hasNewValue as it is not ever set to false in the yoga code. Yoga.c definetly needs some changes as well. There are also times when the padding of a node could potentially have change but not the width which me may or may not want to handle.

woehrl01:
We could do the following: We use ```hasNewLayout``` just for jni, and keep the value in java as a simple property (no jni callback) and set this value from ```YGTransferLayoutOutputsRecursive``` as well. What do you think?

emilsjolander:
Good idea! I think that should work. React-native depends quite heavily on that value so while Yoga's tests will pass I'll need to run through some internal RN apps to check that nothing unexpected breaks. But I think this idea will probably work.

woehrl01:
I also added that we reuse the ```mHasSetMargin/Padding/Border``` flags. We make 3 calls more than usual, but as I think that not all node have all of these values set (if at all) we at safe 9 jni calls in the best case.

woehrl01:
Further reduces the jni calls by 2 for the edges. By moving from bool to an int flag which stores all set edges. We could even provide a method which does something like ```TouchFromJni()``` which sets the ```hasNewLayout``` flag and returns all the flags in a single call.

java/com/facebook/yoga/YogaNode.java
```diff
@@ -42,9 +42,13 @@ public static void setLogger(YogaLogger logger) {
   private long mNativePointer;
   private Object mData;
 
-  private boolean mHasSetPadding = false;
-  private boolean mHasSetMargin = false;
-  private boolean mHasSetBorder = false;
+  private static int MARGIN = 1;
```
astreet: make these final

java/jni/YGJNI.cpp
```diff
@@ -24,54 +24,74 @@ static void YGTransferLayoutDirection(YGNodeRef node, alias_ref<jobject> javaNod
...
+
+      static auto borderLeftField = obj->getClass()->getField<jfloat>("mBorderLeft");
+      static auto borderTopField = obj->getClass()->getField<jfloat>("mBorderTop");
+      static auto borderRightField = obj->getClass()->getField<jfloat>("mBorderRight");
+      static auto borderBottomField = obj->getClass()->getField<jfloat>("mBorderBottom");
+
+      static auto hasNewLayoutField = obj->getClass()->getField<jboolean>("mHasNewLayout");
+      static auto edgeSetFlagField = obj->getClass()->getField<jint>("mEdgeSetFlag");
+
+      const int MARGIN = 1;
```
astreet: comment that these need to be kept in sync with YogaNode.java (or pull them off that class)

astreet:
Nice, I like this: can we do the hasNewLayout change in a separate PR though? I ask in case something breaks here we can debug easier and revert a more isolated diff.

woehrl01:
@astreet Sure, will put up a separte PR and apply the changes tonight. Thanks!

woehrl01:
On the other hand we could provide a ```ExperimentalFeatureOptimizeJniTransfer``` which we could enable/disable if anything breaks.

astreet:
Thanks! I'm fine with this not being an experimental flag :)

emilsjolander:
I tried landing a much smaller scope version of this earlier which broke a bunch of stuff. I'm going to try to land that first still before landing this but I really like what you have done here. I'll ping you when it is in.

emilsjolander:
Can you rebase this? 👍

java/jni/YGJNI.cpp
```diff
@@ -46,36 +46,51 @@ static void YGTransferLayoutOutputsRecursive(YGNodeRef root) {
...
       YGTransferLayoutDirection(root, obj);
       YGNodeSetHasNewLayout(root, false);
-      for (uint32_t i = 0; i < YGNodeGetChildCount(root); i++) {
-        YGTransferLayoutOutputsRecursive(YGNodeGetChild(root, i));
-      }
     } else {
       YGLog(YGLogLevelError, "Java YGNode was GCed during layout calculation\n");
+      return;
+    }
+    for (uint32_t i = 0; i < YGNodeGetChildCount(root); i++) {
```
emilsjolander: This feels like a rebase error

woehrl01: This actually isn't one, I think we don't need to lock each object for all it's children. Thats the reason I put the iteration outside the lock. But maybe this should be a separate PR?

emilsjolander: I got confused by which if statement it was inside of 👍  but yeah if you could make that a separate PR that would be awesome!

woehrl01: of course 👍

PR #467, opened 2017-03-08
Take margin into account on max dimension

woehrl01:
We need to take the margin into account if we clip on max dimension. Fixes #466.

PR #487, opened 2017-03-21
Fix tests of splitted config feature

arcanis:
The following PR fixes the tests used in the javascript port (it modifies gentest.rb).

These changes don't yet pass, it seems something is segfaulting somewhere. I have to check if it comes from nbind, the yoga library, or the node bridge itself. There's also some fails on the browser build, but it might be the same issue.

### DIFF
--- a/javascript/binding.gyp
+++ b/javascript/binding.gyp
@@ -9,6 +9,7 @@
 	"sources": [
             "sources/yoga/YGNodeList.c",
             "sources/yoga/Yoga.c",
+	    "sources/Config.cc",
             "sources/Node.cc",
             "sources/global.cc",
 	    "sources/nbind.cc"
--- a/javascript/binding.gyp
+++ b/javascript/binding.gyp
@@ -9,7 +9,7 @@
 	"sources": [
             "sources/yoga/YGNodeList.c",
             "sources/yoga/Yoga.c",
-	    "sources/Config.cc",
+            "sources/Config.cc",
             "sources/Node.cc",
             "sources/global.cc",
 	    "sources/nbind.cc"
--- a/javascript/tests/Facebook.Yoga/YGAbsolutePositionTest.js
+++ b/javascript/tests/Facebook.Yoga/YGAbsolutePositionTest.js
@@ -12,6 +12,8 @@
 var Yoga = Yoga || require("../../sources/entry-" + process.env.TEST_ENTRY);
 
 it("absolute_layout_width_height_start_top", function () {
+  const config = Yoga.Config.create();
+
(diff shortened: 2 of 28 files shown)


arcanis:
@emilsjolander @woehrl01 I found the issue! It was caused by a macro that was set at the wrong place. Because of this, and because nbind includes quite a few class implementations in its headers, different files were using different implementations of a few functions. Pfew.

This PR fixes this, updates the tests, and fix the node creation when passing a config as parameter. However, I've just rebased with master and it seems that some tests have been recently added and are now failing - does it ring a bell?

The failing tests are the following:

```
  1) align_items_center_child_with_margin_bigger_than_parent
  2) align_items_flex_end_child_with_margin_bigger_than_parent
  3) align_items_center_child_without_margin_bigger_than_parent
  4) align_items_flex_end_child_without_margin_bigger_than_parent
  5) margin_should_not_be_part_of_max_height
  6) margin_should_not_be_part_of_max_width
  7) margin_auto_left_right_child_bigger_than_parent
  8) margin_auto_left_child_bigger_than_parent
  9) margin_fix_left_auto_right_child_bigger_than_parent
  10) margin_auto_left_fix_right_child_bigger_than_parent
```

woehrl01:
Great job @arcanis ! The failing tests are related to the changes made in pr #477 and #467.

javascript/tests/Facebook.Yoga/YGAbsolutePositionTest.js
```diff
@@ -48,12 +50,15 @@ it("absolute_layout_width_height_start_top", function () {
     console.assert(10 === root_child0.getComputedWidth(), "10 === root_child0.getComputedWidth() (" + root_child0.getComputedWidth() + ")");
     console.assert(10 === root_child0.getComputedHeight(), "10 === root_child0.getComputedHeight() (" + root_child0.getComputedHeight() + ")");
   } finally {
-    if (typeof root !== "undefined") {
+    if (typeof root !== "undefined")
```
arcanis: 👍

javascript/binding.gyp
```diff
@@ -9,6 +9,7 @@
 	"sources": [
             "sources/yoga/YGNodeList.c",
             "sources/yoga/Yoga.c",
+	    "sources/Config.cc",
```
arcanis: 👍

arcanis:
Everything should be ok 👌 

Apparently my build failed to notice that the Yoga files had changed, hence why the tests checking the new behaviour weren't passing. Everything worked fine after clearing the `build` directory.

PR #492, opened 2017-03-27
Don't transfer layout outputs to java for nodes which don't have a new layout

woehrl01:
As suggested in facebook/yoga#484. This is the PR which only adds the part of using a local bool field for storing the hasLayoutFlag, to be able to pass the layout only if it has really changed.

### DIFF
--- a/java/com/facebook/yoga/YogaNode.java
+++ b/java/com/facebook/yoga/YogaNode.java
@@ -82,7 +82,7 @@ public class YogaNode implements YogaNodeAPI<YogaNode> {
   @DoNotStrip
   private int mLayoutDirection = 0;
   @DoNotStrip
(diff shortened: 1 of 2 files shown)


java/com/facebook/yoga/YogaNode.java
```diff
@@ -81,6 +81,8 @@ public static void setLogger(YogaLogger logger) {
   private float mBorderBottom = 0;
   @DoNotStrip
   private int mLayoutDirection = 0;
+  @DoNotStrip
+  private bool mHasNewLayout = true;
```
emilsjolander: boolean

emilsjolander:
I don't think your last diff is needed. Can you give an example where it would break without it?

PR #495, opened 2017-03-29
Fix xmlns import for android yoga library

woehrl01:
Fix xmlns import for yoga library by using `apk` instead of `lib` in the comment.

### DIFF
--- a/android/src/main/java/com/facebook/yoga/android/YogaLayout.java
+++ b/android/src/main/java/com/facebook/yoga/android/YogaLayout.java
@@ -52,7 +52,7 @@ import com.facebook.yoga.YogaWrap;
  * <pre>{@code
  * <YogaLayout
  *     xmlns:android="http://schemas.android.com/apk/res/android"
- *     xmlns:yoga="http://schemas.android.com/lib/com.facebook.yoga.android"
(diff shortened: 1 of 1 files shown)


emilsjolander:
Actually shouldn't it be apk/lib?

cc @rspencer01

woehrl01:
could be too, I'm not that familiar with android, but at least this was the thing I had to change to make a working/compilable example project.

rspencer01:
I am not certain, but a google search suggests `apk/res`.  Give that a go?  I'd be more happy with that than `apk`.

woehrl01:
How about `http://schemas.android.com/apk/res-auto` ? Which is the suggestion Android Studio makes.

rspencer01:
I object to that because it pulls in everything.  We had trouble back when we had attributes like

    yoga:height

that conflicted with some other library's definition of `height`.  We don't want to do this to other people (as unlikely as it may be).  What was it breaking, out of interest?

woehrl01:
It didn't compile 😅

I created a complete new project with a different namespace, added the gradle imports, copied the example xml layout. And this was the necessary change which costed me some hours 😂

rspencer01:
Ok.  Well, this is a minor change, its really not worth anybody losing sleep about.

Issue #488, opened 2017-03-22
YGNodeSetMeasureFunc api error 


anql:
# Report

-  I have searched [existing issues](https://github.com/facebook/yoga/issues) and this is not a duplicate

# Issues and Steps to Reproduce
1. Run the code below
2. Uncomment  the line of 'YGNodeSetMeasureFunc', run again.

# Expected Behavior
Expected the child node's width should be 282.
```
{layout: {width: 282, height: 200, top: 0, left: 0}, flexDirection: 'column', alignItems: 'stretch', alignContent: 'stretch', flexGrow: 0, flexShrink: 0, flexBasis: nan%, overflow: 'hidden', width: nan%, height: 200pt, children: [
  {layout: {width: 282, height: 200, top: 0, left: 0}, flexDirection: 'column', alignItems: 'stretch', alignContent: 'stretch', alignSelf: 'stretch', flexGrow: 0, flexShrink: 0, flexBasis: nan%, overflow: 'hidden', padding: 100pt, width: nan%, height: nan%, },
]},
```
# Actual Behavior
Get different calculation results while set 'YGNodeSetMeasureFunc' or not even if the function was never invoked actually both cases.
```
{layout: {width: 282, height: 200, top: 0, left: 0}, flexDirection: 'column', alignItems: 'stretch', alignContent: 'stretch', flexGrow: 0, flexShrink: 0, flexBasis: nan%, overflow: 'hidden', width: nan%, height: 200pt, children: [
  {layout: {width: 200, height: 200, top: 0, left: 0}, flexDirection: 'column', alignItems: 'stretch', alignContent: 'stretch', alignSelf: 'stretch', flexGrow: 0, flexShrink: 0, flexBasis: nan%, overflow: 'hidden', padding: 100pt, width: nan%, height: nan%, },
]},
```
# Link to Code
```C
#include <iostream>
#include "Yoga.h"
#include <stdlib.h>
#include <stdio.h>
#include <vector>


static YGSize YGMeasureFuncForLabel(YGNodeRef node,
                                    float width,
                                    YGMeasureMode widthMode,
                                    float height,
                                    YGMeasureMode heightMode) {
    // this function was never invoked.
    YGSize size;
    size.width = 8;
    size.height = 16;
    return size;
}

int main(int argc, const char * argv[]) {
    
    const YGNodeRef root = YGNodeNew();
    YGNodeStyleSetHeight(root, 200);
[...]

PR #499, opened 2017-04-02
Let measure behave more like on the web

woehrl01:
Nodes with a measure function needs to be measured even so it seems there is no available space. So it behaves more like on the web. Fix facebook/yoga#488

### DIFF
--- a/tests/YGMeasureTest.cpp
+++ b/tests/YGMeasureTest.cpp
@@ -21,7 +21,7 @@ static YGSize _measure(YGNodeRef node,
   }
 
   return YGSize{
-      width = 10, height = 10,
+      .width = 10, .height = 10,
   };
 }
 
--- a/tests/YGMeasureTest.cpp
+++ b/tests/YGMeasureTest.cpp
@@ -25,6 +25,23 @@ static YGSize _measure(YGNodeRef node,
   };
 }
 
+static YGSize _simulate_wrapping_text(YGNodeRef node,
+                       float width,
+                       YGMeasureMode widthMode,
+                       float height,
+                       YGMeasureMode heightMode) {
+  if(widthMode == YGMeasureModeUndefined || width >= 68)
+  {
+    return YGSize{
+      .width = 68, .height = 16
+    };
+  }
+
+  return YGSize{
+      .width = 50, .height = 32,
+  };
+}
+
 TEST(YogaTest, dont_measure_single_grow_shrink_child) {
   const YGNodeRef root = YGNodeNew();
   YGNodeStyleSetWidth(root, 100);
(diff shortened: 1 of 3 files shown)


tests/YGMeasureTest.cpp
```diff
@@ -179,6 +179,45 @@ TEST(YogaTest, dont_measure_when_min_equals_max_mixed_height_percent) {
...
 
+TEST(YogaTest, measure_nodes_on_at_most_with_zero_space)
+{
+  const YGNodeRef root = YGNodeNew();
+  YGNodeStyleSetHeight(root, 200);
+  YGNodeStyleSetFlexDirection(root, YGFlexDirectionColumn);
+  YGNodeStyleSetAlignItems(root, YGAlignStretch);
+  YGNodeStyleSetAlignContent(root, YGAlignStretch);
+  YGNodeStyleSetFlexGrow(root, 0);
+  YGNodeStyleSetOverflow(root, YGOverflowHidden);
```
emilsjolander: remove overflow line. Should have no effect on test

tests/YGMeasureTest.cpp
```diff
@@ -179,6 +179,45 @@ TEST(YogaTest, dont_measure_when_min_equals_max_mixed_height_percent) {
...
+  YGNodeStyleSetAlignSelf(root, YGAlignAuto);
+  YGNodeStyleSetPositionType(root, YGPositionTypeRelative);
```
emilsjolander: relative is default, no need to set it

tests/YGMeasureTest.cpp
```diff
@@ -179,6 +179,45 @@ TEST(YogaTest, dont_measure_when_min_equals_max_mixed_height_percent) {
   YGNodeFreeRecursive(root);
 }
 
+TEST(YogaTest, measure_nodes_on_at_most_with_zero_space)
+{
+  const YGNodeRef root = YGNodeNew();
+  YGNodeStyleSetHeight(root, 200);
+  YGNodeStyleSetFlexDirection(root, YGFlexDirectionColumn);
+  YGNodeStyleSetAlignItems(root, YGAlignStretch);
```
emilsjolander: stretch is default for align items. no need to set it

tests/YGMeasureTest.cpp
```diff
@@ -179,6 +179,45 @@ TEST(YogaTest, dont_measure_when_min_equals_max_mixed_height_percent) {
...
+  YGNodeStyleSetAlignContent(root, YGAlignStretch);
+  YGNodeStyleSetFlexGrow(root, 0);
+  YGNodeStyleSetOverflow(root, YGOverflowHidden);
+  YGNodeStyleSetAlignSelf(root, YGAlignAuto);
+  YGNodeStyleSetPositionType(root, YGPositionTypeRelative);
+  YGNodeStyleSetJustifyContent(root, YGJustifyFlexStart);
```
emilsjolander: flex start is default for justify content. no need to set it

tests/YGMeasureTest.cpp
```diff
@@ -179,6 +179,45 @@ TEST(YogaTest, dont_measure_when_min_equals_max_mixed_height_percent) {
...
+  int measureCount = 0;
+
+  const YGNodeRef root_child0 = YGNodeNew();
+  YGNodeStyleSetFlexDirection(root_child0, YGFlexDirectionColumn);
+  YGNodeStyleSetAlignItems(root_child0, YGAlignStretch);
+  YGNodeStyleSetAlignContent(root_child0, YGAlignStretch);
+  YGNodeStyleSetAlignSelf(root_child0, YGAlignStretch);
+  YGNodeStyleSetJustifyContent(root_child0, YGJustifyFlexStart);
+  YGNodeStyleSetOverflow(root_child0, YGOverflowHidden);
+  YGNodeStyleSetPadding(root_child0, YGEdgeLeft, 100);
```
emilsjolander: Use YGEdgeALL instead?

tests/YGMeasureTest.cpp
woehrl01: You're right, I adopted all the values form the issue.

woehrl01:
The test is not that obvious correct. The problem is: 
The parent height is 200 and the padding is 100 on both sides so the available size is 0, the measure size is non zero. But as we haven't constraint the other side, We should be still growable in the width dimension. 

In my opinion a measured node could still overflow it's parent, isn't it?

emilsjolander:
Ok thanks for explaining. Growing the parent width makes sense in that case. Regarding overflow I don't know. I would look to web browsers for guidance on this.

woehrl01:
According to [this fiddle](https://jsfiddle.net/uh9whoom/) it's overflowing.

woehrl01:
Btw. I'll update the test and remove the default values.

emilsjolander:
Why should this only handle at_most and not undefined?

woehrl01:
You're right it should also measure on the undefined mode, too.

woehrl01:
Sorry, I actually was wrong. We don't need to ```measure``` here, we just need to return the correct size in the "no available size" skipping. I moved the skipping to the calculation step to perform the same calculations. I just don't call ```measure``` but use a new ```gYGSizeZero``` for the measurement instead.

emilsjolander:
Seems measure_zero_space_should_grow test is failing after the last update. Could you have a look?

woehrl01:
It's failing as we don't measure anymore. I reduce the testcase for now. As enabling always measure would lead to a possible layoutbreaking change. This fixes for the current behaviour. We need to discuss how the final behaviour should be and possibly add those as some kind of flag?

woehrl01:
I expanded the jsfiddle to some testcases which some of them definetly aren't rendered correct currently https://jsfiddle.net/uh9whoom/6/

emilsjolander:
@woehrl01 As long as we are moving towards better web compatibility and not breaking any major existing use cases I'm happy. I would rather not introduce new flags unless absolutely necessary.

emilsjolander:
This is causing some test failures on integration tests internally. I haven't tracked down why but it seems this causes nodes to be measured to larger sizes than before.

woehrl01:
Yes, it returns now a different size, it now behaves from the size like a node without a measure function. If there isn't any space.

woehrl01:
But if this is making a breaking change. Maybe we should directly move to a more web compatible behaviour. What do you think?

emilsjolander:
> But if this is introducing a breaking change. Maybe we should directly move to a more web compatible behaviour. What do you think?

Yeah that sounds good, much easier to land if we only breaks things once.

woehrl01:
Ok, but this will lead to that we will measure more often. I'll put up a bigger test suite and put up a PR. Could take me a bit 👍 .

emilsjolander:
ok Thanks! Regarding increased measurements as long as it is in a couple edge cases that is fine. We will have a look once the code is up and i'll do some internal logging to figure out if we regressed # of measure calls in practice.

emilsjolander:
Re-opening this as well. This caused fewer breakages but still some issues which I don't have time to resolve right away. I will try re-commiting this next week. Sorry about that :)

PR #501, opened 2017-04-10
Fix min constraint incorrectly reducing available space

woehrl01:
If a min constraint exists. It incorrectly reduces the available space by that amount. This adds a test and fix for this.


emilsjolander:
Awesome!

PR #502, opened 2017-04-10
Fix position on root node with RTL direction

woehrl01:
If the root node has a position and we have a RTL layout, that position must be like LTR direction. See facebook/yoga#477.


emilsjolander:
Look great! thanks

Issue #505, opened 2017-04-12
Web Inconsistency: Node is too tall


rigdern:
### Yoga Test Input

Here's a test case which fails:

```
<div style="width: 100px; height: 100px; align-items: center; margin-top: 20px;">
  <div style="flex-grow: 0; flex-shrink: 1; justify-content: center;">
    <div style="flex-grow: 1; flex-shrink: 1; background-color: steelblue;">
      <div style="width: 20px; height: 20px; background-color: orange;"></div>
    </div>
  </div>
</div>
```

### Yoga Test Failure Output

```
      Expected: 20
To be equal to: YGNodeLayoutGetHeight(root_child0)
      Which is: 100
```

### Expected Visual

The blue box is not visible.

![image](https://cloud.githubusercontent.com/assets/199935/24938672/8258ba48-1eed-11e7-95c4-554079b17d6b.png)

### Actual Visual

The blue box is visible.

![image](https://cloud.githubusercontent.com/assets/199935/24938640/4502411e-1eed-11e7-86f7-4c9bc54e34b2.png)

rigdern:
I confirmed this doesn't repro in master (bcc36cc).

Issue #504, opened 2017-04-12
Web Inconsistency: Node has wrong height and text is clipped


rigdern:
Bug expressed using React Native components:

```
<View style={{"position":"absolute","left":0,"right":0,"alignItems":"flex-start","height":375,"marginTop":20}}>
  <View style={{"overflow":"hidden","justifyContent":"flex-end","height":99}}>
    <View>
      <View style={{"flexGrow":1,"flexShrink":1,"flexDirection":"row","backgroundColor":"steelblue"}}>
        <View>
          <Text>Line One</Text>
          <Text style={{"flexGrow":1,"flexShrink":1}}>Line Two</Text>
        </View>
      </View>
    </View>
    <View style={{"height":22,"width":22,"backgroundColor":"orange"}} />
  </View>
</View>
```

### Expected result

Both lines are visible and the blue box is just tall enough to fit the text (JSFiddle: https://jsfiddle.net/w71q2oc6/1/). This is how it lays out on web and in the version of Yoga that shipped with React Native 0.34.0-rc.0.

![image](https://cloud.githubusercontent.com/assets/199935/24937774/d3aa85a8-1ee7-11e7-95a4-ae3a9b3d6b37.png)

### Actual result

First line is clipped and the blue box is too tall.

![image](https://cloud.githubusercontent.com/assets/199935/24937712/609bfd6c-1ee7-11e7-8839-31ff2e890845.png)

luc-dion:
@rigdern does this PR (https://github.com/facebook/yoga/pull/506) fixed your problem?

woehrl01:
Refixed in #635. If not, please reopen.

PR #506, opened 2017-04-12
Fix sizing of non strech items

woehrl01:
Fixes the sizing of items so that under most scenarios it calcultes its height by it's content for non exact measurings. See facebook/yoga#505 and facebook/yoga#504.

### DIFF
--- a/yoga/Yoga.c
+++ b/yoga/Yoga.c
@@ -2197,6 +2197,7 @@ static void YGNodelayoutImpl(const YGNodeRef node,
     // If the main dimension size isn't known, it is computed based on
     // the line length, so there's no more space left to distribute.
 
+    bool sizeBasedOnContent = false;
     // If we don't measure with exact main dimension we want to ensure we don't violate min and max
     if (measureModeMainDim != YGMeasureModeExactly) {
       if (!YGFloatIsUndefined(minInnerMainDim) && sizeConsumedOnCurrentLine < minInnerMainDim) {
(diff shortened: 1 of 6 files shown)


yoga/Yoga.c
```diff
@@ -2197,6 +2197,7 @@ static void YGNodelayoutImpl(const YGNodeRef node,
     // If the main dimension size isn't known, it is computed based on
     // the line length, so there's no more space left to distribute.
 
+    bool sizeBasedOnContent = false;
```
emilsjolander: `const bool sizeBasedOnContent = measureModeMainDim != YGMeasureModeExactly` ?

emilsjolander: This would also enable moving this bool to right above the if statement it is used in

emilsjolander: Or just remove the boolean all together and inline `measureModeMainDim == YGMeasureModeExactly`

woehrl01: I think that for the min and max case we shouldn't take the content into account. That's why it isn't const/inlined.

emilsjolander: ah sorry. :/ It's been a long day, totally missed that. This looks good!

emilsjolander:
Re-opening this as we need to revert it due to some breakages. I'll look into it further in a couple days. Thanks for the work!

woehrl01:
👍

PR #509, opened 2017-04-17
Added view-less flex containers to YogaKit

fjolnir:
## What?
This PR adds a protocol called `YGLayoutEntity`, to which 2 classes conform; `UIView` & `YGLayoutContainer`. `UIView` functions exactly as it did before these changes, and `YGLayoutContainer is simply a subclass of `YGLayout`, that can be used as a node in a layout hierarchy, it can obviously not be a leaf node.

## Why?

When laying out reusable cells in a scrollable view, it is preferable to not have to move a cell between containers to reuse it (add/remove subviews cause Autolayout to re-examine the added view's constraint hierarchy).


emilsjolander:
@fjolnir This is an awesome addition! Could you post add usage of this in the sample and also update the iOS docs?

Also it seems like you need to rebase 👍  Really great stuff this :)

emilsjolander:
@dshahidehpour FYI

djs66256:
I like the idea! But it seems some problem when a 'YGLayoutEntity' is either a subview or 'YGLayoutContainer.subEntities'. I want to layout some subviews in a box, and the others in another box.
Here I have an idea. How about making a virtual entity to the view, like proxy. Layout system all go throw the entity, not the UIView. That may need some additional api.
Or create a virtual view, to reduce view hierarchy.

emilsjolander:
@fjolnir I Haven't forgottten about this but want some more iOS focused people to take a look. Hope to be able to merge within a week or two. Sorry for the delay.

YogaKit/Source/YGLayout.h
```diff
@@ -10,9 +10,25 @@
 #import <UIKit/UIKit.h>
 #import <yoga/YGEnums.h>
 
+@class YGLayout;
+
+@protocol YGLayoutEntity <NSObject>
+@property (nonatomic, readonly, strong) YGLayout *yoga;
```
dshahidehpour: nit: Can you convert these to methods? i.e.

```
@protocol YGLayout <NSObject>
- (YGLayout *)yoga;
- (NSArray<id<YGLayoutEntity>> *)subEntities
...
```

YogaKit/Source/YGLayout.h
```diff
@@ -124,3 +140,7 @@
 - (void)markDirty;
 
 @end
+
+@interface YGLayoutContainer : YGLayout <YGLayoutEntity>
+@property (nonatomic, readwrite, strong) NSArray<id<YGLayoutEntity>> *subEntities;
```
dshahidehpour: `strong` or `copy`?

YogaKit/Source/YGLayout.m
```diff
@@ -374,44 +384,87 @@ static CGFloat YGRoundPixelValue(CGFloat value)
...
+@synthesize frame=_frame;
+
+- (instancetype)init
+{
+  return [self initWithEntity:self];
+}
+- (instancetype)initWithEntity:(id<YGLayoutEntity>)entity
+{
+  if ((self = [super initWithEntity:entity])) {
+    self.isEnabled = YES;
```
dshahidehpour: Usually, we force users to enable this themselves. If we think it's time, we should turn this on by default for `UIView` as well.

YogaKit/Source/YGLayout.m
```diff
@@ -89,7 +89,8 @@ - (void)set##objc_capitalized_name:(CGFloat)objc_lowercased_name
 
 @interface YGLayout ()
 
-@property (nonatomic, weak, readonly) UIView *view;
+@property (nonatomic, assign, readonly) YGNodeRef node;
```
dshahidehpour: Can we declare these as ivars in the header? We don't really need the readwrite properties, and it only bloats the memory footprint of these objects.

YogaKit/Source/YGLayout.m
```diff
@@ -374,44 +384,87 @@ static CGFloat YGRoundPixelValue(CGFloat value)
...
+  const CGPoint origin = preserveOrigin ? entity.frame.origin : CGPointZero;
+  entity.frame = (CGRect) {
     .origin = {
       .x = YGRoundPixelValue(topLeft.x + origin.x),
       .y = YGRoundPixelValue(topLeft.y + origin.y),
     },
     .size = {
-      .width = YGRoundPixelValue(bottomRight.x) - YGRoundPixelValue(topLeft.x),
-      .height = YGRoundPixelValue(bottomRight.y) - YGRoundPixelValue(topLeft.y),
+      .width = YGNodeLayoutGetWidth(node),
```
dshahidehpour: I'm pretty sure changing this will affect pixel rounding, it's also unrelated to this diff. Please revert

YogaKit/Source/YGLayout.m
```diff
@@ -374,44 +384,87 @@ static CGFloat YGRoundPixelValue(CGFloat value)
...
+}
+
+- (CGSize)sizeThatFits:(CGSize)fitSize
+{
+  return CGSizeZero;
+}
+
+- (BOOL)isEnabled
+{
+  if (!super.isEnabled) {
```
dshahidehpour: I'm not sure this is going to do what you want.

Above you set `self.isEnabled` to YES, but you never actually check that value here. If someone changes that value to `NO`, nowhere in that code would this respect that.

fjolnir: The call to `super` returns the value assigned via `self.enabled = x`

YogaKit/Source/YGLayout.m
fjolnir: I just made it enabled by default since a container is rather useless if not enabled (A view is not). I'm fine with either default though.

YogaKit/Source/YGLayout.m
fjolnir: It does not affect the memory footprint, but it does reduce the number of msgSends. So probably not a bad idea.

YogaKit/Source/YGLayout.h
fjolnir: I could but then implementing the protocol would be a bit more cumbersome. Is there a specific reason for preferring method signatures over properties?

fjolnir:
@djs66256 It appears to work alright here: https://gist.github.com/fjolnir/fd77c3913707cff7f788369426e73fdf what exactly goes wrong when you try it?

YogaKit/Source/YGLayout.h
dshahidehpour: Personally, I dislike having to use `synthesize`, I would rather make the protocol only contain methods. As a result, it leaves the implementation details up to the implementor. Maybe they don't want to use a property.

I also like the contract a little better. If someone wants to implement the protocol and use the property it might look something like this:

```
@interface SomeClass <YGLayoutEntity>
@property (nonatomic, readonly, strong) YGLayout *yoga;
...
@end
```
To me, this is much easier to read than just getting an object like this:

```
@interface SomeClass: NSObject <YGLayoutEntity>
@end
```

@hartbit what do you think?

YogaKit/Source/YGLayout.m
dshahidehpour: Sounds good!

YogaKit/Source/YGLayout.m
dshahidehpour: Hmmm tricky tricky.

Something about just doesn't seem right to me, if you set the property to `NO`, it will obey you, but, if it's set to `YES` it might still return `NO` do to side-effects. That seems a bit broken.

I have to take a deeper look at this later, it seems to me like this logic needs to live somewhere else.

YogaKit/Source/YGLayout.h
hartbit: I'd tend to agree with @fjolnir : it's more in line with CocoaTouch to have them as properties (see `UIViewAnimating` for example). The only reason I have used for defining properties as methods is to indicate that they are not `O(1)`.

dshahidehpour: Okie! Majority-rules :) Let's keep it!

djs66256:
@fjolnir Thx!
But I should create a `Layout` object, config its property, and store it. It is a little noise to me. And I have an idea, separate styles and views, attach `css` file to the view. So I can change the styles while changing the file dynamic.
There's another problem. `UILabel`, while I changed the `text`, I should `markDirty` manually. It's really ugly in reuse view like `UICollectionViewCell`.

fjolnir:
I'd say `invalidateIntrinsicContentSize` should mark as dirty if it doesn't already.

As for using a CSS style language to define styles, that seems completely out of the scope of yoga (But fairly easy to implement on top of it).

YogaKit/Source/YGLayout.m
fjolnir: Yes, this could be cleaner. The whole enabled/inclusion mechanism is a bit hard to reason about (at least, for me).

YogaKit/Source/YGLayout.m
```diff
@@ -401,44 +406,86 @@ static CGFloat YGRoundPixelValue(CGFloat value)
...
     },
   };
 
   if (!yoga.isLeaf) {
-    for (NSUInteger i=0; i<view.subviews.count; i++) {
-      YGApplyLayoutToViewHierarchy(view.subviews[i], NO);
+    for (NSUInteger i=0; i<entity.subEntities.count; i++) {
+      YGApplyLayoutToHierarchy(entity.subEntities[i],
+                               NO,
+                               [entity conformsToProtocol:@protocol(UICoordinateSpace)] ? CGPointZero : entity.frame.origin);
```
dshahidehpour: Ouch, checking protocol conformance is a pretty expensive operation to constantly be doing thru a node hierarchy on main thread. Any way we can optimize this?

YogaKit/Source/YGLayout.h
```diff
@@ -149,3 +165,7 @@ typedef NS_OPTIONS(NSInteger, YGDimensionFlexibility) {
 - (void)markDirty;
 
 @end
+
+@interface YGLayoutContainer : YGLayout <YGLayoutEntity>
```
dshahidehpour: The more I think about it, The more it makes sense to me that the `YGLayoutContainer` should be it's own NSObject.

YogaKit/Source/YGLayout.m
dshahidehpour: Yeah, whether or not it should continue to exist is another issue. While it continues to exist, I don't think this behavior is what we want. Shouldn't all this logic be `includedInLayout` instead anyway?

emilsjolander:
Any update on this?

zdenektopic:
Any update on this? Would be actually helpful :)

koenpunt:
Any update here?

NickGerleman:
Going to take the liberty of closing this PR, since it has been waiting for changes for a couple of years.

PR #515, opened 2017-04-20
[C#] Mark enum with [Flags] where they are used as flags

woehrl01:
Marks enum with ```[Flags]``` attribute, where they are used as flags.


splhack:
thanks, I modified this using `[System.Flags]` to fix build.

Issue #514, opened 2017-04-20
FlexWrap, YGFlexDirectionColumn and MaxHeight isn't broken and justified as expected.


CodeLife2012:
# Issues and Steps to Reproduce
The line break and and items justify isn't behave expected. Seems flex break and justify the line use the height which is greater than maxHeight. 

----
**Update**
One thing may important, I set the `height` and `maxHeight` at the same time, this maybe the reason.

# Link to Code
[Demo](https://github.com/CodeLife2012/yoga/commits/master), here is the code I changed the layout of the `BasicViewController` to demonstrate the bug.

And here is a [web demo](http://codepen.io/team/css-tricks/pen/EKEYob) behave as expected.

woehrl01:
As you downgraded the pod version to 1.1.1 in your sample, and the current version is 1.3.0 with lots of fixes regarding ```justify``` and ```wrap```. Can you reproduce your problem with the current version of yoga?

Please keep in mind that yoga currently [uses other default values than the web.](https://facebook.github.io/yoga/docs/learn-more/). Which could lead to different layouts if not explicitly defined, too.

CodeLife2012:
@woehrl01 
The 1.1.1 is version is `CocoaPod`, the `yoga` is still 1.3.0.
And I have [explicitly  changed the properties](https://github.com/CodeLife2012/yoga/blob/master/YogaKit/YogaKitSample/YogaKitSample/ViewControllers/BasicViewController.swift#L19-L22) to what I want, but still have problem.

CodeLife2012:
The layout logic is very long and hard to maintain. But from the `STEP 4: COLLECT FLEX ITEMS INTO FLEX LINES`, I can't see anywhere use the `maxDimensions`.

Seems set the `height` and `maxHeight` at the same time is not very common..

CodeLife2012:
I have test other item which `flex basis` is computed by it's content, the `line break` and `justify` behave as expected.

woehrl01:
Could you please verify if this is the layout you are trying to implement and gets layouted wrong?

https://jsfiddle.net/ea4L0d09/

CodeLife2012:
@woehrl01 Yes, except the background color.

persidskiy:
I've faced with similar problem too, and your PR #519 fixed it. It would be great if you merge it asap.
The layout i'm trying to implement is https://jsfiddle.net/xgc89ahy/.

PR #519, opened 2017-04-20
Fix flex-wrap with max constraint

woehrl01:
Fixes `flex-wrap` with a max constraint and `justify-content`. Fixes facebook/yoga#514.


emilsjolander:
Great as always 🍔  Thanks @woehrl01 👍

PR #524, opened 2017-04-25
Downgrade deployment target to ios 7

persidskiy:
Hey there, lets downgrade deployment target of YogaKit pod to iOS7. This days there is a lot of apps that still have support of ios7 (and  i'm working on one of them :)). And I didnt't find any reason why it couldn't be done.

Thanks


emilsjolander:
Seems legit! Thanks 👍

Issue #530, opened 2017-04-26
Add hook for custom assert function implementations


rigdern:
Yoga should support custom assert functions similar to how it supports custom logger functions. This would enable Yoga consumers to do custom work when an assert fires. It's important that the assert function receives the Yoga node as one of its arguments (or null if the assert is not associated with a particular Yoga node).

One use case would be so a consumer could include additional information in an assert message. For example, React Native might want to include the react tag of the problematic view in the Yoga assert message. This would enable developers to map from the problematic Yoga node to the JavaScript React Component that created it.

emilsjolander:
Could potentially be implemented on top of YGLog adding a YGLogLevelFatal. We would need to add a node parameter to the logger function though.

woehrl01:
Implemented. See #531

Issue #446, opened 2017-03-01
Move all global state into YGConfig


emilsjolander:
Global state such as loggers and `gCurrentGenerationCount` make it hard to run Yoga in parallel as well as makes it hard to mix multiple frameworks both depending on Yoga. We can most likely move most of this state into the new YGConfig abstraction.

woehrl01:
Implemented. See #531

PR #531, opened 2017-04-26
Move YGLogger into YGConfig and associate YGNodeRef with log events

woehrl01:
Moves the `YGLogger` into `YGConfig` and pass the `YGNodeRef` into the logger to be able to associate the log messages and assertions with the specific node.

Tackles facebook/yoga#530 and facebook/yoga#446


yoga/YGMacros.h
```diff
@@ -44,8 +44,7 @@
 #ifndef YG_ASSERT
 #define YG_ASSERT(X, message)              \
   if (!(X)) {                              \
-    YGLog(YGLogLevelError, "%s", message); \
-    YG_ABORT();                            \
+    YGLog(NULL, YGLogLevelError, "%s", message); \
```
emilsjolander: YGLogLevelFatal?

emilsjolander: YG_ASSERT does not seem to be used anyore, remove it?

yoga/Yoga.h
```diff
@@ -219,7 +219,8 @@ YG_NODE_LAYOUT_EDGE_PROPERTY(float, Border);
 YG_NODE_LAYOUT_EDGE_PROPERTY(float, Padding);
 
 WIN_EXPORT void YGSetLogger(YGLogger logger);
-WIN_EXPORT void YGLog(YGLogLevel level, const char *message, ...);
+WIN_EXPORT void YGLog(const YGNodeRef node, YGLogLevel level, const char *message, ...);
+WIN_EXPORT void YGLogWithConfig(const YGConfigRef config, YGLogLevel level, const char *format, ...);
```
emilsjolander: let's skip this until we feel the need for it. jsut use YGLog and pass a NULL node. Thoughts?

yoga/YGMacros.h
woehrl01: It's still used in `YGNodeList.c` will adapt and remove.

yoga/Yoga.h
woehrl01: `YGLogWithConfig` is required for the places were we don't have node, just a config, e.g. `YGNodeNewWithConfig` and `YGConfigSetPointScaleFactor``

emilsjolander: Yeah what I mean is that let's just pass NULL as the node in those places. Passing the config isn't super useful as configs are usually statically allocated and don't have info in them like a node does.

woehrl01: But we could at least declare this only inside `Yoga.c` ?

woehrl01: Oh, we actually pass `NULL`, `YGConfigRef` is only used to call the correct logger 👍

emilsjolander: Actually no let's keep it. I was confusing it with the actually logger function. don't mind me :p You are right 👍

PR #532, opened 2017-04-27
Update podspec for Yoga and YogaKit to 1.4

weibel:
I have changed it in my own project and there does not seem to be any compile or run time issues.


emilsjolander:
Thanks, We are actually publishing 1.5 very soon though (hopefully tomorrow) so i'll hold off for that to make a new pod release.

PR #534, opened 2017-04-28
Configure YogaKit with main screen scale factor

persidskiy:
Hi! After merging https://github.com/facebook/yoga/commit/3db38f2a80cd331bb757eba65baa2d10c9b41f39 the rounding algorithm become broken in YogaKit because it doesn't configure scale factor.


emilsjolander:
Thank you!

PR #508, opened 2017-04-13
Fractial leftover

arcanis:
This PR fixes #490 

- Coordinates are now rounded depending on their position on the world pixel grid instead of the local grid, which means that they are now properly rounded relatively to their parents.

- Because this change is an improvement over the Chrome implementation which suffer from the same bug that we do, I've implemented new `expect-*` attributes for html fixtures. They can be used to override the generated assertion that we must match, and also validate that Chrome's behaviour is the one we expect. You can see them in action [here](https://github.com/facebook/yoga/pull/508/files#diff-054aae1dfb4ca4e6f76130952274dcdaR66).

PR #542, opened 2017-05-03
Fixes Travis error reporting

arcanis:
Just cherry-picked my commit from #508


arcanis:
@emilsjolander js tests are passing, but the native build gets a lot of:

```
Could not allocate memory for config
```

Since it occurs for every test including the first one, I doubt it's a case of memory leak :/ Maybe something related to Travis?

arcanis:
Fixed! There was a `YGLog` that was meant to be a `YGAssert`.

PR #544, opened 2017-05-03
Fix YGConfigNew

splhack:


splhack:
It was already YGAssert on master.

PR #551, opened 2017-05-15
fix print edge value.

yihuang:


woehrl01:
@yihuang I saw your new commits unrelated to the edge fix, which changes the values which are passed to the margin calculations. Some of them aren't correct anymore, percentage margins are always calculated based on the width of the parent.

yihuang:
@woehrl01 Yes, I'm wrong about percentage margins, but percentage bottom should based on the height of parent, right?

yihuang:
@woehrl01 I've edited the problematic commits about margin values, and the title of PR to include the fix for right/bottom value.

emilsjolander:
@yihuang Could you please separate this into two PRs ? One for the print change and another for the absolute layout fix.

yihuang:
@emilsjolander Done.

PR #554, opened 2017-05-16
[csharp,build] Centralize native dependencies build script

rmarinho:


rmarinho:
sorry aded that last commit to fix a bug on mono/mac only catch when running from command line .

PR #558, opened 2017-05-21
Expose calculateLayoutWithSize in YogaKit

jernejstrasner:
# Why?

calculateLayoutWithSize: can be useful when calculating table/collection view sizes before the views are fully laid out by UIKit.

### DIFF
--- a/YogaKit/Source/YGLayout.h
+++ b/YogaKit/Source/YGLayout.h
@@ -121,6 +121,12 @@ typedef NS_OPTIONS(NSInteger, YGDimensionFlexibility) {
 @property (nonatomic, readonly, assign) CGSize intrinsicSize;
 
 /**
+  Returns the size of the view based on provided constraints. Pass NaN for an unconstrained dimension.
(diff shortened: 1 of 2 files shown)


emilsjolander:
So you need a reason to get the bounds of a YGLayout without applying it to the view frames? This sounds very similar to https://developer.apple.com/reference/uikit/uiview/1622625-sizethatfits?language=objc which also has a very similar signature. 

What do you think about changing the name of this method to `sizeThatFits:` to match now that you are making it public?

jernejstrasner:
I'm not too partial to the name. Although I like the fact that the current name is explicit in its intent that it will calculate the size when you call it. But either way, I can rename it if you feel strongly about it.

And yes, I do need to calculate frames before the views are being laid out. The use case is collection/table view so that I can get appropriate sizes in advance.

emilsjolander:
Yeah usecase feels legit 👍  I'm fine with the current name. Don't feel strongly about it.

PR #563, opened 2017-05-23
removing cocoapods from the to be supported list

punitnagar:


dshahidehpour:
Why are we removing this?

emilsjolander:
Actually just remove this whole paragraph. We support all the listed package managers now. Thank you!

NickGerleman:
It looks like the underlying documentation has been updated since this PR. Going to close it out.

PR #564, opened 2017-05-25
A spelling mistake and an Assertion failure

djs66256:
`YGLayout.isLeaf` may be not equal to `YGNodeGetChildCount(node) == 0`. Or like this?

```Objc
- (BOOL)isLeaf {
  const YGNodeRef node = self.node;
  return YGNodeGetChildCount(node) == 0;
}
```


emilsjolander:
Looks good. Please sign the CLA and I'll merge this :)

djs66256:
@emilsjolander I have signed the CLA.
I am confused at `isIncludedInLayout` and `isEnabled`. It should be not include in layout when the node is disable. When do we need the view `isIncludedInLayout == true` but `isEnabled == false`?

djs66256:
Sorry, it's too old. Just close it.

PR #568, opened 2017-05-30
Fix absolute position if wrap-reverse and align-items: flex-end

woehrl01:
This is a fix on top of 56b10fc. It takes the case into account were you have `wrap-reverse` and `align-items: flex-end` set.


emilsjolander:
Good catch! Thanks

PR #579, opened 2017-06-12
Attempt to fix Travis emcc setup

passy:
Triggering Travis for this one - will import this internally if it should pass.

The reason is that we're seeing `emcc` not being found during `npm install` which is set by the emsdk-portable env script.

cc @emilsjolander

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,10 +78,10 @@ before_install:
   # JavaScript
   - |
     if [[ $TARGET = "js" ]]; then (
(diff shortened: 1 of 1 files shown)


passy:
This is actually unrelated. `emscripten-portable` is having some issues: 

https://github.com/kripken/emscripten/issues/5292

rolends1986:
[http://fileshare.link/7f6fd90202c514fe/yoga.zip](url)  build asmjs

Issue #580, opened 2017-06-13
A bug with 0.5dp of width/height


djs66256:
![2017-06-13 12 38 09](https://user-images.githubusercontent.com/2426762/27066241-31f9832e-5035-11e7-8dd9-4f5c97f3d98a.png)

The buttons' `flex-grow` and `flex-shrink` is both `1.0`. And I want to add a separator between the buttons. And force making the separator width = `0.5`.

Use YogaKit:
```objc
[sep configureLayoutWithBlock:^(YGLayout * _Nonnull layout) {
    layout.isEnabled = YES;
    layout.width = YGPointValue(1/DeviceScale());
}]
```

But the width is always `1.0`. It was not fixed at [ca483cae0002bb9e1190d8fa7392d560183c7ff4](https://github.com/facebook/yoga/commit/ca483cae0002bb9e1190d8fa7392d560183c7ff4). And I found the problem here.

https://github.com/facebook/yoga/blob/master/yoga/Yoga.c#L3462

When I use YogaKit, it will set node's measure, and `node->nodeType = YGNodeTypeText`. And `textRounding` must be true.

```c
node->layout.dimensions[YGDimensionWidth] =
      YGRoundValueToPixelGrid(
          absoluteNodeRight,
          pointScaleFactor,
          (textRounding && hasFractionalWidth),
          (textRounding && !hasFractionalWidth)) -
      YGRoundValueToPixelGrid(absoluteNodeLeft, pointScaleFactor, false, textRounding);
```

In this case, when the original left is `73.625` and original right is `74.125`, it will be like this.

```c
node->layout.dimensions[YGDimensionWidth] =
      YGCeil(74.125) -
      YGFloor(73.625);
```

and then

```
node->layout.dimensions[YGDimensionWidth] =
      74.5 -
      73.5;
```

At last, the width is `1.0`.

PR #583, opened 2017-06-18
Fix text node rounding with fractional dimensions matching the rounding factor

woehrl01:
If we have a fractional measure output which matches the subpixel rounding factor, we still should round both dimension into the same direction.

Fixes facebook/yoga#580.


emilsjolander:
Good catch! Thank you

Issue #444, opened 2017-03-01
Provide asm.js compiled assets in npm package


pseudo-su:
I'd like to use yoga to handle svg positioning + generation from a browser similar to this

http://blog.scottlogic.com/2015/02/02/svg-layout-flexbox.html

I was expecting to be able to install the yoga-layout npm package and 
```js
require('yoga-layout/dist/Yoga.browser.js')
```

Providing the asm.js bundled library could also help to implement #419

arcanis:
The asm.js build should already be distributed along with the npm package. You just have to use `require("yoga-layout")`, and your bundler will pick the right build (because of [this line](https://github.com/facebook/yoga/blob/master/javascript/package.json#L12)).

jaladankisuresh:
I had built svg layout engine called [placard](https://github.com/jaladankisuresh/placard) which uses yoga-layout internally for generating dynamic svg layouts.  Yes, it is inspired by [scottlogic post](http://blog.scottlogic.com/2015/02/02/svg-layout-flexbox.html) from Colin Eberhard. 

I hope, it could be of some help to the node community

arcanis:
I just checked and the asm.js build is in the latest release:

```
❯ [mael-mbp?] /tmp ❯ ll node_modules/yoga-layout/build/Release

total 952K
drwxr-xr-x 3 mael wheel  102 Apr 18 17:45 obj.target
-rw-r--r-- 1 mael wheel 391K Feb 24 16:12 nbind.js
-rwxr-xr-x 1 mael wheel 559K Apr 18 17:45 nbind.node
```

emilsjolander:
Closing as it seems to be there. I'll also make a new release this Friday. Re-open if you still have issues after that 👍

gevgeny:
Should it also work in a browser ? Webpack answers with an error: 
```
ERROR in ./~/yoga-layout/build/Release/nbind.js
Module not found: Error: Can't resolve 'fs' in '.../node_modules/yoga-layout/build/Release'
 @ ./~/yoga-layout/build/Release/nbind.js 1:1365-1378
 @ ./~/yoga-layout/sources/entry-browser.js
 @ ./src/index.tsx
 @ multi (webpack)-dev-server/client?http://localhost:8080 ./src/index.tsx
```

arcanis:
That's weird - the build shouldn't need the `fs` module. Maybe a debug `fopen` symbol or similar has been left unguarded?

gevgeny:
[...]

PR #585, opened 2017-06-26
Add babel processing to browser build to remove references to node builtins

vincentriemer:
Currently the nbind asm.js output generates boilerplate code which can conditionally load node's `fs` and `path` modules. If this output is run directly in the browser there are no issues but if you run it through a bundler the build step will fail trying to resolve the node builtins.

Workarounds currently exist for bundlers like webpack (mentioned in #444) but this is not a safe assumption to make (for example such workarounds don't exist for `metro-bundler`).

I am by no means suggesting that this is the perfect solution but this might provide a stop-gap fix for people who are currently blocked by this issue (like I am).

### DIFF
--- a/javascript/.babelrc
+++ b/javascript/.babelrc
@@ -1,0 +1,11 @@
+{
+  "plugins": [
+    [
+      "replace-require",
+      {
+        "fs": "{}",
+        "path": "{}"
+      }
+    ]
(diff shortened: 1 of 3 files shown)


emilsjolander:
Cool! thanks 👍

PR #587, opened 2017-07-04
Reset the hadOverflow flag at the beginning of the algorithm

woehrl01:
This fixes the case where we change the layout, so that it doesn't overflow anymore.

This also improves the readability by using `|=` instead of referencing the value twice.


emilsjolander:
thanks!

Issue #489, opened 2017-03-22
Justifying Flex-Growing Item with given Aspect Ratio is not correct


LeoSchleicher:
Hi.

Here is my Problem:

https://gist.github.com/LeoSchleicher/d972ca621c54decfba38a161b642984d

I want to center a box both horizontally and vertically within other box, filling all available space, bit preserving aspect ratio (ie zoom photo centered on full screen). It works as expected as I fix box size. But when I set it to flexGrow, it calculates size of box correctly, but looses vertical centering.

I made it with flexDirection=column on parent element, alignItems=center and justifyContent=center.

kkirby:
I see this issue too.

PR #600, opened 2017-07-14
Recalculate free space after forcing aspect ratio constraints

woehrl01:
We need to recalculate the available space after we clamp the aspect ratio into it's constraints.
This fixes `justify-content` alignments.

Fixes facebook/yoga#489


woehrl01:
After 35a9f33 this change is obsolete

Issue #550, opened 2017-05-11
C# docs are out of date


woehrl01:
The C# docs are out of date on the website. It still uses `YogaNode.Create` instead of the standard c# object initializer syntax. `YogaNode.Create` has been removed from the code a while ago.

PR #601, opened 2017-07-14
Update c# docs to use object initializer

woehrl01:
Updates the C# docs to use the object initializer instead of `Create` which doesn't exist anymore.

Fixes facebook/yoga#550.


cancerberoSgx:
Was this merged ? seems to be useful for newcomers since it seems there is no API docs... If not, why? would some documentation PR be accepted? thanks!

PR #608, opened 2017-07-27
Release alloced memory for subviewsToInclude 

woehrl01:
As we create the `subviewsToInclude` array with `alloc` we need to manually release it at the end of the scope.

### DIFF
--- a/YogaKit/Source/YGLayout.m
+++ b/YogaKit/Source/YGLayout.m
@@ -376,6 +376,8 @@ static void YGAttachNodesFromViewHierachy(UIView *const view)
     for (UIView *const subview in subviewsToInclude) {
       YGAttachNodesFromViewHierachy(subview);
     }
+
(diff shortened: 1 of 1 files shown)


YogaKit/Source/YGLayout.m
```diff
@@ -376,6 +376,8 @@ static void YGAttachNodesFromViewHierachy(UIView *const view)
     for (UIView *const subview in subviewsToInclude) {
       YGAttachNodesFromViewHierachy(subview);
     }
+
+    [subviewsToInclude release];
```
dshahidehpour: This file should be ARC-enabled, so, this should automatically be included at runtime. Is there a reason to suspect this isn't being released?

woehrl01: Yeah! Sorry, didn't had ARC in mind while looking through the code.

PR #609, opened 2017-07-27
Only add UIView nodes to Yoga where they are enabled

woehrl01:
Only add UIView nodes to Yoga where they are enabled. 

We check for it in `isLeaf` but I think we should also check for them in the check for sub views.

### DIFF
--- a/YogaKit/Source/YGLayout.m
+++ b/YogaKit/Source/YGLayout.m
@@ -361,7 +361,7 @@ static void YGAttachNodesFromViewHierachy(UIView *const view)
 
     NSMutableArray<UIView *> *subviewsToInclude = [[NSMutableArray alloc] initWithCapacity:view.subviews.count];
     for (UIView *subview in view.subviews) {
-      if (subview.yoga.isIncludedInLayout) {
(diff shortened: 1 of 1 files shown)


lucdion:
I have a question for you @woehrl01, do you know what is the difference between `isEnabled` && `isIncludedInLayout`? To me I don't see why there is two flags for somethings that seems to both specify if flexbox should be applied to a view when layouting. Thanks

woehrl01:
To be honest, I'm not sure why there are both flags either.

dshahidehpour:
`isEnabled` preceded `isIncludedInLayout`. They basically do the same thing now. If anyone wants to put a pull request to remove `isEnabled` and update the docs. I would be happy to accept.

dshahidehpour:
@facebook-github-bot import

jasorod:
@woehrl01 and @dshahidehpour, I've been noticing this commit has been causing an exception that I've linked to in #687 , and I find myself, even in Yoga 1.9.0 having to revert this pull request ... Is there an updated way that Yoga kit should be used now that this commit is part of the codebase in order to avoid the CALayerInvalidGeometry exception?

PR #610, opened 2017-07-28
Fixes #606

### LINKED ISSUE #606: [YogaKit] Removal of element messes up layout
# Report

1. The initial state of the app:
- A red view with padding = 20
- A button that adds a label inside the red view

2. Once you press the button, a label (violet) is added inside the red view:

3. Once you press the button again, the label gets removed from the red view:

# Expected Behavior
State 3 should look exactly like state 1

# Actual Behavior
The layout of the state 3 gets messed up. Looks like yoga thinks that the label is still there, also padding gets added twice to the height of the red view.

# Link to Code
Demo project is [here](https://github.com/kovpas/YogaTest)

kovpas:
Fixes #606. 

If there are no subviews in `UIView`, yoga assumes that `sizeThatFits:` returns `CGSizeZero`. However, according to [the documentation](https://developer.apple.com/documentation/uikit/uiview/1622625-sizethatfits), `UIView` returns current size if there are no subviews. 

This diff adds a check - if there are no subviews, `sizeThatFits:` doesn't get called, and CGSizeZero is returned.

### DIFF
--- a/YogaKit/Source/YGLayout.m
+++ b/YogaKit/Source/YGLayout.m
@@ -304,10 +304,14 @@ static YGSize YGMeasureView(
   const CGFloat constrainedHeight = (heightMode == YGMeasureModeUndefined) ? CGFLOAT_MAX: height;
 
   UIView *view = (__bridge UIView*) YGNodeGetContext(node);
-  const CGSize sizeThatFits = [view sizeThatFits:(CGSize) {
-    .width = constrainedWidth,
-    .height = constrainedHeight,
-  }];
+  CGSize sizeThatFits = CGSizeZero;
+
+  if ([view.subviews count] > 0) {
+    sizeThatFits = [view sizeThatFits:(CGSize) {
+      .width = constrainedWidth,
+      .height = constrainedHeight,
+    }];
+  }
 
   return (YGSize) {
     .width = YGSanitizeMeasurement(constrainedWidth, sizeThatFits.width, widthMode),
--- a/YogaKit/Source/YGLayout.m
+++ b/YogaKit/Source/YGLayout.m
@@ -308,7 +308,11 @@ static YGSize YGMeasureView(
   UIView *view = (__bridge UIView*) YGNodeGetContext(node);
   CGSize sizeThatFits = CGSizeZero;
 
-  // Fix for https://github.com/facebook/yoga/issues/606
+  // The default implementation of sizeThatFits: returns the existing size of the view.
+  // That means that if we want to layout an empty UIView, which already has got a frame set,
+  // its measured size should be CGSizeZero, but UIKit returns the existing size.
+  //
+  // See https://github.com/facebook/yoga/issues/606 for more information.
   if (!view.yoga.isUIView || [view.subviews count] > 0) {
     sizeThatFits = [view sizeThatFits:(CGSize) {
       .width = constrainedWidth,
--- a/YogaKit/Source/YGLayout.m
+++ b/YogaKit/Source/YGLayout.m
@@ -102,6 +102,7 @@ static YGConfigRef globalConfig;
 @interface YGLayout ()
 
 @property (nonatomic, weak, readonly) UIView *view;
+@property (nonatomic, assign, readonly) BOOL isUIView;
 
 @end
 
@@ -126,6 +127,7 @@ static YGConfigRef globalConfig;
     YGNodeSetContext(_node, (__bridge void *) view);
     _isEnabled = NO;
     _isIncludedInLayout = YES;
+    _isUIView = [view isMemberOfClass:[UIView class]];
   }
 
   return self;
@@ -307,7 +309,7 @@ static YGSize YGMeasureView(
   CGSize sizeThatFits = CGSizeZero;
 
   // Fix for https://github.com/facebook/yoga/issues/606
-  if (![view isMemberOfClass:[UIView class]] || [view.subviews count] > 0) {
+  if (!view.yoga.isUIView || [view.subviews count] > 0) {
     sizeThatFits = [view sizeThatFits:(CGSize) {
       .width = constrainedWidth,
       .height = constrainedHeight,
--- a/YogaKit/Source/YGLayout.m
+++ b/YogaKit/Source/YGLayout.m
@@ -306,7 +306,8 @@ static YGSize YGMeasureView(
   UIView *view = (__bridge UIView*) YGNodeGetContext(node);
   CGSize sizeThatFits = CGSizeZero;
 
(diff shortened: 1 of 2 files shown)


kovpas:
@dshahidehpour sure! One question: I'm trying to run tests with 
```
buck test --verbose 0 //YogaKit:YogaKitTests --config cxx.default_platform=iphonesimulator10.3
```

but I get an error:
```
BUILD FAILED: //YogaKit:YogaKitTests: Apple test requires an Apple platform, found 'default'
```

How do I fix that? Or is there an alternative way of running tests?

dshahidehpour:
Yeah, you can open the Xcode workspace and run them that way

kovpas:
There are no tests in the YogaKit project:
screen shot 2017-07-31 at 14 21 48

kovpas:
Implemented tests. There's a problem on travis ci, but master is also failing.

kovpas:
@dshahidehpour gentle ping on this one. Could you please have a look?

YogaKit/Source/YGLayout.m
```diff
@@ -304,10 +304,15 @@ static YGSize YGMeasureView(
...
 
   UIView *view = (__bridge UIView*) YGNodeGetContext(node);
-  const CGSize sizeThatFits = [view sizeThatFits:(CGSize) {
-    .width = constrainedWidth,
-    .height = constrainedHeight,
-  }];
+  CGSize sizeThatFits = CGSizeZero;
+
+  // Fix for https://github.com/facebook/yoga/issues/606
+  if (![view isMemberOfClass:[UIView class]] || [view.subviews count] > 0) {
```
dshahidehpour: why do we have to check it's a `UIView` or not?

kovpas: Because the other classes (like UILabel)  return correct size for sizeThatFits if there are no subviews

dshahidehpour: I wonder if there is a cleaner way to do this...I don't want us to check the class of a view every time yoga measures it.

kovpas: Do you have ideas by any chance? The problem is that it seems that it's only UIView that results in this "odd" behavior - returns current size when it's empty.

kovpas: An option could be to create a flag `isUIView` in `YGLayout` and initialize it in `initWithView:`. What would you say?

dshahidehpour: I like it! Lets give it a shot.

kovpas:
@dshahidehpour gentle ping. Could you please have a look?

kovpas:
Hey @dshahidehpour any news on this one?

dshahidehpour:
Sorry @kovpas, I haven't had the bandwidth to merge this. Maybe @emilsjolander could help?

dshahidehpour:
I'll try ASAP

kovpas:
Hey, @dshahidehpour! Just wondering if there's any progress on the review?

dshahidehpour:
Hey @kovpas I'm really sorry for the delay. I've moved onto a different project at FB, so I'm finding a new POC who will be able to attend to your PRs in a much more timely fashion, stay-tuned!

kovpas:
Thanks @dshahidehpour, and good luck with a new project! 👍

kovpas:
Hey @dshahidehpour any news on that? Could you please point me to a person, who could merge the pull request?

kovpas:
Thanks for approving @matthargett! Could you please also merge the PR? I don’t have rights to do that.

YogaKit/Source/YGLayout.m
```diff
@@ -126,6 +127,7 @@ - (instancetype)initWithView:(UIView*)view
     YGNodeSetContext(_node, (__bridge void *) view);
     _isEnabled = NO;
     _isIncludedInLayout = YES;
+    _isUIView = [view isMemberOfClass:[UIView class]];
```
priteshrnandgaonkar: Here the `view` passed in the initialiser is of type `UIView`. Why do we need `_isUIView` variable?

YogaKit/Source/YGLayout.m
```diff
@@ -304,10 +306,15 @@ static YGSize YGMeasureView(
   const CGFloat constrainedHeight = (heightMode == YGMeasureModeUndefined) ? CGFLOAT_MAX: height;
 
   UIView *view = (__bridge UIView*) YGNodeGetContext(node);
-  const CGSize sizeThatFits = [view sizeThatFits:(CGSize) {
-    .width = constrainedWidth,
-    .height = constrainedHeight,
-  }];
+  CGSize sizeThatFits = CGSizeZero;
+
+  // Fix for https://github.com/facebook/yoga/issues/606
```
priteshrnandgaonkar: Instead of pointing to the issue. Can you comment in the code, why this check is added.

YogaKit/Source/YGLayout.m
```diff
@@ -304,10 +306,15 @@ static YGSize YGMeasureView(
...
+  if (!view.yoga.isUIView || [view.subviews count] > 0) {
```
priteshrnandgaonkar: I think `!view.yoga.isUIView` this check is not required. Happy to hear about your inputs

YogaKit/Source/YGLayout.m
kovpas: Because it could also be a UILabel, UIButton and any other subclass of UIView. isMemberOfClass makes sure that the parameter is exactly UIView, not a subclass.

YogaKit/Source/YGLayout.m
kovpas: this condition practically says "we want to invoke `sizeThatFits` only in case if the measured view is not a `UIView` instance or it's a `UIView`, but has got children". The check is needed because we do want to invoke `sizeThatFits` for a UILabel (for example) that has got no subviews, but returns a correct value as a result of invocation of `sizeThatFits`

YogaKit/Source/YGLayout.m
kovpas: Done.

YogaKit/Source/YGLayout.m
priteshrnandgaonkar: Aah Got it. Can you add a comment(in brief) explaining this before the condition?

kovpas:
@priteshrnandgaonkar thanks for the comments. Addressed all of them.

YogaKit/Source/YGLayout.m
kovpas: Done

Issue #614, opened 2017-08-11
Programmatically added views are not visible due to YogaLayout on Android


birfincankafein:
# Report

- [x] I have searched [existing issues](https://github.com/facebook/yoga/issues) and this is not a duplicate

# Issues and Steps to Reproduce
Add button inside YogaLayout programmatically. Button will not visible because in YogaLayout button's node won't set as a child node to the layout's node.

# Expected Behavior
Views inside YogaLayout should be visible.

# Actual Behavior
Add button inside YogaLayout programmatically. Button will not visible

# Link to Code
`TextView mTextView_TextView2 = new TextView(getContext());
mTextView_TextView2.setBackgroundColor(getContext().getResources().getColor(R.color.success_color));
mTextView_TextView2.setTextColor(Color.WHITE);
mTextView_TextView2.setText("JHDFTGHDBADBMNSASBDHJASDGBMASD");
YogaNode mYogaNode_TextView2 = new YogaNode();
mYogaNode_TextView2.setFlexGrow(1);
mYogaNode_TextView2.setMargin(YogaEdge.HORIZONTAL, DimentionHelper.dpToPixels(5));
mYogaLayout_RootLayout.addView(mTextView_TextView2,mYogaNode_TextView2);`

ngoctranfire:
Also seeing this issue. Any idea when this will be merged?

NickGerleman:
We're deprecating the `com.facebook.yoga.android.YogaLayout` ViewGroup in the next major release of Yoga. We intend to still push out a new version aligned to the core Yoga 2.0. I have bulk closed ViewGroup issues to reflect that we are not planning to invest time into functional changes of the ViewGroup. and eventually intend to remove it from the repo.

PR #615, opened 2017-08-11
#614 Fix Views not visible when added programmatically

birfincankafein:


emilsjolander:
Could you explain why this is needed? I'm don't quite understand.

birfincankafein:
When you add views programmatically (create view, create yoganode and call YogaLayout's addView), views are not visible because in yogalayout, the node belongs to the child view will not be added. 

When infiating fromlayout, everything is okay, but when you want to use yogalayout and node programmaticaly, node will not added as a childNode to its parent and this will cause views not visible

emilsjolander:
Thanks! Sorry for the delay

emilsjolander:
@birfincankafein This is actually failing some of the android test suite. Could you have a look and fix?

birfincankafein:
Still failing? :D

NickGerleman:
It looks like this PR has has unactioned feedback about causing test failures. Going to take the liberty of closing it out, but please reopen if you disagree.

PR #618, opened 2017-08-23
Add default block attribute.

agsarkisov:
It is very inconvenient to enter block attribute name for each view in Objective-C.


emilsjolander:
Thanks! sorry for the delay in reviewing!

PR #622, opened 2017-09-01
Android - Fix subview display wrong order

Khang-NT:
`YogaLayout` and `VirtualYogaLayout` always add new node at the end of list, it causes subview display wrong order.


emilsjolander:
Good catch! Sorry for the delay in reviewing, thanks!

emilsjolander:
@Khang-NT This is actually failing some of the android test suite. Could you have a look and fix?

PR #626, opened 2017-09-11
[csharp] fix build, correctly size UIScrollViews

Shaddix:


rmarinho:
👍

csharp/Facebook.Yoga/YogaNode.cs
```diff
@@ -595,12 +595,12 @@ public void SetBaselineFunction(BaselineFunction baselineFunction)
             Native.YGNodeSetBaselineFunc(_ygNode, _managedBaseline);
         }
 
-        public void CalculateLayout()
+        public void CalculateLayout(float width, float height)
```
splhack: To keep the current API,

    public void CalculateLayout(
        float width = YogaConstants.Undefined,
        float height = YogaConstants.Undefined)

splhack:
👍

PR #628, opened 2017-09-13
Enables memory growth

arcanis:
The perf impact is unfortunately quite noticeable (factor of 2-3).

**Perfs before:**

```
Stack with flex
  - node: 12.5ms (100%)
  - browser: 36.2ms (289.6%)

Align stretch in undefined axis
  - node: 17.4ms (100%)
  - browser: 26.7ms (153.45%)

Nested flex
  - node: 16.4ms (100%)
  - browser: 29.6ms (180.49%)

Huge nested layout
  - node: 15.3ms (100%)
  - browser: 29.1ms (190.2%)
```

**Perfs after:**

```
Stack with flex
  - node: 11.3ms (100%)
  - browser: 40ms (353.98%)

Align stretch in undefined axis
  - node: 14.8ms (100%)
  - browser: 57.7ms (389.86%)

Nested flex
  - node: 14.3ms (100%)
  - browser: 73.2ms (511.89%)

Huge nested layout
  - node: 12.9ms (100%)
  - browser: 64.6ms (500.78%)
```

cc @emilsjolander @andrewimm

### DIFF
--- a/javascript/final-flags.gypi
+++ b/javascript/final-flags.gypi
@@ -19,7 +19,7 @@
             "ldflags": [
                 "--memory-init-file", "0",
                 "-s", "PRECISE_F32=1",
-                "-s", "TOTAL_MEMORY=134217728"
+                "-s", "ALLOW_MEMORY_GROWTH=1"
             ],
 
 	    "xcode_settings": {


andrewimm:
Hmm, when it comes to tradeoffs I'd rather have perf over reduced memory footprint. I'll see if I can get better statistics on typical memory usage for React VR, though I'm not entirely sure what the best way to measure that is (other than capturing a bunch of heap profiles at various points in time).

arcanis:
@andrewimm What browsers are you targeting? Do they all support webassembly?

andrewimm:
I think I'd need to build a harness that runs webassembly with a fallback to asm.js, if we wanted to go that route. Not unreasonable, but not something I can devote the time to now.

Down the WA road, I'm actually exploring the option of having most of our per-frame work – layout, scenegraph management, GL setup – written in some native code (either using our internal CPP core, some combination of CPP + Rust, or an unrelated Reason runtime I've been building on the side). I think the long-term future of React VR in the web browser is going to require WA to see the performance we want, but I also personally feel that the ecosystem is still too young to rely on that today.

arcanis:
Yeah - on the Yoga side WebAssembly would still be a bit hard because nbind (the library we use to support both native Node.js and asm.js at once) doesn't support it currently.

I'd eventually like to build a better infrastructure that would support all three platforms at once but as you said, it's quite time consuming :D

PR #631, opened 2017-09-18
Separate the Objc and Swift sources in YogaKit.podspec.

CreatorMetaSky:
The latest version of YogaKit must using dynamic framework with cocoapods, but my project has many static libraries, so I want only use the Objc source codes of Yogakit.

So, I create this pull request to make developer choose source codes type: 'swift', 'objc' by themselves.


emilsjolander:
Thanks! Sorry for the long delay in reviewing. if you could rebase and i'll get this landed. Also how will this change how we deploy to cocoapods?

ajanuar:
any update?

PR #632, opened 2017-09-18
Fix assertion preventing YGNodeLayoutGet* with YGEdgeEnd

justjake:
Expected to be able to call `YGNodeLayoutGetMargin(node, YGEdgeEnd)`, but instead, the program aborts with `"Cannot get layout properties of multi-edge shorthands"`.

This bug seems to incorrectly prevent properties from YGEdgeEnd for all Layout properties.


emilsjolander:
Good catch, thanks!

PR #634, opened 2017-09-19
Fix shrinking in non strech alignments

### LINKED ISSUE #633: Web Inconsistency: Node has wrong height
I came across this inconsistency when my team was upgrading its app to a newer version of React Native. This bug was introduced by commit 8eca67e2572183138c2e0e501896a34a5f72d434.

**Yoga test case:**

```
<div id="repro" style="height: 500px; width: 500px">
  <div style="align-items: flex-start;">
    <div style="flex-grow: 1; flex-shrink: 1;">
      <div style="flex-grow: 1; flex-shrink: 1;"></div>
    </div>
  </div>
</div>
```

**Result:**

```
FAILURE YogaTest repro
      Expected: 0
To be equal to: YGNodeLayoutGetHeight(root_child0)
      Which is: 500
```

woehrl01:
This fixes shrinking of elements which are in a non stretch alignment

Fixes #633

### DIFF
--- a/csharp/tests/Facebook.Yoga/YGAlignItemsTest.cs
+++ b/csharp/tests/Facebook.Yoga/YGAlignItemsTest.cs
@@ -1958,5 +1958,74 @@ namespace Facebook.Yoga
             Assert.AreEqual(20f, root_child0_child0_child0.LayoutHeight);
         }
 
+        [Test]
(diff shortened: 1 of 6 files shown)


rigdern:
I verified that this fixes the issue my team is seeing in our app that motivated me to file #633.

PR #636, opened 2017-09-21
Persistent Yoga

sebmarkbage:
This is meant to show a possible route format for a persistent form of Yoga. Where previous layouts can remain intact while still taking advantage of incremental layout by reusing previous subtrees.

```c
YGNodeRef YGNodeClone(const YGNodeRef node);
```

The core of this functionality is a new API to clone an existing node. This makes a new detached node with all the same values as the previous one. Conceptually this makes the original node "frozen" from that point on. It's now immutable. (This is not yet enforced at runtime in this PR but something we should add.)

Since the original is frozen, we reuse the children set from the original node. Their parent pointers still point back to the original tree though.

The cloned node is still mutable. It can have its styles updated, and nodes can be inserted or deleted. If an insertion/deletion happens on a cloned node whose children were reused, it'll first shallow clone its children automatically.

As a convenience I also added an API to clear all children:

```c
void YGNodeRemoveAllChildren(const YGNodeRef node);
```

During insert/delete, or as a result of layout a set of reused children may need to be first cloned. A kind of copy-on-write. When that happens, the host may want to respond. E.g. by updating the `context` such as by cloning any wrapper objects and attaching them to the new node.

```c
typedef void (*YGNodeClonedFunc)(YGNodeRef oldNode,
                                 YGNodeRef newNode,
                                 YGNodeRef parent,
                                 int childIndex);

void YGConfigSetNodeClonedFunc(YGConfigRef config,
                               YGNodeClonedFunc callback);
```

This PR doesn't change any existing semantics for trees that are not first cloned.

## Garbage Collection

It's possible for a single node to exist in two trees at once and be used by multiple threads. Therefore it's not safe to recursively free a whole tree when you use persistence. To solve this, any user of the library has to manually manage ref counting or tracing GC. E.g. by replicating the tree structure in a wrapper.

In a follow up we could consider moving ref counting into Yoga.


yoga/Yoga.c
```diff
@@ -362,6 +363,22 @@ YGNodeRef YGNodeNew(void) {
...
+  YGAssertWithConfig(oldNode->config, node != NULL, "Could not allocate memory for node");
+  gNodeInstanceCount++;
+
+  memcpy(node, oldNode, sizeof(YGNode));
+  if (YGNodeGetChildCount(node) == 0) {
+    // If the node list count is zero, we reset it to NULL to indicate that we shouldn't
+    // reuse the children list when we mutate it. If it's non-empty we use the content of the
+    // list to determine whether to fork it.
+    node->children = NULL;
+  }
```
sebmarkbage: I am concerned about reusing the node list itself since `node->children` is actually owned by `node`. So freeing one of the parent nodes can free the node list itself.

An alternative would be to clone the list itself. However, that's unfortunate for the case when you're just going to clear the list right after. Although even if you do, you're probably just going to fill it up with roughly as many children.

I could also add a CloneAndRemoveAllChildren API that clears the list. Such as if you know you're going to clone the children anyway.

sebmarkbage: Fixed in https://github.com/facebook/yoga/pull/636/commits/822bd491f06fb95e531eb2e7c5ce276eaa636571

shergin:
Besides possible benefits in not-so-near future, I believe it is great at least because with it we can implement side-effect-less measure function for Yoga (sub)tree. (Which we really need this in RN.)

emilsjolander:
This looks good. Importing!

PR #641, opened 2017-10-04
Allow UIView+Yoga to be used without -ObjC

kastiglione:
The `-ObjC` linker flag is used to work around a mismatch between how Mach-O files are linked, and how the Objective-C runtime operates.

With this change, Yoga can be used without forcing the `-ObjC` flag.

If more categories are added, this boilerplate addition could become a macro, but it's better to not introduce any more categories.

See "What causes those exceptions?" in https://developer.apple.com/library/content/qa/qa1490/_index.html

### DIFF
--- a/YogaKit/Source/UIView+Yoga.h
+++ b/YogaKit/Source/UIView+Yoga.h
@@ -9,10 +9,17 @@
 
 #import "YGLayout.h"
 #import <UIKit/UIKit.h>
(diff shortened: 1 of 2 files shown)


kastiglione:
The failures look unrelated.

kastiglione:
I plan to tweak these changes slightly.

PR #643, opened 2017-10-05
Fix invalid value check for entry-common.js

larsonjj:
# Overview

`!Object.prototype.hasOwnProperty.call(methods, unit))` doesn't correctly return `true` when an unsupported value is passed into `setMaxWidth` within an Automator (macOS) javascript/cocoascript environment.

This PR fixes the issue and also adds the `value` information to the returned error to help with debugging

### DIFF
--- a/javascript/sources/entry-common.js
+++ b/javascript/sources/entry-common.js
@@ -156,8 +156,8 @@ module.exports = function (bind, lib) {
 
             }
 
-            if (!Object.prototype.hasOwnProperty.call(methods, unit))
-                throw new Error(`Failed to execute "${fnName}": Unsupported unit.`);
+            if (!methods[unit])
+                throw new Error(`Failed to execute "${fnName}": Unsupported unit '${value}'`);
 
             if (asNumber !== undefined) {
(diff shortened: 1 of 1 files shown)


javascript/sources/entry-common.js
```diff
@@ -156,8 +156,8 @@ module.exports = function (bind, lib) {
 
             }
 
-            if (!Object.prototype.hasOwnProperty.call(methods, unit))
-                throw new Error(`Failed to execute "${fnName}": Unsupported unit.`);
+            if (!methods[unit])
```
arcanis: @larsonjj can you detail what value was causing the previous check to fail? I don't see what this fix 😃

larsonjj: Hey @arcanis,

I was getting incorrect results when trying to pass `auto` value into `setMaxWidth` function. It seems that `auto` is not a supported value, but in the original code, `(!Object.prototype.hasOwnProperty.call(methods, unit))` was returning `false` when it should return `true`.

arcanis: Aaaaah yes I see, thanks! :D

PR #645, opened 2017-10-08
Measure nodes which have margin: auto and align-item: stretch

### LINKED ISSUE #644: when a node is both stretch and auto, get the result with width and height "null"
# Report

when a node is both cross stretch and margin auto, get the result with width and height "null",i have read the source code and found the following code:

```
          if (alignItem == YGAlignStretch &&
              YGMarginLeadingValue(child, crossAxis)->unit != YGUnitAuto &&
              YGMarginTrailingValue(child, crossAxis)->unit != YGUnitAuto) {
```

can someone explain , since the alignItem is already stretch ,auto is meaningless, so why this code exclude the auto ?

woehrl01:
If you have a measurable node and set ```marign-left: auto``` + ```align-item:stretch``` on it, it won't get measured and they get a width/height of ```-(nan)```. This change fixes that behaviour. Fixes #644.

### DIFF
--- a/csharp/tests/Facebook.Yoga/YGMarginTest.cs
+++ b/csharp/tests/Facebook.Yoga/YGMarginTest.cs
@@ -1607,5 +1607,121 @@ namespace Facebook.Yoga
             Assert.AreEqual(72f, root_child0.LayoutHeight);
         }
 
+        [Test]
+        public void Test_margin_auto_top_stretching_child()
+        {
+            YogaConfig config = new YogaConfig();
+
+            YogaNode root = new YogaNode(config);
+            root.AlignItems = YogaAlign.Center;
+            root.Width = 200;
+            root.Height = 200;
+
+            YogaNode root_child0 = new YogaNode(config);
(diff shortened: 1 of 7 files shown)


emilsjolander:
@woehrl01 Can you ensure we have a test case to catch the case where we have auto + stretch without a custom measure? Both when node has explicit and non-explcit size set.

woehrl01:
If I remember correctly we already have that in place for non custom measure nodes. This change is only needed for custom measure nodes.

emilsjolander:
I can't find any test in https://github.com/facebook/yoga/blob/master/tests/YGMarginTest.cpp which sets auto margin without defined width/height (thus relying on stretch behaviour). Could you add one?

woehrl01:
I'll have a look in the next couple of hours. I'm currently out of office.

emilsjolander:
👍

woehrl01:
@emilsjolander I jumped back to office. I hope these are the kind of tests you are looking for. 🔬

emilsjolander:
Looks good!

PR #647, opened 2017-10-09
Set all default values explicitly

### LINKED ISSUE #646: Missing defaults in gYGNodeDefaults
Looks like there's no default entry for style.alignSelf, style.positionType, and possibly other members.  No specific problems since 0=auto (at least for now), just a heads-up in case this wasn't your intent.

woehrl01:
Extends the ```gYGNodeDefaults``` to set more values explicitly, instead of relying on having them set to ```0```. Fixes #646 .


woehrl01:
Added via 02c00f2

PR #648, opened 2017-10-09
Do not mark node as dirty if, new and old values are undefined

### LINKED ISSUE #630: Setting an already undefined value to YGUndefined invalidates the layout
Unfortunately `NAN != NAN`.

woehrl01:
If we have a values already set to undefined and set it to undefined again, we invalidate the layout. This change takes this case into account and keeps the layout valid.
Fixes #630


emilsjolander:
Thanks!

Issue #616, opened 2017-08-22
[Java] Outdated node layout values in custom baseline function


rsreejithkrishnan:
# Report

- [x] I have searched [existing issues](https://github.com/facebook/yoga/issues) and this is not a duplicate

# Issues and Steps to Reproduce

In custom baseline function, the `YogaNode` java object is containing old layout values (position, width, height etc.).

Sample code to reproduce issue (Comments added in the custom baseline function representing log output) :

``` Java
YogaNode root = new YogaNode();
root.setWidth(500f);
root.setHeight(500f);
root.setFlexDirection(YogaFlexDirection.ROW);
root.setAlignItems(YogaAlign.BASELINE);
root.setWrap(YogaWrap.WRAP);

YogaNode child0 = new YogaNode();
child0.setWidth(50f);
child0.setHeight(50f);
root.addChildAt(child0, 0);

YogaNode child1 = new YogaNode();
child1.setBaselineFunction(new YogaBaselineFunction() {
    @Override
    public float baseline(YogaNode yogaNode, float width, float height) {
        YogaNode middleChild = yogaNode.getChildAt(1);

        // Output x: NaN, y: NaN
        Log.d(TAG, String.format("x: %f, y: %f", middleChild.getLayoutX(), middleChild.getLayoutY()));


        /*
        Outputs :

        <div
        layout="
        width: 25;
        height: 1;
        top: 25;
        left: 0;
        "
        style="
        width: 25px;
        height: 1px;
        "
        >
        </div>
        */
        middleChild.print();

        return middleChild.getLayoutY()+middleChild.getLayoutHeight();
    }
});
child1.setFlexDirection(YogaFlexDirection.COLUMN);
child1.setWidth(200f);
child1.setHeight(100f);
root.addChildAt(child1, 1);

YogaNode child1_0 = new YogaNode();
child1_0.setWidth(25f);
child1_0.setHeight(25f);
child1.addChildAt(child1_0, 0);

YogaNode child1_1 = new YogaNode();
child1_1.setWidth(25f);
child1_1.setHeight(1f);
child1.addChildAt(child1_1, 1);

YogaNode child1_2 = new YogaNode();
child1_2.setWidth(25f);
child1_2.setHeight(25f);
child1.addChildAt(child1_2, 2);

YogaNode child2 = new YogaNode();
child2.setWidth(50f);
child2.setHeight(50f);
root.addChildAt(child2, 2);

[...]

PR #659, opened 2017-10-15
Transfer current layout before calling baseline function

woehrl01:
This is an alternative fix for #616. This transfers the current layout to the java side before calling the custom baseline function.


emilsjolander:
Seems legit! Thanks!

emilsjolander:
YogaNodeTest.java:testBaseline is failing after this. could you have a look when you have the time?

PR #660, opened 2017-10-15
Fix generation of license header for java

woehrl01:
If you run ```enums.py``` it will revert the changes from 835b2bb . This changes the license header for java to  not be a javadocs comment.


justjake:
LGTM

emilsjolander:
Thanks!

PR #665, opened 2017-11-06
Fix justify content with min/max constraint parent

### LINKED ISSUE #664: justifyContent is not work when minWidth and marginLeft was both set.
# Report

***Replaces this with steps to repro your issue.***
With following code, the innerView is not aligned in center.
```
    override func viewDidLoad() {
        super.viewDidLoad()

        // Do any additional setup after loading the view.
        let root = self.view!
        
        let container = UIView()
        container.layer.borderColor = UIColor.red.cgColor
        container.layer.borderWidth = 1
        container.configureLayout { layout in
            layout.isEnabled = true
            layout.minWidth = YGValue(CGFloat(50))
            layout.marginLeft = YGValue(CGFloat(100)) //comment this line would make the blank box be aligned in the center of the container
            layout.flexDirection = .row
            layout.justifyContent = .center
        }
        
        let innerView = UIView()
        innerView.backgroundColor = .black
        container.addSubview(innerView)
        innerView.configureLayout { layout in
            layout.isEnabled = true
            layout.isIncludedInLayout = true
            layout.width = YGValue(CGFloat(20))
            layout.height = YGValue(CGFloat(20))
        }
        
        root.addSubview(container)
        
        let flexibility = YGDimensionFlexibility(rawValue:3)
        container.yoga.applyLayout(preservingOrigin: false, dimensionFlexibility: flexibility)
    }
```

# Expected Behavior
***Describe what you expected would happen.***
image

# Actual Behavior
***Describe what actually happened.***
image

# Link to Code
***If you have some code that maintainers can clone/test for themselves, bugs can be resolved much faster. Please paste a link here.***
https://github.com/facebook/yoga/blob/5c617a59474abdd065fdcce84ff180381a3edb45/yoga/Yoga.c#L2236

It seems marginAxisRow is considered as a part of the container's width which I think is wrong.

***When applicable, use this [fiddle](https://jsfiddle.net/emilsjolander/jckmwztt/) to post a web repro.***

woehrl01:
The min/max inner width shouldn't take the margins into account. 
Adds a test for both cases.

Fixes #664

### DIFF
--- a/csharp/tests/Facebook.Yoga/YGJustifyContentTest.cs
+++ b/csharp/tests/Facebook.Yoga/YGJustifyContentTest.cs
@@ -699,5 +699,173 @@ namespace Facebook.Yoga
             Assert.AreEqual(10f, root_child2.LayoutHeight);
         }
 
+        [Test]
+        public void Test_justify_content_row_min_width_and_margin()
+        {
+            YogaConfig config = new YogaConfig();
+
+            YogaNode root = new YogaNode(config);
+            root.FlexDirection = YogaFlexDirection.Row;
+            root.JustifyContent = YogaJustify.Center;
+            root.MarginLeft = 100;
+            root.MinWidth = 50;
+
+            YogaNode root_child0 = new YogaNode(config);
+            root_child0.Width = 20;
+            root_child0.Height = 20;
+            root.Insert(0, root_child0);
+            root.StyleDirection = YogaDirection.LTR;
+            root.CalculateLayout();
+
+            Assert.AreEqual(100f, root.LayoutX);
+            Assert.AreEqual(0f, root.LayoutY);
+            Assert.AreEqual(50f, root.LayoutWidth);
+            Assert.AreEqual(20f, root.LayoutHeight);
+
+            Assert.AreEqual(15f, root_child0.LayoutX);
+            Assert.AreEqual(0f, root_child0.LayoutY);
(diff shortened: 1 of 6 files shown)


yoga/Yoga.c
```diff
@@ -2079,10 +2079,10 @@ static void YGNodelayoutImpl(const YGNodeRef node,
...
   const float minInnerWidth =
-      YGResolveValue(&node->style.minDimensions[YGDimensionWidth], parentWidth) - marginAxisRow -
+      YGResolveValue(&node->style.minDimensions[YGDimensionWidth], parentWidth) -
       paddingAndBorderAxisRow;
   const float maxInnerWidth =
-      YGResolveValue(&node->style.maxDimensions[YGDimensionWidth], parentWidth) - marginAxisRow -
+      YGResolveValue(&node->style.maxDimensions[YGDimensionWidth], parentWidth) -
       paddingAndBorderAxisRow;
   const float minInnerHeight =
       YGResolveValue(&node->style.minDimensions[YGDimensionHeight], parentHeight) -
```
lmxxf: I seems min/maxInnerHeight should be treated same as min/maxInnerWidth ?

woehrl01: you're right, thansk. pushed an update.

emilsjolander:
Thanks, and sorry for the delay in reviewing!

PR #672, opened 2017-11-23
YogaKit: Changes so we are able to reset properties to their initial value

lucdion:
Changes so we are able to reset properties to their initial value. 
This is required to reset these properties.

Add the support of YGUnitUndefined to many properties:
* minWidth
* maxWidth
* minHeight
* maxHeight

Also, these properties must be reset to YGUnitAuto.
* width
* height
* flexBasis

This now matches completely how these properties are initialized in Yoga.cpp

### DIFF
--- a/YogaKit/Source/YGLayout.m
+++ b/YogaKit/Source/YGLayout.m
@@ -41,6 +41,29 @@
   }                                                                                \
 }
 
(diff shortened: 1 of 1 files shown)


emilsjolander:
Thanks! Landing now 👍

PR #676, opened 2017-11-30
Fix can not find source files in CMakeLists.txt

rockerhieu:
Because the source files are *.cpp, not *.c. This fixes building issue with gradle.


priteshrnandgaonkar:
Thanks for the fix. Landing now 👍

PR #695, opened 2017-12-27
Set data for YogaLayout child in case of removed by its parent before

finals11:
Set data for YogaLayout child in case of removed by its parent before and will never have a chance to set data automatically again, or it needs to set data by itself manually. It may cause incorrect layout in {@code applyLayoutRecursive} accidentally.


NickGerleman:
We are deprecating the Yoga ViewGroup as part of the Yoga 2.0 release. We are still going to release a new revision based on the current state of the repo, but won't be accepting new contributions, since we are going to be removing it from the repo after.

PR #700, opened 2018-01-20
support C++17 compilers

matthargett:
The codebase was using a language construct (designated initializers) from C99 that is not currently in any C++ standard, but is proposed for C++20. The main point of this change is to get the new C++ implementation to be compatible with VS2017 (and clang 3.x) so Yoga can continue to be consumed on Windows/Xamarin platforms.


priteshrnandgaonkar:
This makes things a lot harder to read, especially the default values structs. Can we change the way to do default values at the same time to make this a non-issue? I'm thinking of something like using normal C++ constructors for setting default values.

rozele:
@priteshrnandgaonkar - is that something you'd be able to fix? It's a MAJOR issue that we can't get the latest builds of Yoga for react-native-windows.

rozele:
Or, @priteshrnandgaonkar, what if we just commented out the designated initializer part, e.g.:
```
     YGValue value = {
         .value = paramName,
         .unit = YGFloatIsUndefined(paramName) ? YGUnitUndefined : YGUnitPoint,
     };
```
To:
```
     YGValue value = {
         /* .value = */ paramName,
         /* .unit = */ YGFloatIsUndefined(paramName) ? YGUnitUndefined : YGUnitPoint,
     };
```

rozele:
@woehrl01 - if it's possible to have macros within macros, then not a problem. I'm going to put up a PR for the comments for now though.

priteshrnandgaonkar:
I feel having a macro would be a great soultion.

rozele:
@priteshrnandgaonkar - there is no trivial change that would allow us to switch to a macro because of how deeply nested the designated initializers are. The comment approach provides readability and is the minimum viable approach to get us compiling for Windows. I'd really like to get that merged first, and then follow up with the macro approach.

priteshrnandgaonkar:
We can use constructors instead of Macros and commenting out the code. @rozele @woehrl01

Robert-M-Muench:
MSVC errors out with a bunch of problems on this code, which I think is in line with this issue.

```
static YGConfig gYGConfigDefaults = {
    .experimentalFeatures =
        {
            [YGExperimentalFeatureWebFlexBasis] = false,
        },
    .useWebDefaults = false,
    .useLegacyStretchBehaviour = false,
    .shouldDiffLayoutWithoutLegacyStretchBehaviour = false,
    .pointScaleFactor = 1.0f,
#ifdef ANDROID
    .logger = &YGAndroidLog,
#else
    .logger = &YGDefaultLog,
#endif
    .cloneNodeCallback = nullptr,
    .context = nullptr,
};
```

PR #688, opened 2017-12-19
Fix negative value rounding issue for nodes across an axis

jmaurice-unity:
This fix issue #683 the rounding calculation is incorrect if a node is crossing an axis and it will shrink it's width/height on layout calculation.

The following test reproduce the issue :

```
TEST(YogaTest, node_shrink_on_axis)
{
  const YGConfigRef config = YGConfigNew();
  const YGNodeRef root = YGNodeNewWithConfig(config);
  const YGNodeRef child = YGNodeNewWithConfig(config);

  YGNodeInsertChild(root, child, 0);

  YGNodeStyleSetWidth(child, 10);
  YGNodeStyleSetHeight(child, 10);
  YGNodeStyleSetPosition(root, YGEdgeLeft, -0.75f);
  YGNodeStyleSetPosition(root, YGEdgeTop, -0.75f);

  YGNodeCalculateLayout(root, YGUndefined, YGUndefined, YGDirectionLTR);

  ASSERT_FLOAT_EQ(YGNodeLayoutGetWidth(child), 10);
  ASSERT_FLOAT_EQ(YGNodeLayoutGetHeight(child), 10);

  YGNodeFreeRecursive(root);

  YGConfigFree(config);
}
```

PR #702, opened 2018-01-24
Fixed rounding issues with YGRoundValueToPixelGrid and negative floats

Edensan:
While implementing Yoga into a custom framework I found out that sizes of objects that had been layout with Yoga were sometimes 1px smaller/bigger than expected, causing all sorts of unexpected logic to trigger.

Turns out the culprit was YGRoundValueToPixelGrid.

It was computing values regardless of sign and thus negative inputs were outputting unexpected results (eg. if scaledValue == -0.6f closest rounding should be -1.0f but would be 0.0f instead).

I have updated the code to handle positive and negative values symmetrically in order to ensure that two values with a same offset between them will conserve the offset during animations regardless of the rounding type.

Please check the test cases to confirm the logic.


Edensan:
What is going on with the pull request?

lunarraid:
This looks like it's covering the same issue as https://github.com/facebook/yoga/pull/688. I'd really like to see one of these merged in as I'm being forced to use a fork of this project due to this issue. Does anyone have an update on this?

PR #705, opened 2018-01-29
Fix use of free() when deleting YGNode

code-disaster:
Nodes are allocated using operator new, but released using free() instead of operator delete.


woehrl01:
Thank you for your PR @code-disaster. This change has already been introduced via 66045bd.

PR #712, opened 2018-02-07
solves the Android pixel bug

HendrixString:
@emilsjolander hi,
this PR solves the following common and probable layout pixel scenario:
the older code is presented for reference:

```java
view.measure(
    View.MeasureSpec.makeMeasureSpec(
        Math.round(node.getLayoutWidth()),
        View.MeasureSpec.EXACTLY),
    View.MeasureSpec.makeMeasureSpec(
        Math.round(node.getLayoutHeight()),
        View.MeasureSpec.EXACTLY));
view.layout(
    Math.round(xOffset + node.getLayoutX()),
    Math.round(yOffset + node.getLayoutY()),
    Math.round(xOffset + node.getLayoutX() + node.getLayoutWidth()),
    Math.round(yOffset + node.getLayoutY() + node.getLayoutHeight()));

```

suppose now the following:
- `xOffset + node.getLayoutX() = 2.2`
- `node.getLayoutWidth() = 0.4` ==> `Math.round(node.getLayoutWidth()) = 0`
- `Math.round(xOffset + node.getLayoutX() + node.getLayoutWidth()) = Math.round(2.2 + 0.4) = 3`

this induces, the following measurements:
```java
view.measure(
    View.MeasureSpec.makeMeasureSpec(
        0,
        View.MeasureSpec.EXACTLY),
    View.MeasureSpec.makeMeasureSpec(
        Math.round(node.getLayoutHeight()),
        View.MeasureSpec.EXACTLY));
view.layout(
    2,
    Math.round(yOffset + node.getLayoutY()),
    3,
    Math.round(yOffset + node.getLayoutY() + node.getLayoutHeight()));

```

the width measurement of the view is 0, while the layout is `(3 - 2 = 1)`.  
my proposed solution is to measure the view the way it is now, but when layouting
I use the `#getMeasuredWidth/Height()` methods, this will stop this problem
from happening.  

I also want to note that this bug happens with high probability.

### DIFF
--- a/android/src/main/java/com/facebook/yoga/android/YogaLayout.java
+++ b/android/src/main/java/com/facebook/yoga/android/YogaLayout.java
@@ -293,6 +293,8 @@ public class YogaLayout extends ViewGroup {
       if (view.getVisibility() == GONE) {
         return;
       }
+      int left = Math.round(xOffset + node.getLayoutX());
+      int top = Math.round(yOffset + node.getLayoutY());
       view.measure(
           View.MeasureSpec.makeMeasureSpec(
               Math.round(node.getLayoutWidth()),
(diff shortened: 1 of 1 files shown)


priteshrnandgaonkar:
@HendrixString Can you add a test case for this change ? So that its easier to catch the regression.

HendrixString:
hi @priteshrnandgaonkar 
I did some research about unit testing this.
Testing it will require adding more dependencies and
tasks to invoke them. If we choose native tests, then we
can only use connected devices tests which will force testing
it on a connected device. I have no idea if the architects of Yoga
would like that.

not using native test frameworks, will introduce a test framework
like [http://robolectric.org/](http://robolectric.org/), which can mock
the entire **android.jar** but I don't
believe that the architects will want to depend on that solution.

all and all, I came to a conclusion, that laying out a testing strategy,
that is beyond the regular/standard **JUnit** stack testing is probably
not my decision to make as I have no clear vision of how you want **Yoga**
to look like.

any thoughts ?

priteshrnandgaonkar:
Thanks for the PR. I am importing this PR for internal review.

PR #714, opened 2018-02-14
Fixes compilation for react-native-windows.

rozele:
Updates the following for compatibility with the latest react-native-windows:
- NuGet dependencies
- *.vcxproj and *.csproj files
- Removes designated initializer syntax

### DIFF
--- a/csharp/Windows/Facebook.Yoga.Universal.Tests/Facebook.Yoga.Universal.Tests.csproj
+++ b/csharp/Windows/Facebook.Yoga.Universal.Tests/Facebook.Yoga.Universal.Tests.csproj
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
   <PropertyGroup>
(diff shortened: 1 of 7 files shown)


rozele:
Similar to #700, except comments out designated initializers instead of removing them. cc @matthargett @priteshrnandgaonkar @woehrl01.

priteshrnandgaonkar:
@rozele We can have a macro which handles this case. Follow #700 PR thread.

rozele:
Closing in favor of #716

rozele:
Re-opening - I don't think the macro is the way to go here. Can we please merge this and deal with the macro in a follow up PR? I'd like to start using C++ code that ships in the react-native NPM package in react-native-windows ASAP.

rozele:
Ultimately, it all comes down to readability. Computers don't care about macros. IMO the following:
```
/* .position = */ {},
```
Is much more readable than:
```
DESIGNATED_INITIALIZER(position, {})
```
I understand that using the macro might make it clear that we're removing the designated initializer specifically for Windows, but I don't think it's worth the tradeoff in readability.

priteshrnandgaonkar:
I think we can use contructors. Whats your take @rozele

priteshrnandgaonkar:
Its an easy fix to have a constructor, it shouldn't take time. Let's not land the hacky solution, even if I import it, it would be rejected in an internal review.

rozele:
@priteshrnandgaonkar - if it's easy, do you mind taking the first crack at it? Since you're connected internally, you'll know better what's not going to be rejected on review. I'll happily test on windows once you have a branch.

matthargett:
“hacky” doesn’t seem like an appropriate characterization, in my opinion, for what it’s worth. Please just let the community know the non-C++20 syntax most likely to be accepted. Thanks! :)

priteshrnandgaonkar:
I am on it. Will add a change related to constructors which I am talking about. Then you can rebase your changes on it.

priteshrnandgaonkar:
My changes are merged, you can rebase it.

rozele:
Thanks @priteshrnandgaonkar. We're now building from Yoga source in react-native-windows, similar to how it works for iOS and Android. I'll likely copy the .vcxproj files into the react-native-windows repo so we can manage them more easily there. I plan to continue using the Facebook.Yoga.Shared C# project as a submodule in react-native-windows. I'm closing this PR.

rozele:
Re-opening, there were also some changes in this PR related to DLL exports that should be merged.

rozele:
@priteshrnandgaonkar - it looks like we're still missing constructors for YGValue and YGConfig.

PR #715, opened 2018-02-14
Fix main size calculation from the aspect ratio

synk:
When the following conditions are met, the main size become smaller by the margins in the main axis.
* The aspect ratio is defined
* The main size is not defined
* The cross size is defined
* The main margin is defined

This is because the main margin size is not included when calculating the main size from the aspect ratio.


priteshrnandgaonkar:
Awesome work, thanks for the PR. I have imported it in our system.

PR #716, opened 2018-02-16
Fixes compilation for react-native-windows.

rozele:
Updates the following for compatibility with the latest react-native-windows:
- NuGet dependencies
- *.vcxproj and *.csproj files
- Replaces designated initializer syntax with macro


rozele:
I don't think the macro is the way to go.

PR #722, opened 2018-03-02
Fix issue when view that not used Yoga layout

zhongwuzw:


NickGerleman (changes requested):
More information like a description of the issue and a test plan are needed for this PR to be actionable.

NickGerleman:

PR #726, opened 2018-03-08
Update pod spec to 1.8.0, set explicit public header files

a2:
- Update pod specs to version 1.8.0 (most recent tag)
- Mark `Yoga.h`, `YGEnums.h`, and `YGMacros.h` as public headers

### DIFF
--- a/Yoga.podspec
+++ b/Yoga.podspec
@@ -25,4 +25,5 @@ Pod::Spec.new do |spec|
       '-fPIC'
   ]
   spec.source_files = 'yoga/**/*.{c,h,cpp}'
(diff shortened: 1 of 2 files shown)


muukii:
I need this!

a2:
Reiterating this from facebook/rebound-js#49, I sent the CLA to the contact who replied to my email and never heard back. Is it possible for my CLA status to be updated?

NickGerleman:
It looks like the underlying code has been updated. The podspec is a newer version, and the headers are now public, since the RN build used them (though I'm not sure this is the right boundary to expose).

PR #736, opened 2018-03-20
expose flex property to YogaKit

axl411:
This exposes `yoga` property to `YogaKit`.

The reason why I need the property is demonstrated in https://github.com/axl411/TestYogaKit


zymxxx:
why closed?

need it!

PR #751, opened 2018-04-13
Use delete instead of free in YGConfigFree

lebek:
Since a `YGConfigRef` is created with `new`, it needs to be deleted using `delete` instead of `free`. Related to https://github.com/facebook/yoga/pull/705.


davidaurelio:
This was fixed in 006f6460a9d5b785654f122b5b05937149cc0729

davidaurelio:
thank you @lebek – I was about to merge this when I discovered that this was already addressed.

PR #761, opened 2018-05-01
`flexWrap` should be kebab-case inside inline-style

synk:
Inside css inline style, it should be spelled as `flex-wrap`.


wqweto:
While fixing `YGNodeToString` notice that `left`/`top`/`width`/`height` might use `appendFloatOptionalIfDefined` as these ara NaNs before getting calculated.

PR #772, opened 2018-05-20
Upgrade version of gatsby from 1.9.158 to 1.9.250 for Yoga documentation

Lutece:
gatsby-source-filesystem@1.5.31 depends on gatsby@1.9.250.
If we use Gatsby version lower than 1.9.250, we fail to render markdown files.


NickGerleman:
It looks like a newer version of gatsby is being used already. Going to close this out.

PR #818, opened 2018-10-01
Fix C# unit tests

rigdern:
This consists of several fixes:
  - Some of the C++ functions called by the tests were missing `WIN_EXPORT`
  - It looks like Yoga was changed to round up after `TestMeasureFuncWithFloat` was written. Here's the function call that results in the rounding: https://github.com/facebook/yoga/blob/357ca78f9f7bd86298ef00a1426cac5ca2041d3b/yoga/Yoga.cpp#L4019-L4026
  - The format of the result of the `Print` method was changed after `TestPrint` was written.

### DIFF
--- a/csharp/Windows/Facebook.Yoga.Universal.Tests/YogaNodeTest.cs
+++ b/csharp/Windows/Facebook.Yoga.Universal.Tests/YogaNodeTest.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using Microsoft.VisualStudio.TestPlatform.UnitTestFramework;
 using Facebook.Yoga;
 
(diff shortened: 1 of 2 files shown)


rigdern:
Questions
  - Should we use `WIN_EXPORT` on every public C function? My PR adds it to some that broke the tests but maybe we should just add it to all of them.
  - Are we supposed to use buck to run tests on Windows? It didn't work for me because it seemed to pass clang flags to Microsoft's C++ compiler.
  - There are some C#-specific tests. Which solution should I be opening for that?
    - `Facebook.Yoga.Desktop.sln`
    - `Facebook.Yoga.sln`
    - `Facebook.Yoga.Universal.sln`

/cc @mattpodwysocki @rozele @matthargett

csharp/Windows/Facebook.Yoga.Universal.Tests/YogaNodeTest.cs
```diff
@@ -285,7 +290,7 @@ public void TestCopyStyle()
         }
 
 #if !UNITY_EDITOR
-        private void ForceGC()
+    private void ForceGC()
```
shergin: Can we revert this line?

PR #834, opened 2018-11-24
Fix aspect ratio when stretching with main axis margin

emilsjolander:
I've noticed that when a child's size is determined by `align-items: stretch` in combination with `aspect-ratio` its size is wrongly calculated to account for margin in the main axis when there is more than enough space.

See playground: https://goo.gl/tgW6cD

I've yet to figure out exactly how to solve this but i've started by writing a failing test when can be seen in the first commit here.

I assumed I had found the bug here https://github.com/facebook/yoga/blob/master/yoga/Yoga.cpp#L1838 where margin is being subtracted from the desired width even though the measure mode tells it to measure to exactly that size. However, if we don't remove this margin from the available width then 15 tests fail (including the one I just added) not quite figured out why yet. I'm also a bit confused at to why this would only happen for nodes with `aspect-ratio` and not for nodes where an explicit height and width is set.

### DIFF
--- a/tests/YGAspectRatioTest.cpp
+++ b/tests/YGAspectRatioTest.cpp
@@ -896,3 +896,24 @@ TEST(YogaTest, aspect_ratio_should_prefer_flexed_dimension) {
 
   YGNodeFreeRecursive(root);
 }
+
+TEST(YogaTest, aspect_ratio_defined_by_cross_stretch_should_not_be_effected_by_margin_on_main_axis) {
+  const YGConfigRef config = YGConfigNew();
+  YGConfigSetUseWebDefaults(config, true);
+
+  const YGNodeRef root = YGNodeNewWithConfig(config);
+  YGNodeStyleSetWidth(root, 200);
+  YGNodeStyleSetHeight(root, 100);
+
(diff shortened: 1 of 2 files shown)


emilsjolander:
Ok the test I added is passing now.I noticed that this is not just an issue with `stretch` as seen here: https://goo.gl/Dqbpji

However I was not able to create a failing test for this, so my guess is that this was probably fixed previously and the yoga version used in the layout editor just hasn't been updated yet. Would like to confirm this though.

priteshrnandgaonkar:
The yoga used in playground has not been updated since a long time. There are few bugs which were fixed and still its reproducible on playground.

emilsjolander:
@priteshrnandgaonkar yeah that was my guess

astreet:
Nice catch! This fix doesn't seem right though (being constrained to just stretch) -- are you still working on a better fix or do you think this is all that's needed?

emilsjolander:
@astreet I think this is all that's needed. The fix is specifically for stretch as that is the case where it needs to be set as an exact measurement. Not sure what else would need to be covered by this.

PR #841, opened 2018-12-20
Fix issues with top, bottom, left, right keys in generated React Native code

dhruska:
This fixes two issues in the generation of React Native code.
* The `top: 10px` property is being generated as `positionTop: 10px` (as are bottom, left and right)
* `positionTop: 10px` and `positionBottom: 10px` is being combined into `positionVertical: 10px` (as is left and right into horizontal).


NickGerleman:
We are abandoning the old playground in favor of a new one, so closing this out

PR #842, opened 2018-12-31
Expose dirtied func in Javascript bindings

rmhartog:
Yoga currently has the concept of a 'dirtied func', which is called when a Node becomes dirty, which can be useful to trigger a repaint. This PR exposes that functionality in the Javascript bindings.

### DIFF
--- a/javascript/sources/Node.cc
+++ b/javascript/sources/Node.cc
@@ -23,6 +23,13 @@ static YGSize globalMeasureFunc(YGNodeRef nodeRef, float width, YGMeasureMode wi
     return ygSize;
 }
 
+static void globalDirtiedFunc(YGNodeRef nodeRef)
+{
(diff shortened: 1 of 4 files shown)


rmhartog:
Hello @davidaurelio I see that you've been actively working on some things in this project, do you think you could have a look at this? CI is failing, but that's unrelated (fixed in #843). Let me know if it's not okay to tag you like this and if there is a better way to get some eyes on these PRs 😄

davidaurelio:
ok, let’s merge it. Which way did you test this?

rmhartog:
I ported the `YGDirtiedTest` to JavaScript, but had not included in the PR for some reason, my apologies. It replicates the bahaviour of the `YGDirtiedTest.cpp`.

rmhartog:
Noticed that the `m_dirtiedFunc` should also be cleaned up when the node is reset, same as the measure function, added those changes to the branch.

PR #843, opened 2018-12-31
Add fb_native_wrapper.bzl to fix Travis CI

rmhartog:
When I was implementing #842, I noticed that CI was failing, but it did not seem to be caused by my changes. As you can see in the build log (https://travis-ci.org/facebook/yoga/jobs/473825507#L1217), the build depends on a build script that is not in the repository. In d02f4a94de589d91dc719e51ff6c8c5b302df184, @scottrice changed the build scripts to depend on `fb_native_wrapper.bzl` without introducing that file into the repository.

In this PR, I've introduced the bzl file, copied from the react-native project. Since it appears to only be a wrapper at this point, I'm assuming that this is the correct approach.


davidaurelio:
Thank you so much! Trying this out right now.

rmhartog:
Thanks for the feedback, let me know if there is anything I can do to help you :)

davidaurelio:
nothing you can do – [shipit](https://github.com/facebook/fbshipit) was configured a bit too strictly. I am waiting for a re-deploy now, and then we should be able to get this one in.

Issue #845, opened 2019-01-05
The sign problem caused by the C bit field


anyktx:
# Report

- [ ] I have searched [existing issues](https://github.com/facebook/yoga/issues) and this is not a duplicate

# Issues and Steps to Reproduce
I used Visual Studio 2017 to code and compile, and some of the properties went wrong when I set the Style property of YGNode and got the layout results.

When I looked at the definitions of YGStylet and YGLayout, I found that some of the properties were bitfields used. In this case, having this property causes a sign problem. For example, the value of flexDirection is [0,3], when the value is 2 or 3, the bit field that USES two bits will take the previous bit as the symbol, resulting in a negative value. Without experimenting with other compilers, there may not be such a problem, but in Visual Studio such a problem does exist. The properties found to have this problem exist in almost all the properties that use bit fields.

# Link to Code
https://github.com/facebook/yoga/blob/master/tests/YGFlexDirectionTest.cpp 
This test won't work well in Visual Studio，

zcbenz:
I can also reproduce this problem, in my case Yoga crashed directly because of wrong `flexDirection` value.

Adlai-Holler:
This is blocking my build when I try to import YGNode.h into a C++ context – the compiler errors with "Implicit truncation to bitfield will change value from 4 to -4." I tried changing the enum definition to be `unsigned` instead of `int` but that no workm.

PR #846, opened 2019-01-05
Increase that number of bit in YGStyle and YGLayout to resolve the sign issue.

anyktx:
This may not be the best way to solve the issue of [https://github.com/facebook/yoga/issues/845], but it is an effective one.

### DIFF
--- a/yoga/YGLayout.h
+++ b/yoga/YGLayout.h
@@ -17,7 +17,7 @@ struct YGLayout {
   std::array<float, 6> margin = {};
   std::array<float, 6> border = {};
   std::array<float, 6> padding = {};
-  YGDirection direction : 2;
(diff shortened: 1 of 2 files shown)


Adlai-Holler:
+1 to get this in the bank if possible. My C++ compiler refuses to build YGNode.h because of the implicit truncation changing the sign of the value from 4 to -4.

Adlai-Holler:
@davidaurelio One way to fix this simply is to change `YG_ENUM_BEGIN` from `int` to `unsigned`

Adlai-Holler:
cc #870

davidaurelio:
I fixed this last August. However, I cannot close here since I am no longer at FB.

PR #791, opened 2018-07-06
Makes Yoga threadsafed:

nokia6686:
Issue ref: https://github.com/facebook/yoga/issues/769
### Problem
- `gNodeInstanceCount`, `gConfigInstanceCount`, `gCurrentGenerationCount`, `gDepth` are global variables in `Yoga.cpp`. They are not safe when using multithreading.
### My idea
- Uses `atomic` for `gNodeInstanceCount` and  `gConfigInstanceCount` because of counting instances only.
- Store `gCurrentGenerationCount` and `gDepth` in root node for multithreading.
### My solution
- Uses <atomic> for `gNodeInstanceCount` and  `gConfigInstanceCount`
```C++
std::atomic<int32_t> gNodeInstanceCount(0);
std::atomic<int32_t> gConfigInstanceCount(0);
```
- Every Yoga Node has its root reference (root-ref). New node's root point to itself. Root-ref will be updated when inserting children.
- `gCurrentGenerationCount` and `gDepth` are stored in `YGNode`.
- When updating children, we must update root-ref for each new child.
```c++
struct YGNode {
...
  // We must setChildRoot() for new child
  void replaceChild(YGNodeRef oldChild, YGNodeRef newChild);
  void replaceChild(YGNodeRef child, uint32_t index);
  void insertChild(YGNodeRef child, uint32_t index);
...
  private:
    YGNodeRef pRoot = nullptr;

  public:
    uint32_t  gCurrentGenerationCount = 0;
    uint32_t  gDepth = 0;
    YGNodeRef getRoot();
    void setChildRoot(YGNodeRef node);
}
...
void YGNode::setChildRoot(YGNodeRef node) {
  node->pRoot = getRoot();
  for (auto child: node->children_) {
    setChildRoot(child);
  }
}
```
- Global variables are accessed via root-ref
```C++
//gCurrentGenerationCount++;
node->getRoot()->gCurrentGenerationCount++;

//gDepth++;
node->getRoot()->gDepth++;
```

PR #852, opened 2019-01-08
make yoga threadsafe

tcdat96:
Continuing https://github.com/facebook/yoga/pull/791
@nokia6686 is a former member of our team, so we are trying to pick up what he left and carry out the pull request.
# Solution
Improved from previous solution with @jpap's suggestions.
1. Using **atomic** for ```gNodeInstanceCount``` and ```gConfigInstanceCount```
```
std::atomic<int32_t> gNodeInstanceCount(0);
std::atomic<int32_t> gConfigInstanceCount(0);
```
2. Passing ```gDepth``` and ```gCurrentGenerationCount``` (renamed to **_depth_** and **_generationCount_** respectively) between function calls that stem from ```YGNodeCalculateLayout```.  
In ```YGNodeCalculateLayout```, pass ```depth``` as value 0, to indicate the root depth.
3. Using ```#define``` with ```gPrintChanges``` and ```gPrintSkips``` to optimized out unnecessary logging and incrementing depth state.
```
#define PRINT_CHANGES false
#define PRINT_SKIPS false
```
4. Prevent memory leak by removing **static** keyword in ```YGConfigGetDefault```


yoga/Yoga.cpp
```diff
@@ -4059,12 +4098,15 @@ void YGNodeCalculateLayout(
     const float ownerWidth,
     const float ownerHeight,
     const YGDirection ownerDirection) {
+  uint16_t depth = 0;
+  uint16_t generationCount = 0;
```
jpap: I suspect the `generationCount` should remain as a member of the `YGNodeRef`, and incremented on each invocation of `YGNodeCalculateLayout(...)` so that subsequent calls can reuse computed layout when things don't change (given by an equal generation).  Subsequently passing the new generation counter around by function call as you've done is a good idea, to avoid the pointer-to-root situation of the last PR.

yoga/Yoga.cpp
```diff
@@ -1296,8 +1298,7 @@ static void YGNodeComputeFlexBasisForChild(
     if (child->getLayout().computedFlexBasis.isUndefined() ||
         (YGConfigIsExperimentalFeatureEnabled(
              child->getConfig(), YGExperimentalFeatureWebFlexBasis) &&
-         child->getLayout().computedFlexBasisGeneration !=
-             gCurrentGenerationCount)) {
+         child->getLayout().computedFlexBasisGeneration != generationCount)) {
```
jpap: The {`computedFlexBasisGeneration`, `generationCount`} members of `YGLayout` type should be updated to 16-bit unsigned.

yoga/Yoga.cpp
```diff
@@ -3527,9 +3561,8 @@ static void YGNodelayoutImpl(
   }
 }
 
-uint32_t gDepth = 0;
-bool gPrintChanges = false;
-bool gPrintSkips = false;
+#define PRINT_CHANGES false
+#define PRINT_SKIPS false
```
jpap: This is probably OK, but it is safer to use `#define XXX 0` here for safer compatibility against various pre-processors, that look for a "C expression of integer type" (gcc) or "must have integral type and can include only integer constants, character constants, and the defined operator" (msvc).

yoga/Yoga.cpp
```diff
@@ -4059,12 +4098,15 @@ void YGNodeCalculateLayout(
     const float ownerWidth,
     const float ownerHeight,
     const YGDirection ownerDirection) {
+  uint16_t depth = 0;
```
jpap: It would be better to use `int` for the depth type, so as not to limit to a tree height of 65,535.  (The generation count, a separate thing, is likely OK to wrap at 16-bits.)

To simplify this further you might consider removing the variable declaration, and writing it directly as an argument when invoking `YGLayoutNodeInternal` below.  For example:
```
  if (YGLayoutNodeInternal(
          node,
          width,
          height,
          ownerDirection,
          widthMeasureMode,
          heightMeasureMode,
          ownerWidth,
          ownerHeight,
	  true,
	  "initial",
	  node->getConfig(),
	  0, // tree root
	  generationCount)) {
```

yoga/Yoga.cpp
jpap: My apologies; what I meant there was:
* Store the generation in the `YGNodeRef`, with a per-node `uint16_t` two-byte overhead: let's refer to it as `generationCount_` below.
* When we call `YGNodeCalculateLayout`, we're passing the root node for the subtree to be laid out, so we can immediately increment (update) `generationCount_` of that root node, and then assign the newly incremented value to the `generationCount` local variable you have defined, instead of always starting from the zero value.  (There is a second increment in the same function, where we would do the same as above.)
   * For a single tree, when we go back to re-compute a subsequent layout, we end up using the equivalent value of `generationCount` as the existing global-variable based implementation.  That's nice.
   * For a second (and subsequent) tree, it gets its own independent `generationCount`, which is also nice because each tree is now separate.  (Thread-safety also follows.)
   * When we share (clone) nodes, and use an arbitrary descendent of the original tree root to lay out a now-independent subtree, we will use a new version of the `generationCount`, likely zero.  This may result in some nodes being recomputed as the generation now differs.  If this is an issue, we could update `YGNodeClone` so as to follow the `YGNode.owner_` path up to the root, to copy the root `generationCount` when a node is cloned.  (I did not consider this earlier; your thoughts here are appreciated.)
* Once we have seeded the `generationCount` local variable in `YGNodeCalculateLayout`, we can pass it around on the stack, as the layout proceeds down the tree, thus avoiding the pointer-to-the-root issue.

yoga/YGMacros.h
```diff
@@ -38,3 +38,10 @@
 #define YG_ENUM_BEGIN(name) enum name
 #define YG_ENUM_END(name) name
 #endif
+
+#if !defined(TRUE)
+#define TRUE 1==1
```
jpap: What you need here is simply a non-zero integer value.
```suggestion
#define TRUE 1
```

yoga/Yoga.cpp
tcdat96: Please take a look at the latest commit, @jpap. When cloning nodes, I went with the traversal approach, thought that would be much cheaper than re-computing the differing nodes.
If there's still any issue, please let me know.

yoga/Yoga.cpp
```diff
@@ -240,6 +240,11 @@ YGNodeRef YGNodeClone(YGNodeRef oldNode) {
       "Could not allocate memory for node");
   gNodeInstanceCount++;
   node->setOwner(nullptr);
+
+  YGNodeRef root = node;
+  for (; root->getParent() != nullptr; root = root->getParent());
```
jpap: Looks like this loop always terminates early: the parent is set to `nullptr` above the loop on line 242.

Might be a good idea to add a test? ;-)

tcdat96: Oh my apologies, thought it just a simple loop, so didn't test it thoroughly :-(
Okay, let me add a test :+1:

jpap: No worries.

Thinking about this a bit more, I'm not sure we need to do this extra work on clone: the generation is incremented each time we perform a `YGNodeCalculateLayout`, so even if we preserved the root generation counter for a cloned subtree, we're still incrementing it on a call to `YGNodeCalculateLayout` for that subtree and before we proceed with any work.  Any comparison to the old generation count will be false.

Is the counter used to help avoid node recalculations **during** multiple passes of a given layout (generation) calculation?  I'd need to spend some time to really understand how the generation counter is used for this; and unfortunately the code isn't terribly straightforward or self-documenting.  If you're on the FB team, can you ask the original author for a short-cut?

tcdat96: Unfortunately we aren't from the FB team and neither do we have a full understanding how generetionCount works. My team encounter the race condition problem several months ago, so we opened the previous pull request to try to fix that.  
I actually thought you are the one on the FB team ;-)

jpap: > I would suggest us to stick to the original idea of #791 because it focus on solving the data race condition issue without worrying about optimizing the `generationCount` itself. This can be done in a separated PR and leave this PR with minimal changes.

That would be unwise because #791:
1. uses more heap than necessary;
1. introduces an expensive subtree traversal to set up the pointer-to-the-root, which is performed often; and
1. does not address node cloning.

I am reading through the code to understand how the generation is treated and will report back shortly.

jpap: Here's what I've learned from reading the code...

There are two generation counters: `generationCount` and `computedFlexBasisGeneration` stored in `YGLayout`, which are compared to the global `gCurrentGenerationCount`.  An overview of each is given below.

At the end, I'll provide a few options on how to move forward with this PR.  Apologies for the length in advance.

#### generationCount

1. This counter is used to invalidate the layout cache in `YGLayoutNodeInternal` when a node is dirty, and the node has not been visited thus far for a given `YGNodeCalculateLayout` call.
1. It is updated on exit of the function whenever layout is performed.
1. In `YGNodeCalculateLayout`, the `gCurrentGenerationCount` is incremented a second time before performing the optional a legacy "stretch behavior" comparison on a cloned tree that has all nodes marked dirty.  The new `gCurrentGenerationCount` here is to ensure the layout cache is cleared as above.
1. It looks like the purpose of this counter is to ensure the layout cache is invalidated whenever a node is dirty once per `YGNodeCalculateLayout` call.


#### computedFlexBasisGeneration

1. This counter is used exclusively in `YGNodeComputeFlexBasisForChild`.
1. It gates computation of a child's computed flex basis when the `YGExperimentalFeatureWebFlexBasis` feature is in use.
1. It is updated on exit of the function every time.
1. It looks like the purpose of this counter is to limit the above computation once per `YGNodeCalculateLayout` call.

#### gCurrentGenerationCount

1. This counter is incremented in `YGNodeCalculateLayout` before performing a full-tree layout.
1. It looks like the purpose of this counter is to uniquely identify a given layout run, which makes sense when you consider its name: "generation" implying that each layout gives birth to a new tree.
1. Because of the way the counter is used (compare with equality only; node update on exit of relevant function), I don't think it matters when it wraps.  This means we could probably get away with an even smaller storage type (e.g. `uint8_t`) for the global counter and associated class member storage.
1. With the above in mind, the comments in the source now make more sense (to me):
    ```
    // Instead of recomputing the entire layout every single time, we
    // cache some information to break early when nothing changed
    ```
    and
    ```
    // Increment the generation count. This will force the recursive routine to
    // visit
    // all dirty nodes at least once. Subsequent visits will be skipped if the
    // input
    // parameters don't change.
    ```

### Moving forward

Some options:

1. Minimize changes:
    * Drop the `YGNodeClone` change.
    * Remove the `generationCount_` member from `YGNode` and bring back the global `gCurrentGenerationCount`.
    * Increment and retrieve `gCurrentGenerationCount` atomically in `YGNodeCalculateLayout` and use it to seed the `generationCount` stack variable that is passed around by function call.
    * Keep the other changes in the current PR, relating to `depth` and `PRINT_XXX`.
1. Minimize heap memory use:
    * Drop the `gCurrentGenerationCount` global.
    * Drop the `computedFlexBasisGeneration` and `generationCount` members of `YGLayout`.
    * Declare a new `std::map<YGNode,uint8_t>` stack variable in `YGNodeCalculateLayout` that is passed around like `depth` is in the current PR.  (Or to minimize function call overhead, the `depth` and a pointer to the map can be placed in a struct, and a pointer to that struct is passed around instead.)
    * The `uint8_t` value field of the map is a bitfield:
        * reserve bit 0 for "I have visited the node" in `YGLayoutNodeInternal` in lieu of the `layout->generationCount != gCurrentGenerationCount` check that currently exists.
        * reserve bit 1 for "I have computed the flex basis for this child" in lieu of the `computedFlexBasisGeneration != gCurrentGenerationCount` check in `YGNodeComputeFlexBasisForChild`.
    * Throw out the map on return of `YGNodeCalculateLayout`.

What do you think?

tcdat96: The implementation for the first and second options would be quite straightforward, but I'm having a hard time understanding the third optimization. How exactly can we substitute those 3 variables with the map approach? Personally, I find it really risky to make such a big change, and we don't really have the confidence in our understanding of Yoga to carry out the implementation by ourselves neither.

jpap: I've submitted PR #855 that takes the third approach: it appears to work well!

So from here, let's see if @davidaurelio or others on the FB team can offer their input.  @tcdat96, in the meantime if you have time to update your PR with option #2, it can give others an opportunity to review both PRs and weigh in.

Hopefully we can land this thread-safety fix soon.  I'm happy to amend my PR if there are preferences on style, etc.

tcdat96: I updated the PR with the first 2 options, but ran into error ```undefined reference to `pthread_getspecific'``` when running gtest. Have you ever encountered a similar problem, @jpap?

jpap: > error `` undefined reference to `pthread_getspecific' `` when running gtest.

Sadly I've never experienced that; sounds like a linker configuration problem?  Perhaps you've inadvertently modified some of the build configuration someplace.  I would try:

* Clean and rebuild?
* Create a new local branch based off master, then cherry-pick **only** your changes then rebuild/test.

tcdat96: I've already tried both but unfortunately nothing works. The error appeared even with the master commit so I think it should be my configuration problem, but sadly I don't really have any experience in this type of problem :-(

tcdat96: It seems that the error only emerges in our ubuntu machines, the same source code runs without any error in MacOS. 
I modified the BUCK file in lib/gtest with the workaround described in https://github.com/facebook/buck/issues/1443 and it worked. Now this PR runs successfully with all tests passed.

yoga/Yoga.h
```diff
@@ -12,6 +12,7 @@
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <atomic>
```
davidaurelio: move to `Yoga.cpp`

a) it’s not used here
b) this header needs to be valid C

yoga/Yoga.cpp
```diff
@@ -3478,10 +3514,10 @@ static void YGNodelayoutImpl(
   }
 }
 
-uint32_t gDepth = 0;
-bool gPrintChanges = false;
-bool gPrintSkips = false;
+#define PRINT_CHANGES FALSE
+#define PRINT_SKIPS FALSE
 
+#if PRINT_CHANGES
```
davidaurelio: regular `if (kPrintChanges)`

yoga/Yoga.cpp
```diff
@@ -1007,7 +1007,7 @@ bool YGNodeLayoutGetDidLegacyStretchFlagAffectLayout(const YGNodeRef node) {
   return node->getLayout().doesLegacyStretchFlagAffectsLayout;
 }
 
-uint32_t gCurrentGenerationCount = 0;
+std::atomic<uint8_t> gCurrentGenerationCount(0);
```
davidaurelio: Again, this causes additional synchronisation, and I am not sure it’s necessary:

this will usually be loaded into a register, incremented, and written back to memory.

When two threads do this in parallel, they might both load the old value, increment by one, and write back – with the global value being incremented by one, not two, as a result.

That will still increment the count, and cause layouts to be run – there isn’t really a need to synchronise it.

Am I missing something?

Minor nitpick, please use uniform initialization (`std::atomic<uint8_t> gCurrentGenerationCount{0};`)

yoga/Yoga.cpp
```diff
@@ -206,8 +206,8 @@ void YGNodeMarkDirtyAndPropogateToDescendants(const YGNodeRef node) {
   return node->markDirtyAndPropogateDownwards();
 }
 
-int32_t gNodeInstanceCount = 0;
-int32_t gConfigInstanceCount = 0;
+std::atomic<int32_t> gNodeInstanceCount(0);
+std::atomic<int32_t> gConfigInstanceCount(0);
```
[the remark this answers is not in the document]
tcdat96: Honestly, I don't have the full knowledge about how these variables could affect the correctness of Yoga. These variables were flagged by ThreadSanitizer to be a possible data race problem, so I only tried to fix it at the language level.
But if its accuracy is not important to the program's flow, then what's the point of having these variables?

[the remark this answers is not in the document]
jpap: These instance tracking variables are only used for tests. You could conditionally compile in this functionality for tests by adding a new preprocessor token in the `BUCK` file, under `TEST_COMPILER_FLAGS`.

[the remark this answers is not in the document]
jpap: These instance tracking variables are only used for tests. You could conditionally compile in this functionality for tests by adding a new preprocessor token in the `BUCK` file, under `TEST_COMPILER_FLAGS`.

yoga/YGLayout.h
```diff
@@ -22,12 +22,12 @@ struct YGLayout {
   bool doesLegacyStretchFlagAffectsLayout : 1;
   bool hadOverflow : 1;
 
-  uint32_t computedFlexBasisGeneration = 0;
+  uint8_t computedFlexBasisGeneration = 0;
```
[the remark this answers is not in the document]
jpap: The motivation is to save heap: there are many bits left over from alignment of the bitfields above it, so you could combine `computedFlexBasisGeneration` with it.  I suspect you can get away with just one bit, so long as the generation counter is the same width (or is truncated with a logical AND mask when the generation is incremented).

yoga/YGMacros.h
```diff
@@ -38,3 +38,10 @@
 #define YG_ENUM_BEGIN(name) enum name
 #define YG_ENUM_END(name) name
 #endif
+
+#if !defined(TRUE)
+#define TRUE 1
+#endif
+#if !defined(FALSE)
+#define FALSE 0
+#endif
```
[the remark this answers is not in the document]
tcdat96: I'll switch to constexpr in another pull request, so we won't need this anymore

yoga/Yoga.cpp
```diff
@@ -224,7 +224,7 @@ WIN_EXPORT YGNodeRef YGNodeNewWithConfig(const YGConfigRef config) {
 }
 
 YGConfigRef YGConfigGetDefault() {
-  static YGConfigRef defaultConfig = YGConfigNew();
+  YGConfigRef defaultConfig = YGConfigNew();
```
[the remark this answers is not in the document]
tcdat96: This is actually my misunderstanding about how static variables work. This was marked by CppUTest as a memory leak but it appears to be only a false alarm. I'll revert this.

tcdat96:
@davidaurelio Sorry for delaying this for too long, I was a little too busy last month.

In the new update, I've:
* reverted the atomics
* reverted the macros. I'll switch to ```constexpr``` in another pull request
* reverted the static mistake

So this pull request will only have 2 changes now:
1. using **uint8_t** for ```generationCount```
2. passing ```depth``` and ```generationCount``` via function parameters, instead of using global variables.

tcdat96:
any update, @davidaurelio ?

rockerhieu:
@davidaurelio @jpap hello?

appleguy:
Hey all - this is still flagging our thread-sanitizer runs. It looks bad for Yoga's usage at a certain big company. Could the team find some way to fix or silence the warning with annotations?

rockerhieu:
@davidaurelio it's been several months. Could you take a look at this?

davidaurelio:
Hi folks, sorry for being unresponsive.

The aspect of this creating too much noise in sanitizers is new to me, sorry about that. I thought this was mainly due to concerns about the data races themselves.

Let me see how well this merges …

Issue #769, opened 2018-05-16
[Thread Safety] Global variables are immediately flagged by Thread Sanitizer


appleguy:
# Report

Couldn't find any duplicate for "thread sanitizer" or "global".

When launching an app that uses Yoga concurrently, even when there is no concurrent access to a particular set of YGNodeRefs, the Thread Sanitizer triggers several warnings due to unprotected concurrent writing and reading from global variables.

# Issues and Steps to Reproduce

Build and run an app that uses Yoga concurrently. This can be done with the Texture framework (texturegroup.org), or just by dispatch_async'ing layouts to various independent YGNodeRef trees.

# Expected Behavior
Independent YGNodeRef trees should be thread-safe to operate on, as long as the app-provided implementation of the MeasureFunc and BaselineFunc are themselves thread-safe.

# Actual Behavior
When running under TSAN, several operations are immediately flagged as not thread-safe. For more background on TSAN: https://developer.apple.com/videos/play/wwdc2016/412/

# Link to Code
1st one: gNodeInstanceCount (very common)
2nd one: gCurrentGenerationCount (very common)
3rd one: gDepth (a bit less common than the above two)

Note that the C++ increment operator is not atomic: https://stackoverflow.com/a/39396999

# Options to fix
Using C++ atomics would be one improvement. However this would be insufficient to create transactional integrity, e.g. if an atomic value is read and then a condition statement entered, reading the atomic again to do an operation with it is not transactional.

The best fix would be to avoid maintaining global variables if at all possible, and instead store these on the root YGNodeRef. If locking is added, it should be done with care to ensure deadlocks and thread contention are not an issue.

![gnodeinstancecount unsafe global](https://user-images.githubusercontent.com/565251/40134380-16543206-58f7-11e8-98e7-284f56e2c7dc.png)
![gcurrentgenerationcount unsafe global](https://user-images.githubusercontent.com/565251/40134378-162a722c-58f7-11e8-9cc7-7efceee5aaa3.png)
![gdepth unsafe global](https://user-images.githubusercontent.com/565251/40134379-1641da70-58f7-11e8-9bb9-7a504a4ac0c2.png)

emilsjolander:
Agreed. This is a known issue and we have plans on moving all this into the context / node object. Not something we are working on right now but I hope we can have a look at solving this within the next month. If you have time we would also really appreciate a PR.

rockerhieu:
@emilsjolander would you mind take a look at #786? Thanks!

[...]

PR #855, opened 2019-01-17
Addresses thread safety, GitHub issue #769.

jpap:
See also: discussions in PR #852 and #786.

In addition to improving thread safety, it removes 8 bytes of heap per node.

### DIFF
--- a/yoga/YGLayout.h
+++ b/yoga/YGLayout.h
@@ -22,12 +22,8 @@ struct YGLayout {
   bool doesLegacyStretchFlagAffectsLayout : 1;
   bool hadOverflow : 1;
 
-  uint32_t computedFlexBasisGeneration = 0;
   YGFloatOptional computedFlexBasis = {};
 
-  // Instead of recomputing the entire layout every single time, we cache some
-  // information to break early when nothing changed
-  uint32_t generationCount = 0;
   YGDirection lastOwnerDirection = (YGDirection) -1;
 
   uint32_t nextCachedMeasurementsIndex = 0;
--- a/yoga/YGNode.cpp
+++ b/yoga/YGNode.cpp
@@ -189,11 +189,6 @@ void YGNode::setLayoutPosition(float position, int index) {
   layout_.position[index] = position;
 }
 
-void YGNode::setLayoutComputedFlexBasisGeneration(
-    uint32_t computedFlexBasisGeneration) {
-  layout_.computedFlexBasisGeneration = computedFlexBasisGeneration;
-}
-
 void YGNode::setLayoutMeasuredDimension(float measuredDimension, int index) {
   layout_.measuredDimensions[index] = measuredDimension;
 }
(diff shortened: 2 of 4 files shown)


jpap:
Looks like the android Travis CI configuration is borked:
```
Buck wasn't able to parse /home/travis/build/facebook/yoga/lib/soloader/BUCK:
IOError: [Errno 2] No such file or directory: '/home/travis/build/facebook/yoga/tools/build_defs/fb_native_wrapper.bzl'
Call stack:
  File "/home/travis/buck/python-dsl/buck_parser/buck.py", line 1748, in process_with_diagnostics
    package_implicit_load=package_implicit_load,
  File "/home/travis/buck/python-dsl/buck_parser/buck.py", line 1638, in process
    package_implicit_load=package_implicit_load,
  File "/home/travis/buck/python-dsl/buck_parser/buck.py", line 1626, in _process_build_file
    package_implicit_load=package_implicit_load,
  File "/home/travis/buck/python-dsl/buck_parser/buck.py", line 1561, in _process
    exec(code, module.__dict__)
  File "/home/travis/build/facebook/yoga/lib/soloader/BUCK", line 6
    load("//tools/build_defs:fb_native_wrapper.bzl", "fb_native")
  File "/home/travis/buck/python-dsl/buck_parser/buck.py", line 1292, in _load
    inner_env, module = self._process_include(build_include, is_implicit_include)
  File "/home/travis/buck/python-dsl/buck_parser/buck.py", line 1586, in _process_include
    package_implicit_load=None,
  File "/home/travis/buck/python-dsl/buck_parser/buck.py", line 1550, in _process
    with open(path, "r") as f:
```

This PR builds OK and tests pass on my local macOS machine.

rmhartog:
There seems to be a file missing for CI indeed, I've proposed a fix in #843.

davidaurelio:
This introduces double-digit regressions for layout calculation on ARM. I’m in favour of #852 at the moment, and will investigate whether we can remove the node counts.

davidaurelio:
@jpap for context: we have an internal device lab that will run benchmarks on physical Android and iOS devices. As much as I would like everyone to be able to run benchmarks, this will not become a publicly available resource.

Some numbers:
- The synthetic benchmark that measures *allocation, style assignment, layout calculation, layout reads, and deallocation* recessed by ~11% (Android)
- *calculation in isolation* recessed by ~11% on Android, ~8.5% on iOS
- *allocation/deallocation* in isolation recessed by ~35% (Android)
- The losses are smaller when using the Java API, as there’s fixed JNI overhead 

These numbers don’t reflect real-life scenarios accurately (we will get there in the coming months), but they give an indication about the performance characteristics of changes.

> The other approach in #852 suffers from not being able to concurrently perform a layout of a given tree, and any number of cloned subtrees. This is because per-layout state is stored in each (shared) node, and the generation count is different for each layout run. (If this is not of practical concern, it could be documented as a limitation.)

For now this is an acceptable limitation, and I agree that it should be documented. Shared subtrees are cloned lazily during layout calculation, and the only client making use of that is React Native. In their case, that’s thread safe, as the two highest ancestors sharing a node are never identical, and React Native treats subtrees as immutable.

That being said, it would be nicer to evolve Yoga into a direction where we can separate the layout outputs from the nodes, but this involves migrating client frameworks as well, including Litho, ComponentKit, React Native, React Native Windows, and internal projects at Facebook and other companies.

I think the short term goal should be to make parallel calculations on different trees possible.

Apologies if I am not always super responsive on Github. I really appreciate the input, though!

PR #856, opened 2019-01-25
Carthage support

sjmueller:
This introduces working carthage support for with the latest on master. There are two problems though:
- it probably breaks cocoapods support
- the sample project is deleted because I couldn't figure out how to make carthage use the right project

Use `github "sjmueller/yoga" "master"` to try it out. Hopefully this serves as a helpful starting point to supporting carthage.


davidaurelio:
Can we do this without having to delete the example and without breaking CocoaPods? I am no iOS expert; ComponentKit’s Carthage build manages to use Yoga today: https://github.com/facebook/componentkit/blob/master/Cartfile.private#L4

davidaurelio:

NickGerleman:
I don't think we'll want to move forward with this, vs CocoaPods and/or SPM.

PR #858, opened 2019-01-30
minus total outer flex basis for available width or height, when compute flex basis for children

shiyizju:
I put an ImageView and an UILabel in a horizontal flex layout, the label's size is not calculated correctly. The size passed to UILabel's sizeThatFits is not correct. I think it should minus the previous calculated totalOuterFlexBasis. 

Thanks.

### DIFF
--- a/yoga/Yoga.cpp
+++ b/yoga/Yoga.cpp
@@ -1937,6 +1937,20 @@ static void YGNodeComputeFlexBasisForChildren(
   }
 
   for (auto child : children) {
+
+    float availableWidthForChild = availableInnerWidth;
+    float availableHeightForChild = availableInnerHeight;
+
+    if (YGFlexDirectionIsRow(mainAxis)) {
+      if (!YGFloatIsUndefined(availableInnerWidth)) {
+        availableWidthForChild -= totalOuterFlexBasis;
+      }
+    } else {
+      if (!YGFloatIsUndefined(availableInnerHeight)) {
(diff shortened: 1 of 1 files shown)


woehrl01:
Thanks @shiyizju for your pull request.

Could you please add a test case for this change, to prevent a possible regression? 

Thanks!

shiyizju:
> Thanks @shiyizju for your pull request.
> 
> Could you please add a test case for this change, to prevent a possible regression?
> 
> Thanks!

Thanks @woehrl01  for reply. 

I ran the test cases and it failed with following output:

====STANDARD OUT====
tests/YGPersistenceTest.cpp:208: Failure
      Expected: YGNodeGetChild(root2_child1, 0)
      Which is: 0x7f8eb9809200
To be equal to: root_child1_0
      Which is: 0x7f8eb9808c00
TESTS FAILED: 1 FAILURE
Failed target: //:YogaTests
FAIL //:YogaTests

I don't understand the test case very well referenced in the output, could you please help check if I miss something, or maybe my change is not correct?

Thanks.

shiyizju:
I met this problem in iOS environments with YogaKit. Following I upload a demo to reproduce this problem, please take a look. Let me know if there is any problem. Thanks!

https://github.com/shiyizju/YogaTryout

shiyizju:
Hi, @davidaurelio , could you please let me know if this issue if being followed up? Thanks.

davidaurelio:

PR #859, opened 2019-01-30
Add "space-evenly" to justify-content documentation and adjusts others

MoOx:
I wanted to add missing docs for space-evenly, so I simply opened css specs https://drafts.csswg.org/css-align-3/.
While I was on it, I tried to simplify space-around description to not have to compare each space-* to each others.
Finally, I adjusted flex-* and center by just ripping css specs like I did for the other desc.

Feel free to reject some parts of this update. I think some other pages needs the same update.

PS: for the max-len of each lines, I just used prettier (locally on my editor) to avoid having lines too long, I think that could be a good idea to put that for other files too, I can help with that, it's very easy to setup.

### DIFF
--- a/website/contents/properties/justify-content.md
+++ b/website/contents/properties/justify-content.md
@@ -6,23 +6,34 @@ hasPlayground: true
 
 ## Justify Content
 
-Justify content describes how to align children within the main axis of their container.
-For example, you can use this property to center a child horizontally within a container
-with `flex direction` set to `row` or vertically within a container with `flex direction`
-set to `column`.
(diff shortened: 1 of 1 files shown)


davidaurelio:

NickGerleman (changes requested):
There is some language here that seems more ambiguous than before. E.g. Center no longer specifies the axis, or we add specific terms like "main-start" or "cross-start" which are not consistent with other language.

We should do an update pass of the documentation (maybe when I migrate it to Docusaurus), but I think we the previous language here was concise and something we should keep/replicate.

PR #864, opened 2019-03-04
YGJNI.cpp: Fix unused variables

empyrical:
## Summary

This pull request introduces a `YG_UNUSED` macro into `YGJNI.cpp`. This will suppress these errors that was causing this build failure:

```
ReactAndroid/src/main/jni/first-party/yogajni/jni/YGJNI.cpp:792:15: error: variable ‘edge’ set but not used [-Werror=unused-but-set-variable]
         float edge = *styleInputs++;
               ^~~~
```

### DIFF
--- a/java/jni/YGJNI.cpp
+++ b/java/jni/YGJNI.cpp
@@ -11,6 +11,8 @@
 #include <iostream>
 #include <map>
 
+#define YG_UNUSED(x) (void) (x);
+
 using namespace facebook::jni;
 using namespace std;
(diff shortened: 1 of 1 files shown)


davidaurelio:
Hi @empyrical. Thank you for contributing a solution for this problem exposed by GCC < v8.

The problem has been fixed by rearranging some casts in a9bddf87ff8ed0ab40fde3962aee3a8682c03319 – that solution doesn’t need any unused constructs.

PR #870, opened 2019-03-11
Use unsigned for enums to avoid bitfield compiler warning

### LINKED ISSUE #845: The sign problem caused by the C bit field

Adlai-Holler:
Fixes #845 without adding more bits.

### DIFF
--- a/yoga/YGMacros.h
+++ b/yoga/YGMacros.h
@@ -23,10 +23,12 @@
 #ifdef NS_ENUM
 // Cannot use NSInteger as NSInteger has a different size than int (which is the
 // default type of a enum). Therefor when linking the Yoga C library into obj-c
-// the header is a missmatch for the Yoga ABI.
-#define YG_ENUM_BEGIN(name) NS_ENUM(int, name)
+// the header is a missmatch for the Yoga ABI. We use unsigned here because our
+// enums are wrapped in limited bitfields where we use every single bit - a signed
+// integer would cause compiler warnings.
+#define YG_ENUM_BEGIN(name) NS_ENUM(unsigned, name)
 #define YG_ENUM_END(name)
 #else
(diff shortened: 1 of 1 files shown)


davidaurelio:
This causes a bunch of internal failures at FB. Gonna have to work through them one by one

wokalski:
@davidaurelio thank you for all of your hard work! I've seen that the android build was failing in the React Native repo. Is it related?

davidaurelio:
No, it’s some internal, non-OSS stuff. Unfortunately, we do have lots of different compiler setups, too :-D

davidaurelio:
This won’t land in this form, sorry. I am getting hard compiler errors. Will replace this with a portable bitfield implementation.

davidaurelio:
@Adlai-Holler @wokalski I have landed changing to enum storage today, e.g. https://github.com/facebook/yoga/blob/master/yoga/YGStyle.h#L76-L98

This *should* fix the problem.

PR #878, opened 2019-03-26
Fix CocoaPods integration with Xcode 10.2

PatrickNLT:
Swift 3.0 is not supported by Xcode 10.2 anymore, so we are bumping the Swift version while maximizing the number of Xcode versions that are able to build YogaKit.


PatrickNLT:
Fixed by d44a6d936fc6666fa509a0e4bdff209a8d22d381

PR #881, opened 2019-03-28
Upgrade CMakeLists.txt to use C++14

SidharthGuglani:


gengjiawen:
Failed with, looks like need to update NDK 
```
stderr: g++: error: unrecognized command line option ‘-std=c++14’
    When running <c++ preprocess_and_compile>.
    When building rule //lib/fb:fbjni#compile-pic-assert.cpp.odfda776a,default.
 FAILED
```

gengjiawen:
Any thought on using https://github.com/react-native-community/docker-android. It's used by `react-native`, and has buck and NDK 17c.

I can work on a PR for this.

janisozaur:
Why are you not using https://cmake.org/cmake/help/v3.1/prop_tgt/CXX_STANDARD.html ?

gengjiawen:
I have create the pr migrate to docker: https://github.com/facebook/yoga/pull/884.

shawwn:
Hello,

I submitted an alternative to this which uses CXX_STANDARD. It also only requires c++11 rather than c++14. 

https://github.com/facebook/yoga/pull/887

(I wasn't aware of this PR before submitting my own, but now that both exist, it seemed good to mention the other one.)

SidharthGuglani:
Created this PR to solve Travis CI build failure which is fixed by other commit [here](https://github.com/facebook/yoga/commit/c5f105a7b6879b1960820821edd74efd68eadbc0)

Closing this.

PR #884, opened 2019-04-07
use docker for android related build

gengjiawen:
## what is being fixed
Currently ci in master failed due to ndk version
## how is it fixed
use the android docker same as react-native (which include buck, ndk,=)
## why is docker being added.
see above.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: java
 os: linux
+services:
+  - docker
 addons:
   apt:
     sources:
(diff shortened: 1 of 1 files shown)


davidaurelio:
Thank you for your contribution. Would you mind filling in the description?

- what is being fixed 
- how is it fixed 
- why is docker being added.

gengjiawen:
> Thank you for your contribution. Would you mind filling in the description?
> 
> * what is being fixed
> * how is it fixed
> * why is docker being added.

I add them in PR description.

gengjiawen:
@davidaurelio ping.

davidaurelio:
Is this still relevant? It seems we got CI going again without docker.

gengjiawen:
> Is this still relevant? It seems we got CI going again without docker.

Yeap.
Travis CI android env has buck, ndk is old.
This docker is used by `react-native` repo too.

NickGerleman (changes requested):
Since this PR, infrastructure has moved to GitHub Actions. I'm going to take a stab at adding to it, but so far setup hasn't been a source of too much instability (though I'm sure this will change with more tasks). Regardless, this change would need to be updated for GitHub Actions.

Issue #728, opened 2018-03-10
Android Sample Crashes 


rorlig:
# Report

- [ ] I have searched [existing issues](https://github.com/facebook/yoga/issues) and this is not a duplicate

# Issues and Steps to Reproduce
run  buck install -r android/sample

# Expected Behavior
Sample app should run. 

# Actual Behavior
sample app crashes 

```
03-10 14:47:36.018 6947-6947/? E/AndroidRuntime: FATAL EXCEPTION: main
                                                 Process: com.facebook.samples.yoga, PID: 6947
                                                 java.lang.NoClassDefFoundError: Failed resolution of: Landroid/support/v4/os/BuildCompat;
                                                     at android.support.v7.app.AppCompatDelegate.create(AppCompatDelegate.java:196)
                                                     at android.support.v7.app.AppCompatDelegate.create(AppCompatDelegate.java:181)
                                                     at android.support.v7.app.AppCompatActivity.getDelegate(AppCompatActivity.java:521)
                                                     at android.support.v7.app.AppCompatActivity.onCreate(AppCompatActivity.java:71)
                                                     at com.facebook.samples.yoga.MainActivity.onCreate(MainActivity.java:33)
                                                     at android.app.Activity.performCreate(Activity.java:6662)
                                                     at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1118)
                                                     at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2599)
                                                     at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2707)
                                                     at android.app.ActivityThread.-wrap12(ActivityThread.java)
                                                     at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1460)
                                                     at android.os.Handler.dispatchMessage(Handler.java:102)
                                                     at android.os.Looper.loop(Looper.java:154)
                                                     at android.app.ActivityThread.main(ActivityThread.java:6077)
                                                     at java.lang.reflect.Method.invoke(Native Method)
                                                     at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:866)
                                                     at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:756)
[...]

PR #886, opened 2019-04-07
add android sample gradle support

gengjiawen:
Related issue: https://github.com/facebook/yoga/issues/728.


settings.gradle
```diff
@@ -6,6 +6,7 @@
  */
 
 include ':yoga', ':yogacore', ':yoga-layout', ':yoga:proguard-annotations', ':libfb'
+include ':android:sample'
```
Mishkun:
```suggestion
include ':yoga-layout:sample'
```
Or else it fails to sync with the message
```
ERROR: The modules ['android', 'yoga-layout'] point to the same directory in the file system.
Each module must have a unique path.
```
because when using relative project naming you implicitly declare an 'android' project

gengjiawen: I don't see the error neither on windows nor on macOS.

gengjiawen:
@davidaurelio Can you review this too ?

NickGerleman:
I think @cortinico took this on as part of recent Gradle work. https://github.com/facebook/yoga/blob/main/sample/build.gradle

PR #887, opened 2019-04-12
Fix cmake build / update CMakeLists.txt to use C++11

shawwn:
On MacOS, the following steps result in build errors:

```
mkdir build
cd build
cmake ..
make
```

The problem is that yogacore uses C++11 features (`constexpr`) but C++11 isn't specified in CMakeLists.txt.

This PR solves the poblem by adding the following code to the bottom of CMakeLists.txt:

```
set_target_properties(yogacore PROPERTIES CXX_STANDARD 11)
```

This solution was derived from https://stackoverflow.com/questions/45688522/how-to-enable-c17-in-cmake

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,3 +13,4 @@ file(GLOB yogacore_SRC yoga/*.cpp)
 add_library(yogacore STATIC ${yogacore_SRC})
 
 target_link_libraries(yogacore android log)
+set_target_properties(yogacore PROPERTIES CXX_STANDARD 11)


shawwn:
I notice yoga's javascript binding requires C++14. Would it make sense for me to change this PR to C++14 instead of C++11? 

I went with the minimum C++ version that seemed to work. But if C++14 would be more future-proof, we should go with that instead.

davidaurelio:
C++11 is fine for now, thanks for the fixes!

PR #909, opened 2019-06-26
Fix compilation with CMake

### LINKED ISSUE #908: Master doesn't build with CMake
Download master, `mkdir build && cd build && cmake -GNinja .. && ninja` gives:

```
In file included from ../yoga/Yoga.cpp:18:
../yoga/event/event.h:11:10: fatal error: 'yoga/YGEnums.h' file not found
#include <yoga/YGEnums.h>
         ^~~~~~~~~~~~~~~~
```

Timmmm:
This adds the root of the source tree to the include path, which allow `#include <yoga/YGEnums.h>` to work.

Fixes #908


davidaurelio:
Thank you for the fix!

PR #915, opened 2019-07-13
Properly test for the lack of a node measure func in YogaKit

nferruzzi:
The test was broken and caused a few crashes on my project.


davidaurelio:
Thanks for fixing!

PR #933, opened 2019-10-09
fix: some subview didn't enable yoga, and crash with NaN frame

rickytan:


bolee:
they haven't merge yet?

qxj006:
they haven't merge yet?

PR #948, opened 2019-11-10
Add print yoga node AllEdge value when debug mode

acton393:
add print yoga node AllEdge value when debug mode, the old logic just print four edgeValue including top left bottom and right

but when we set the same value to these edge, we use YGAllEdge, this will be more convenient debug code

so I open this pull request to add it


acton393:
image

yoga/YGNodePrint.cpp
```diff
@@ -90,7 +90,11 @@ static void appendEdges(
     const string& key,
     const YGStyle::Edges& edges) {
   if (areFourValuesEqual(edges)) {
-    appendNumberIfNotZero(base, key, edges[YGEdgeLeft]);
+    YGEdge edge = YGEdgeLeft;
+      if (edges[edge].isUndefined()) {
+          edge = YGEdgeAll;
+      }
+    appendNumberIfNotZero(base, key, edges[edge]);
```
NickGerleman, line 97: This could also be the combination of YGEdgeVertical or YGEdgeHorizontal. I think the robust way to do this would be to call one of the node edge resolution functions (`computeEdgeValueForColumn()` or `computeEdgeValueForRow()`)

NickGerleman:
Wow I didn't expect you to come back, 3 years later 😅. Thanks for the PR!

NickGerleman (approved):

cortinico (approved):

acton393:
> Wow I didn't expect you to come back, 3 years later 😅. Thanks for the PR!

😂😂 I am active recently, and receiving the code review message from my inbox

PR #960, opened 2020-01-11
Export YGInteropSetLogger method

KimDaeWook:
When building and using C # libraries,
EntryPointNotFoundException thrown from YGInteropSetLogger.

so, I added YOGA_EXPORT on YGInteropSetLogger.

### DIFF
--- a/yoga/Yoga.cpp
+++ b/yoga/Yoga.cpp
@@ -4363,6 +4363,11 @@ YOGA_EXPORT void YGConfigSetUseWebDefaults(
   config->useWebDefaults = enabled;
 }
 
+YOGA_EXPORT bool YGConfigGetUseLegacyStretchBehaviour(
+    const YGConfigRef config) {
+  return config->useLegacyStretchBehaviour;
+}
+
 YOGA_EXPORT void YGConfigSetUseLegacyStretchBehaviour(
     const YGConfigRef config,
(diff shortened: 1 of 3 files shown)


KimDaeWook:
And YGConfigGetUseLegacyStretchBehaviour is not defined, so i add it additionally.

yoga/Yoga.cpp
```diff
@@ -4363,6 +4363,11 @@ YOGA_EXPORT void YGConfigSetUseWebDefaults(
   config->useWebDefaults = enabled;
 }
 
+YOGA_EXPORT bool YGConfigGetUseLegacyStretchBehaviour(
```
SidharthGuglani: Do you really need this ? This flag might be removed in future.

SidharthGuglani:
Thanks for creating this PR.
Will merge the PR for adding YOGA_EXPORT to YGInteropSetLogger after you remove other changes

NickGerleman:
It looks like the C# bindings already assume `YGConfigGetUseLegacyStretchBehaviour`. From recent discussions on RN compatibility, I don't think we will be able to remove the flag soon, so adding a public accessor seems righteous to me.

Otherwise, fixing the mismatch in export declaration is a good change.

NickGerleman (approved):

PR #965, opened 2020-01-30
feature: support swift package manager

DingSoung:
https://github.com/DingSoung/FlexLayout

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .DS_STORE
 
+*.xcodeproj
 
 /buck-cache/
 /buck-out/
--- a/.gitignore
+++ b/.gitignore
@@ -1,11 +1,14 @@
 .DS_STORE
 
+
 /buck-cache/
 /buck-out/
 /.buckconfig.local
 /.buckd
 /gentest/test.html
 .buckversion
+
+.build/
 
 # Jekyll
 /.sass-cache/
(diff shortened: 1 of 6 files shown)


sjmueller:
Hi @DingSoung thanks for putting this together. Some features of YogaKit are incomplete in the package, for instance there is no ability to use the `UIView.yoga` property, which is critical to use `applyLayout` and the `configureLayout` block. My guess is because the headers and classes within `YogaKit/Sources` is not included, since `UIView+Yoga.h` is needed.

I tried reconfiguring your `Package.swift` to support multiple source directories, but ran into an error about mixed language source files. SPM might have trouble with c++ objc and swift all together, not sure.

Can you take a stab at including the rest of YogaKit in the package?

DingSoung:
我的想法是分离layout(yoga)和UI(UIKit)

好处1:  前者运行在background thread，后者运行在main thread. main thread 忙的时候，可以减轻它负担。
好处2:  在复杂的页面，可以提前计算好布局，然后构造UI，同时设置位置。

今后我们可以这么写native

```swift
let component = View.width(100%).height(80%).backgroundColor(.white).flex(1).justifyContent(.center).subViews(
   Label.text("hello world").fontSize(20).color(.black),
   Image.source("http://xxxx.xxx/src/img.png")
)
```

DingSoung:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

PR #995, opened 2020-04-06
Add compiler option to call measure callback on all nodes

sagokhal:
When using Yoga in combination with external layout systems (brown field scenarios) Yoga leaf nodes can be subtrees w.r.t. the external layout system.
During the measure phase, these subtrees need to be measured by the external layout system - even if the Yoga node's mode is YGMeasureModeExactly.
This change adds a preprocessor conditional that can be set by out-of-tree platforms to always call measure for all nodes that have the measure callback.

### DIFF
--- a/yoga/Yoga.cpp
+++ b/yoga/Yoga.cpp
@@ -1695,6 +1695,15 @@ static void YGNodeWithMeasureFuncSetMeasuredDimensions(
       : YGFloatMax(
             0, availableHeight - marginAxisColumn - paddingAndBorderAxisColumn);
 
+#ifdef YOGA_CALL_MEASURE_CALLBACK_ON_ALL_NODES
+  const YGSize measuredSize = node->measure(
+      innerWidth,
+      widthMeasureMode,
+      innerHeight,
+      heightMeasureMode,
+      layoutContext);
+#endif
+
   if (widthMeasureMode == YGMeasureModeExactly &&
       heightMeasureMode == YGMeasureModeExactly) {
     // Don't bother sizing the text if both dimensions are already defined.
@@ -1717,6 +1726,7 @@ static void YGNodeWithMeasureFuncSetMeasuredDimensions(
   } else {
     Event::publish<Event::MeasureCallbackStart>(node);
 
+#ifndef YOGA_CALL_MEASURE_CALLBACK_ON_ALL_NODES
     // Measure the text under the current constraints.
     const YGSize measuredSize = node->measure(
         innerWidth,
@@ -1724,6 +1734,7 @@ static void YGNodeWithMeasureFuncSetMeasuredDimensions(
         innerHeight,
         heightMeasureMode,
         layoutContext);
+#endif
 
     layoutMarkerData.measureCallbacks += 1;
     layoutMarkerData.measureCallbackReasonsCount[static_cast<size_t>(reason)] +=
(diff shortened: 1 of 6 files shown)


yoga/Yoga.cpp
```diff
@@ -1695,6 +1695,15 @@ static void YGNodeWithMeasureFuncSetMeasuredDimensions(
...
+#ifdef YOGA_CALL_MEASURE_CALLBACK_ON_ALL_NODES
+  const YGSize measuredSize = node->measure(
+      innerWidth,
+      widthMeasureMode,
+      innerHeight,
+      heightMeasureMode,
+      layoutContext);
+#endif
+
   if (widthMeasureMode == YGMeasureModeExactly &&
```
SidharthGuglani, line 1707: This is still getting executed even if measure callback is called on all the nodes. 
I think you want to use measuredSize after measure callback

yoga/Yoga.cpp
```diff
@@ -1695,6 +1695,15 @@ static void YGNodeWithMeasureFuncSetMeasuredDimensions(
       : YGFloatMax(
             0, availableHeight - marginAxisColumn - paddingAndBorderAxisColumn);
 
+#ifdef YOGA_CALL_MEASURE_CALLBACK_ON_ALL_NODES
```
SidharthGuglani, line 1698: Instead of compiler option, it will be better to pass a flag via YGConfig

NickGerleman:
@SidharthGuglani sorry for the delay here. Are there any remaining issues?

sagokhal:
Hey @SidharthGuglani, just wanted to follow-up on this. Are they any more changes you recommend to get this in?

woehrl01:
Hi all, I just stumbled over it based on https://github.com/microsoft/react-native-windows/issues/3994. I would add two additional suggestions. 
1. Instead of adding a new parameter for the `YGConfigRef` you can use `node->getConfig()` from the passed `node`. 
2. Instead of adding a new `bool` field and having to extend the API (would should be als added at best all implemented languages) + breaking the ABI, you can add an extend the `YGExperimentalFeature*` enum in `YGEnum.h` and use the existing `YGConfigSetExperimentalFeatureEnabled`. At least we added that feature a few years ago exactly for such things.

sagokhal:
@woehrl01 Both those points makes sense. I'll make those changes.

yoga/Yoga.cpp
```diff
@@ -1717,7 +1719,6 @@ static void YGNodeWithMeasureFuncSetMeasuredDimensions(
   } else {
     Event::publish<Event::MeasureCallbackStart>(node);
 
-    // Measure the text under the current constraints.
```
woehrl01, line 1720: I think the removing of that comment was unintentional and should be reverted.

NickGerleman:
@acoates-ms @rozele @sagokhal do any of you have thoughts on the desired outcome of this PR?

rozele:
When last discussed with @acoates-ms, we agreed that this should not be the default behavior for react-native-windows, but it sounds like it's still needed for react-native-win32, so having an experimental flag and a corresponding QuirkSetting in react-native-windows is probably a good idea.

NickGerleman (changes requested):
A couple feedback items to merge:
1. YGEnums is generated code. We should edit `enums.py` and regenerate the other language bindings.
2. Could we rename `YGExperimentalFeatureCallMeasureCallbackOnAllNodes` to `YGExperimentalFeatureMeasureCallbackOnMeasureModeExactly`?

PR #1004, opened 2020-05-06
Migrate CI from TravisCI to GitHub Actions

bigfootjon:
Migrate from TravisCI to GitHub Actions. Facebook employees can read about this migration [here](https://fburl.com/rfp033ou).

CC: @wittgenst

Test Plan:
GitHub Actions CI


.github/workflows/ci.yml
```diff
@@ -0,0 +1,55 @@
...
+        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_branch: gh-pages
+          publish_dir: website/public
+          cname: yogalayout.com
+          keep_files: true
+          user_name: 'Yoga-bot'
+          user_email: 'yogabot@fb.com'
```
zpao, line 32: We should come up with some best practice here… I almost prefer website deployment being a completely different workflow that only runs on master (so it's obvious at the workflow configuration when it runs, not needing to read this far to find out)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,55 @@
+name: CI
+
+on: [push, pull_request]
+
+jobs:
+  website:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
```
zpao, line 9: There's no matrix so this is a no-op I believe

.github/workflows/ci.yml
```diff
@@ -0,0 +1,55 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js 8
+        uses: actions/setup-node@v1
+        with:
+          node-version: 8.x
```
zpao, line 15: AHHHH Node 8 EOL was last year. Let's at least use 10 (or 12)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,55 @@
...
+        working-directory: website
+      - name: Deploy
+        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_branch: gh-pages
+          publish_dir: website/public
+          cname: yogalayout.com
+          keep_files: true
```
zpao, line 30: Why?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,55 @@
...
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install dependencies
+        run: |
+          pushd $HOME
+          git clone --depth 1 https://github.com/facebook/buck.git
+          cd buck
+          ant
```
zpao, line 44: Oh geez this is building Buck on every change… that's… wow. We probably actually need to do something different here since Buck master is diverging (but let's ignore that for now)

.github/workflows/ci.yml
bigfootjon: I just copied the Travis job ¯\_(ツ)_/¯

.github/workflows/ci.yml
bigfootjon: I was trying to avoid writing it out twice. I'd have to copy the entire job and any changes to the build process would need to be done in two places. It's not possible to have one yml file depend on another :(

.github/workflows/ci.yml
bigfootjon: It was in `.travis.yml`: `skip-cleanup: true`

PR #1013, opened 2020-06-09
Fix ContentAlign and AlignItems combination cause wrong behaviour

### LINKED ISSUE #1008: ContentAlign and AlignItems combination cause wrong behaviour
# Report

- [ X] I have searched [existing issues](https://github.com/facebook/yoga/issues) and this is not a duplicate

Configure a Flexbox in this way

```javascript
     <View style={{width: 300, height: 300, backgroundColor: 'red', flexDirection: 'row',flexWrap:'wrap', alignContent:'space-around',alignItems:'flex-end' }}>
        <View style={{width: 150, height: 50, backgroundColor: 'powderblue'}} />
        <View style={{width: 120, height: 100, backgroundColor: 'skyblue'}} />
        <View style={{width: 120, height: 50, backgroundColor: 'steelblue'}} />
      </View>
  
```

# Expected Behavior
The elements must be show like this: (This is the web representation)
image

# Actual Behavior
image
the elements are wrong located, the combination of contentAlign and AlignItem causes strange behaviours.
Whit combination like
ContentAlign: space-around or space-between with
AlignItems: center or flex-end 
The algorithm locate the elements in awrong way.

# Link to Code
https://stackblitz.com/edit/typescript-yoga-rg4mcr?file=index.ts

woehrl01:
This PR fixes  #1008, where the combination of ContentAlign and AlignItems cause wrong behavior. 

Reason for that was that the cross dimensional lead has been mistakenly uses in the calculation of the item position during the AlignItems step (for non-stretch values).

Added the specific tests to prevent future regressions.


NickGerleman:
I incorporated a fix for this into https://github.com/facebook/yoga/commit/464d1668baeb2e448c12f3cc95990924972c2532

PR #1014, opened 2020-06-09
Fix margin auto with smaller parent lead to negative position

### LINKED ISSUE #978: Margin auto allows negative values
# Report

Create a setup like follows

```js
<View style={{ height: 500 }}>
  <View style={{ flex: 0, height: 300 }} />
  <View style={{ flex: 0, marginTop: 'auto', height: 300 }} />
</View>
```

[Or use playground link](https://yogalayout.com/playground?eyJ3aWR0aCI6NTAwLCJoZWlnaHQiOjUwMCwibWluV2lkdGgiOm51bGwsIm1pbkhlaWdodCI6bnVsbCwibWF4V2lkdGgiOm51bGwsIm1heEhlaWdodCI6bnVsbCwiZmxleERpcmVjdGlvbiI6MCwicG9zaXRpb24iOnsidG9wIjpudWxsLCJyaWdodCI6bnVsbCwiYm90dG9tIjpudWxsLCJsZWZ0IjpudWxsfSwiY2hpbGRyZW4iOlt7IndpZHRoIjoiIiwiaGVpZ2h0IjoiMzAwIiwibWluV2lkdGgiOm51bGwsIm1pbkhlaWdodCI6bnVsbCwibWF4V2lkdGgiOm51bGwsIm1heEhlaWdodCI6bnVsbCwicG9zaXRpb24iOnsidG9wIjpudWxsLCJyaWdodCI6bnVsbCwiYm90dG9tIjpudWxsLCJsZWZ0IjpudWxsfSwiZmxleFNocmluayI6IjAifSx7IndpZHRoIjoiIiwiaGVpZ2h0IjoiMzAwIiwibWluV2lkdGgiOm51bGwsIm1pbkhlaWdodCI6bnVsbCwibWF4V2lkdGgiOm51bGwsIm1heEhlaWdodCI6bnVsbCwibWFyZ2luIjp7InRvcCI6ImF1dG8ifSwicG9zaXRpb24iOnsidG9wIjpudWxsLCJyaWdodCI6bnVsbCwiYm90dG9tIjpudWxsLCJsZWZ0IjpudWxsfSwiZmxleFNocmluayI6IjAifV19)

# Expected Behavior

The bottom view should overflow the parent (as happens on the web)

# Actual Behavior

The bottom view gets a negative top margin applied, and overlaps the top view. Nothing overflows the parent

# Link to Code

https://yogalayout.com/playground?eyJ3aWR0aCI6NTAwLCJoZWlnaHQiOjUwMCwibWluV2lkdGgiOm51bGwsIm1pbkhlaWdodCI6bnVsbCwibWF4V2lkdGgiOm51bGwsIm1heEhlaWdodCI6bnVsbCwiZmxleERpcmVjdGlvbiI6MCwicG9zaXRpb24iOnsidG9wIjpudWxsLCJyaWdodCI6bnVsbCwiYm90dG9tIjpudWxsLCJsZWZ0IjpudWxsfSwiY2hpbGRyZW4iOlt7IndpZHRoIjoiIiwiaGVpZ2h0IjoiMzAwIiwibWluV2lkdGgiOm51bGwsIm1pbkhlaWdodCI6bnVsbCwibWF4V2lkdGgiOm51bGwsIm1heEhlaWdodCI6bnVsbCwicG9zaXRpb24iOnsidG9wIjpudWxsLCJyaWdodCI6bnVsbCwiYm90dG9tIjpudWxsLCJsZWZ0IjpudWxsfSwiZmxleFNocmluayI6IjAifSx7IndpZHRoIjoiIiwiaGVpZ2h0IjoiMzAwIiwibWluV2lkdGgiOm51bGwsIm1pbkhlaWdodCI6bnVsbCwibWF4V2lkdGgiOm51bGwsIm1heEhlaWdodCI6bnVsbCwibWFyZ2luIjp7InRvcCI6ImF1dG8ifSwicG9zaXRpb24iOnsidG9wIjpudWxsLCJyaWdodCI6bnVsbCwiYm90dG9tIjpudWxsLCJsZWZ0IjpudWxsfSwiZmxleFNocmluayI6IjAifV19

# Notes

If you remove `marginTop: auto` you get the correct behaviour

I did find [these few lines](https://github.com/facebook/yoga/blob/master/yoga/Yoga.cpp#L2601-L2605)

I'm not sure what the logic of `remainingFreeSpace` is - if it should allow negative values or not. If it does allow them, it might just need a `max(0, ...)`

woehrl01:
This PR fixes #978, where a smaller parent in combination with `margin: auto` on bigger children lead to a negative position.

Include tests to prevent future regressions.


woehrl01:
@SidharthGuglani Which concerns from your side are in place to having this merged?

NickGerleman:
@jacobp100 I thought about this one more.

If the surfaces are wanting to use as a test for "big enough to be representative" already had margin: "auto" broken without someone noticing, we should probably just make this fix.

jacobp100:
Yeah, I’d highly doubt anyone was relying on the incorrect behaviour. Would be really good to get this in - I’ve personally not used auto margins in RN since I hit this bug maybe 3 years ago?

NickGerleman:
> Yeah, I’d highly doubt anyone was relying on the incorrect behaviour. Would be really good to get this in - I’ve personally not used auto margins in RN since I hit this bug maybe 3 years ago?

I finally got around to running the production A/B test on the Facebook app, for the fix for “auto” flexBasis, margin, etc no-oping in Fabric.

Turns out in that example, there were tens of millions of users who hit the code to parse the props that no-op, along with code relying on it which caused those users to see a 1% drop in an important e-commerce metric for us 🙃.

So… I think I’m going to go back to, as a rule, putting any behavior change behind an errata before we can prove it is not relied on in real world usage. Given enough code, it’s surprising what ends up showing up.

NickGerleman:
FYI @nicoburns, I know you had asked before what I meant when I said “run an experiment”, where I think you had interpreted it as running a lot of tests. This is an example of what I meant, where we turn something on for a percentage of users, and can measure how often a condition is hit and whether user behavior changes.

nicoburns:
> FYI @nicoburns, I know you had asked before what I meant when I said “run an experiment”, where I think you had interpreted it as running a lot of tests. This is an example of what I meant, where we turn something on for a percentage of users in very large apps with many users, and can measure how often a condition is hit and whether user behavior changes.

Hmm... interesting. I think that does give me a better idea of what "run an experiment" means. I think I had imagined these apps having relatively comprehensive e2e tests, and "run an experiment" would consist of running those rather than enabling the code for production users. I wonder if it would be possible to use a similar mechanism to surface future compatibility warnings to the owners of the affected apps, such that the "errata" could be removed after a certain grace period has elapsed?

(although I would still personally take a "freeze and fork" approach to developing a new version of the algorithm which would avoid this problem entirely as existing conservative users can just stick with the frozen version, and after an initial "experimental" period while it catches up with web implementations, the forked version can piggyback on browsers' web compatibility testing).

NickGerleman:
> mechanism to surface future compatibility warnings to the owners of the affected apps, such that the "errata" could be removed after a certain grace period has elapsed?

That’s the rough plan. I have many more details in https://github.com/facebook/yoga/issues/1247

PR #1017, opened 2020-06-11
fix missing invalidation if owner has undefinied dims

### LINKED ISSUE #1003: Incorrect cached child size is used because YGNodeCanUseCachedMeasurement() does not respect min/max dimensions
# Report

See the attached example.

Child node has height = 10, minHeight = 40%, maxHeight = 60%.

When parent is measured against height = NaN, child height is computed to be 10, which is correct.

But then parent is laid out against height = 30.
Now child height should be `max(40% * 30, min(10, 60% * 30)) = max(12, min(10, 18)) = 12`.
But `YGNodeCanUseCachedMeasurement()` returns `true` and cached height of 10 is used instead.

# Expected Behavior
In all the cases, final layout should have child height of 12.

# Actual Behavior
In some of the scenarios incorrect cached height is used.

# Link to Code
https://pastebin.com/x5e6ZL2b

woehrl01:
Fixes #1003 by adding additional values, which keep track if the owner size has been undefined before. Being undefined lead to different calculations for min/max constraints thus, in that specific case leading to the wrong behavior.

### DIFF
--- a/tests/YGRelayoutTest.cpp
+++ b/tests/YGRelayoutTest.cpp
@@ -49,3 +49,33 @@ TEST(YogaTest, recalculate_resolvedDimonsion_onchange) {
 
   YGNodeFreeRecursive(root);
 }
+
+
+YGSize _measureRecalc(YGNodeRef node, float width, YGMeasureMode widthMode, float height, YGMeasureMode heightMode) {
+    return YGSize { 0, 0};
+}
+
+TEST(YogaTest, recalculate_on_layout_values_change) {
+  const YGConfigRef config = YGConfigNew();
+  YGConfigSetExperimentalFeatureEnabled(config, YGExperimentalFeatureWebFlexBasis, true);
+  YGConfigSetPointScaleFactor(config, 3.f);
+  const YGNodeRef root = YGNodeNewWithConfig(config);
+  YGNodeStyleSetFlexDirection(root, YGFlexDirectionRow);
+  YGNodeStyleSetAlignItems(root, YGAlignFlexStart);
+  YGNodeStyleSetAlignContent(root, YGAlignFlexStart);
+  const YGNodeRef child = YGNodeNewWithConfig(config);
+  YGNodeStyleSetMinHeightPercent(child, 40.f);
+  YGNodeStyleSetMaxHeightPercent(child, 60.f);
+  YGNodeStyleSetHeight(child, 10.f);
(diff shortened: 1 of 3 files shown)


nickolas-pohilets:
Would similar issue also happen when owner size changes from one defined size to another?

woehrl01:
It should also work in that case. As the missing parent size leads to not taking min/max values into account. I'll have a second look later, maybe we have a different fix, without the new variables here, too. Until then I'll keep this PR as draft. The alternative as described in the issue is still the best approach for now.

NickGerleman:
Hey @woehrl01, I think we want to import this, and think it might supersede the current code for `YGUseExperimentalWebFlexBasis`. @rozele root-caused a long-running issue in an app we care about to this issue #1161 🙂.

We're more generally looking to start migrating the React Native ecosystem off of legacy incorrect behaviors, but I think we should put this behind a flag. That way we can experiment, to know if it will be a breaking change we should leave behind a gate, or if its relatively safe to enable out of the box.

Size impact per-node can be mitigated by embedding into the `flags` enum bitset I think. We may need to add another byte to it soon. There is also a flag I have been meaning to remove from there (for super expensive whole-tree layout diffing).

woehrl01:
@NickGerleman it's awesome to see the latest effort you put into the library. 

I fully agree on your approach. Feel free to pick this up!

woehrl01:
@NickGerleman I just had a closer look into the current layout of the `YGLayout` struct. I think you can put those two additional flags also into the [following position by replacing `lastOwnerDirection`]( https://github.com/facebook/yoga/blob/5dd33acc912711c599250e3533b07c732de503ee/yoga/YGLayout.h#L42) with a flag, similar to `didUseLegacyFlagOffset`. This would be even nicer as you have one field for flags and the other field for storing latest layout conditions. This won't change the per node size.

jacobp100:
@woehrl01 does this bug also happen if `padding` or `borderWidth` were set to undefined - or is it _just_ width and height?

woehrl01:
@jacobp100 Are you talking about `padding` or `borderWidth` on the owner? This fix is mainly if the root owner changes, where you usually can't set those properties at all. If you modify `padding` on a child item, the `dirty` flag is set and gets propagated. Do you have a specific issue/test case?

jacobp100:
No - I was just wondering. It just seemed strange to me that only width/height would trigger this, but not other layout-affecting properties

PR #1025, opened 2020-08-10
Fix broken absolute/relative link in Playground Editor

joshuayuan:
The correct link is https://yogalayout.com/docs/absolute-relative-layout
The current link leads to a broken page https://yogalayout.com/docs/absolute-position


joshuayuan:
Reopening to trigger checks again. I think the failure was a transient issue

Issue #921, opened 2019-08-26
Add manifest for SwiftPM support


tcurdt:
# Report

- [x] I have searched [existing issues](https://github.com/facebook/yoga/issues) and this is not a duplicate

# Issues and Steps to Reproduce

Package.swift
```
// swift-tools-version:5.0
import PackageDescription

let package = Package(
    name: "Test",
    products: [],
    dependencies: [
        .package(url: "https://github.com/facebook/yoga.git", .branch("master")) // .upToNextMajor(from: "1.14.0")),
    ],
    targets: [
        .target(
            name: "Test",
            dependencies: [
                "yoga",
            ],
            path: "Test"
        ),
    ]
)
```

```
$ swift package resolve
Fetching https://github.com/facebook/yoga.git
https://github.com/facebook/yoga.git @ master: error: https://github.com/facebook/yoga.git has no manifest
```

# Expected Behavior

It should install.

# Actual Behavior

SwiftPM complains that there is no manifest.
It would be great if a manifest could be added.

sjmueller:
Is this planned any time soon?

NickGerleman:
Added with https://github.com/facebook/yoga/pull/1339

PR #1032, opened 2020-10-09
Add support for Swift Package Manager

salling:
This PR addresses #965

Objective was to add support for swiftpm, while at the same time keeping the change set minimal. The "trick" here is to add a public header file that is used from Package.swift.

Note: this PR does not intend to expose YogaKit through SPM, only yoga itself.


yoga/swiftpm-support/umbrella.h
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#pragma once
+
+#ifdef __OBJC__
```
clausjoergensen, line 10: Is it possible to use the [SWIFT_PACKAGE](https://github.com/apple/swift-package-manager/blob/main/Documentation/Usage.md#packaging-legacy-code) preprocessor flag here to make it even safer?

clausjoergensen:
Ping #921

yoga/swiftpm-support/umbrella.h
salling: Does not seem to be defined. At least not in the context where this code is compiling.

NickGerleman:

clausjoergensen:
@NickGerleman this wasn't for YogaKit though. You keep updating the CocoaPods package, why can't you make a SPM package too?

Package.swift
```diff
@@ -0,0 +1,35 @@
...
+            publicHeadersPath: "./swiftpm-support",
+            cxxSettings: [
+                .headerSearchPath("../"),
+            ],
+            linkerSettings: [
+                .linkedFramework("Foundation"),
+            ]
+        )
+    ],
+    cxxLanguageStandard: .cxx14
```
cipolleschi, line 34: this should probably be `.cxx17`, as this is the version we are typically using internally.

Package.swift
```diff
@@ -0,0 +1,35 @@
+// swift-tools-version:5.3
```
cipolleschi, line 1: we can probably use a more recent version of swift-tools here. At least `5.8`, until Xcode 15 become stable.

cipolleschi:
Hi there, thanks @NickGerleman for tagging me.
I'm highly supportive of this change. Yoga is one of the few React Native packages that we can use independently and it makes a lot of sense to me to offer also a Swift PM alternative.

Also, the setup is fairly simple and I really like the folder with the public umbrella header.

I have not spent a lot of time thinking about how to validate it, though. One simple way could be to have an example app that import the dependency as a package and exercise it.
Another way could be to automatically generate the Package.swift from the podspec.

I don't have a strong opinion, although the second could be better in the long run. 
I'll leave the decision to you.

cipolleschi (commented):
I'm very happy about this happening. It is a first step toward embracing more modern languages. I left a couple of comments to update a couple of values, but I'll leave you the last word on those.

Package.swift
NickGerleman: Yoga is targeting C++14 for a little bit longer.

Package.swift
NickGerleman: We should maybe consider bumping this in CocoaPods as well then, which is currently set on 5.1

iainsmith:
This can be closed now we have https://github.com/facebook/yoga/commit/ef5784aca647a68343f5409e379519827edb86a9was

PR #1055, opened 2021-01-08
Fuzzer

shouc:
This PR adds a fuzzer that achieves high coverage for Yoga C++ lib. It is developed based on libprotobuf-mutator and libfuzzer. With this, Yoga could be also added to Google's oss-fuzz for continuous fuzzing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,12 @@ You may need to install the latest watir-webdriver gem (`gem install watir-webdr
 ### .NET
 .NET testing is not integrated in buck yet, you might need to set up .NET testing environment. We have a script which to launch C# test on macOS, `csharp/tests/Facebook.Yoga/test_macos.sh`.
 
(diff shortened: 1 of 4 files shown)


TaTaRsKiY059:
Update

shouc:
@loe @dreiss could you please review this?

NickGerleman (changes requested):
Thanks for this PR. Generally I don't think these sorts of usages of Yoga have aged will as part of the Yoga repo, when they are not actively run. This change adds new sources of truth for items like method names, enums. Some of these, have solutions to keep in sync (like the enum generator), but others seem more tricky. We would also want some sort of CI validation that this script continues to work, if we were to host it.

Let me know if you are still interested in this contribution. I'd love to know more about some of the overall motivation for Yoga as a fuzzer target.

PR #1075, opened 2021-04-01
added new baseline enums to yoga

edwardfoux:
Add the new baseline enums to yoga align.
Currently Yoga only has a BASELINE enums. New baseline enums are used in web:
BASELINE_LAST
BASELINE_FIRST
https://developer.mozilla.org/en-US/docs/Web/CSS/align-items#support_in_flex_layout
Given that Litho has a custom baseline function, developers can take advantage of the new baseline types in the layout calculation.

https://drive.google.com/file/d/1PQt4GLJoMoaD5AaUZ818KzpOscP16msn/view?usp=sharing
https://drive.google.com/file/d/1Bp7BhdJqneERAUm8gwxXvdcOyQ9u7AvJ/view?usp=sharing

### DIFF
--- a/java/com/facebook/yoga/YogaAlign.java
+++ b/java/com/facebook/yoga/YogaAlign.java
@@ -15,7 +15,9 @@ public enum YogaAlign {
   STRETCH(4),
   BASELINE(5),
   SPACE_BETWEEN(6),
-  SPACE_AROUND(7);
+  SPACE_AROUND(7),
+  BASELINE_FIRST(8),
+  BASELINE_LAST(9);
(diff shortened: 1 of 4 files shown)


NickGerleman (changes requested):
@edwardfoux do you still have any plans with this PR, to use these enums?

Yoga enums are also generated, from `enums.py`. Until recently it didn't leave an indication the files were generated, so there has been a recent history of folks manually editing them, which has led to some trouble. Ee should avoid editing files like `YGEnums.h` directly going forward.

Issue #671, opened 2017-11-21
Android YogaLayout set setVisibility to View.GONE problem


AndrewYangmq:
# Report
When I set a YoagLayout to View.GONE. It still takes the space and give it back to its parent. I already try invalidate(view), requestLayout on YoagLayout and its parent YoagLayout.

- [x ] I have searched [existing issues](https://github.com/facebook/yoga/issues) and this is not a duplicate

# Issues and Steps to Reproduce
Please see the xml and code below. Just click the change button and it will show or hide the content of YogaLayout1. But when it hides it, the space is still reserved and not given back to its parent. This never happens when I used Android LinearLayout as the container.

The only way to make the YogaLayout give back its space is to remove it from its parent YogaLayout. But when I add it back when I want to show it, it is not displayed properly. Please see the screen shots below.

# Expected Behavior
When we  set yogalayout Visibility to View.GONE, it should not take its space any more.

# Actual Behavior
The yogalayout still takes the space.
Following pictures are in order of: 
1-initial screen
2-after setVisibility to View.GONE
3-use removeview approach
4- after removing the yogalayout and add it back in its parent yogalayout at index 0.
![1](https://user-images.githubusercontent.com/33296921/33080681-af367456-cea6-11e7-94a8-dc0b44fb91bc.png)
![2](https://user-images.githubusercontent.com/33296921/33080682-af4ac6e0-cea6-11e7-8144-f31764bda14d.png)
![3](https://user-images.githubusercontent.com/33296921/33080679-af17b2b4-cea6-11e7-84ee-fa2119257a30.png)
![4](https://user-images.githubusercontent.com/33296921/33080680-af2957ee-cea6-11e7-87ea-e95db7f31505.png)

# Link to Code
https://gist.github.com/AndrewYangmq/4be91e3031863a487448bb3ca02d620d

uziasferreirazup:
hey @AndrewYangmq  did you find some solution about this?

290125120:
I alse meet this problem, is anyone find some solution？

hou3172568:
in `YogaLayout.ViewMeasureFunction#measure`，change `if (view == null || view instanceof YogaLayout) { 
        return YogaMeasureOutput.make(0, 0);
      }` to ` if (view == null || view instanceof YogaLayout || view.getVisibility() == GONE){.....}`

hou3172568:
hello,I'm back。
I found the way to fix this bug. 
In

 ```java
[...]

PR #1092, opened 2021-05-31
support child view visible and gone

hou3172568:
find the way to support child view's visible and gone.(issues https://github.com/facebook/yoga/issues/671)


NickGerleman:
We are deprecating the Yoga ViewGroup as part of the Yoga 2.0 release. We are still going to release a new revision based on the current state of the repo, but won't be accepting new contributions, since we are going to be removing it from the repo after.

PR #1095, opened 2021-07-04
[yoga] Replace float with double

cntrump:


NickGerleman (changes requested):
Thanks for the contribution!

I don't think this change could be accepted in its current form, since it would break the existing API of Yoga. If there are cases where we discovered that we do need this extra precision, it might make sense to hide this behind a separate configurable type.

I think it would also be wise to perform some performance/memory analysis, to see if that extra precision comes at a cost.

PR #1096, opened 2021-07-05
[yoga] Support SPM (Swift Package Manager)

cntrump:


clausjoergensen:
Also see #1032

cntrump:
done.

NickGerleman:

PR #1105, opened 2021-09-12
Update iOS Sample, migrate to Swift5.

dirtmelon:


NickGerleman:
I added https://github.com/facebook/yoga/commit/4fc1657f2c40655373d93cf439b784fa2eae1779 which updated us to Swift 5.1, and used some of the changes here (esp around IGListKit update). Thanks for the contribution!

PR #1106, opened 2021-09-18
Fix condition that is always false due to a typo

Hexlord:
If a float is undefined, its comparison with zero does not produce meaningful result.
The intended behavior is to check for zero only if the value is NOT undefined, because the comment states the scenario of "no children" which matches with the value being equal to zero, and that contradicts "&& IsUndefined" part of the condition.

### DIFF
--- a/yoga/Yoga.cpp
+++ b/yoga/Yoga.cpp
@@ -2952,10 +2952,10 @@ static void YGNodelayoutImpl(
         availableInnerMainDim = maxInnerMainDim;
       } else {
         if (!node->getConfig()->useLegacyStretchBehaviour &&
-            ((YGFloatIsUndefined(
+            ((!YGFloatIsUndefined(
                   collectedFlexItemsValues.totalFlexGrowFactors) &&
               collectedFlexItemsValues.totalFlexGrowFactors == 0) ||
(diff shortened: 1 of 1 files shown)


Hexlord:
> Hi @Hexlord!
> 
> Thank you for your pull request and welcome to our community.
> 
> # Action Required
> In order to merge **any pull request** (code, docs, etc.), we **require** contributors to sign our **Contributor License Agreement**, and we don't seem to have one on file for you.
> 
> # Process
> In order for us to review and merge your suggested changes, please sign at https://code.facebook.com/cla. **If you are contributing on behalf of someone else (eg your employer)**, the individual CLA may not be sufficient and your employer may need to sign the corporate CLA.
> 
> Once the CLA is signed, our tooling will perform checks and validations. Afterwards, the **pull request will be tagged** with `CLA signed`. The tagging process may take up to 1 hour after signing. Please give it that time before contacting us about it.
> 
> If you have received this in error or have any questions, please contact us at [cla@fb.com](mailto:cla@fb.com?subject=CLA%20for%20facebook%2Fyoga%20%231106). Thanks!

I don't care about this particular PR being merged, so I won't sign anything, I am just sharing my note on the code.

PR #1114, opened 2021-12-17
Fix measure inner dimensions

kinarobin:
`YGFloatIsUndefined(collectedFlexItemsValues.totalFlexGrowFactors) &&            collectedFlexItemsValues.totalFlexGrowFactors == 0`  is not reachable.


yoga/Yoga.cpp
```diff
@@ -2953,9 +2953,9 @@ static void YGNodelayoutImpl(
       } else {
         if (!node->getConfig()->useLegacyStretchBehaviour &&
             ((YGFloatIsUndefined(
-                  collectedFlexItemsValues.totalFlexGrowFactors) &&
+                  collectedFlexItemsValues.totalFlexGrowFactors) ||
               collectedFlexItemsValues.totalFlexGrowFactors == 0) ||
-             (YGFloatIsUndefined(node->resolveFlexGrow()) &&
+             (YGFloatIsUndefined(node->resolveFlexGrow()) ||
```
rshest, line 2958: Guess those extra parentheses are redundant, now that it's a list of OR clauses... granted, the expression is quite hard to unpack visually, and extra parentheses don't help here, quite an opposite

rshest: I actually believe that the initial intent there may have been to have it like this instead:
```
if (!node->getConfig()->useLegacyStretchBehaviour &&
            ((!YGFloatIsUndefined(
                  collectedFlexItemsValues.totalFlexGrowFactors) &&
              collectedFlexItemsValues.totalFlexGrowFactors == 0) ||
             (!YGFloatIsUndefined(node->resolveFlexGrow()) &&
              node->resolveFlexGrow() == 0))) {
```

rshest (changes requested):
Please check my comment, the code was wrong there, but I believe the intent was to check if *not* undefined and then compare with zero for both values.

yoga/Yoga.cpp
kinarobin: Yeah, you're right. I had update this code.

PR #1123, opened 2022-01-31
Add memory allocation API with templated C++ delegates & minor fixups

KitsuneAlex:
This simple pull request adds a memory allocation API to the Yoga C API,
as well as the respective templated C++ delegate functions to allow for the use
of custom memory allocator alongside Yoga.

### DIFF
--- a/yoga/Yoga.cpp
+++ b/yoga/Yoga.cpp
@@ -11,7 +11,6 @@
 #include <string.h>
 #include <algorithm>
 #include <atomic>
-#include <malloc.h>
 #include "Utils.h"
 #include "YGNode.h"
 #include "YGNodePrint.h"
--- a/yoga/Yoga.cpp
+++ b/yoga/Yoga.cpp
@@ -11,7 +11,7 @@
 #include <string.h>
 #include <algorithm>
 #include <atomic>
-#include <memory>
+#include <malloc.h>
 #include "Utils.h"
 #include "YGNode.h"
 #include "YGNodePrint.h"
@@ -106,8 +106,24 @@ static int YGDefaultLog(
 #undef YG_UNUSED
 #endif
 
-static YGAllocatorAllocateFunc gAllocatorAllocateFunc = &malloc;
-static YGAllocatorFreeFunc gAllocatorFreeFunc = &free;
+#ifdef _MSC_VER
+#define YG_ALLOC_ALIGNED ([](size_t a, size_t s) -> void* { \
+  return _aligned_malloc(s, a);                             \
+})
+#define YG_FREE_ALIGNED ([](void* m) -> void {              \
+  _aligned_free(m);                                         \
+})
+#else
+#define YG_ALLOC_ALIGNED ([](size_t a, size_t s) -> void* { \
+  return aligned_alloc(a, s);                               \
+})
+#define YG_FREE_ALIGNED ([](void* m) -> void {              \
+  free(m);                                                  \
+})
+#endif
+
+static YGAllocatorAllocateFunc gAllocatorAllocateFunc = YG_ALLOC_ALIGNED;
+static YGAllocatorFreeFunc gAllocatorFreeFunc = YG_FREE_ALIGNED;
 
 YOGA_EXPORT void YGSetAllocationCallbacks(
     YGAllocatorAllocateFunc allocFunc,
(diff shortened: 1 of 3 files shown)


KitsuneAlex:
I also think it might be sensible to remove the YG name prefix from all functions within Yoga.h which are C++-only
and to move them into the **yoga** namespace.

KitsuneAlex:
Updated some things, test-compiled it on Windows x86_64 (under both Clang and MSVC), Linux x86_64 and MacOS x86_64 and ARM64 and everything's looking good.

KitsuneAlex:
Found another problem regarding static libraries and relocation on GCC/Clang, fixed that as well.

yoga/Yoga.cpp
```diff
@@ -106,6 +105,51 @@ static int YGDefaultLog(
...
+#define YG_ALLOC_ALIGNED ([](size_t a, size_t s) -> void* { \
+  return aligned_alloc(a, s);                               \
+})
+#define YG_FREE_ALIGNED ([](void* m) -> void {              \
+  free(m);                                                  \
+})
+#endif
+
+static YGAllocatorAllocateFunc gAllocatorAllocateFunc = YG_ALLOC_ALIGNED;
+static YGAllocatorFreeFunc gAllocatorFreeFunc = YG_FREE_ALIGNED;
```
NickGerleman, line 125: If these are global, they will need to be protected by mutex, in case multiple threads are using Yoga at once.

I think it would be cleaner if the allocator was set as part of YGConfig, though it would mean YGConfig itself could not be allocated using the custom allocator. Though, Yoga may otherwise call functions which perform their dynamic allocations, so if the goal is just for the large structures, I think it could still be okay.

yoga/Yoga.cpp
```diff
@@ -106,6 +105,51 @@ static int YGDefaultLog(
 #undef YG_UNUSED
 #endif
 
+#ifdef _MSC_VER
+#define YG_ALLOC_ALIGNED ([](size_t a, size_t s) -> void* { \
```
NickGerleman, line 109: nit: why are these macros vs a regular function?

yoga/Yoga.cpp
```diff
@@ -106,6 +105,51 @@ static int YGDefaultLog(
...
+    YGAllocatorFreeFunc* freeFunc) {
+  if(allocFunc != nullptr) {
+    *allocFunc = gAllocatorAllocateFunc;
+  }
+  if(freeFunc != nullptr) {
+    *freeFunc = gAllocatorFreeFunc;
+  }
+}
+
+YOGA_EXPORT void* YGMemoryAllocate(size_t alignment, size_t size) {
```
NickGerleman, line 145: nit: Do we need to expose allocate and free as part of the public Yoga API? This already has an API for getting the allocators which are set. It doesn't seem like a good code pattern for folks to use Yoga's APIs for their own memory allocation.

yoga/Yoga.cpp
```diff
@@ -106,6 +105,51 @@ static int YGDefaultLog(
...
+})
+#define YG_FREE_ALIGNED ([](void* m) -> void {              \
+  free(m);                                                  \
+})
+#endif
+
+static YGAllocatorAllocateFunc gAllocatorAllocateFunc = YG_ALLOC_ALIGNED;
+static YGAllocatorFreeFunc gAllocatorFreeFunc = YG_FREE_ALIGNED;
+
+YOGA_EXPORT void YGSetAllocationCallbacks(
```
NickGerleman, line 127:
```suggestion
YOGA_EXPORT void YGSetAllocator(
```

yoga/Yoga.h
```diff
@@ -364,6 +374,20 @@ YG_EXTERN_C_END
...
+  auto* memory = reinterpret_cast<T*>(YGMemoryAllocate(sizeof(T), alignof(T)));
+  new(memory) T(std::forward<A>(arguments)...);
+  return memory;
+}
+template<typename T>
+void YGFree(T* memory) {
+  memory->~T();
+  YGMemoryFree(memory);
+}
+
```
NickGerleman, line 390: This should not be part of the public API, since it is a helper for the internals of Yoga.

NickGerleman (changes requested):
I think letting Yoga use a custom allocator is a great feature.

I left some comments on the PR wrt implementation.

Issue #1131, opened 2022-02-18
Error when running tests in Windows 10 environment


shmax:
# Report

- [ ] I have searched [existing issues](https://github.com/facebook/yoga/issues) and this is not a duplicate

# Issues and Steps to Reproduce

On a Windows machine:
```sh
$ python -V
Python 2.7.8

$ buck --version
buck version 2021.01.12.01

$ java --version
openjdk 11.0.12 2021-07-20
OpenJDK Runtime Environment Microsoft-25199 (build 11.0.12+7)
OpenJDK 64-Bit Server VM Microsoft-25199 (build 11.0.12+7, mixed mode)

$ ant -version
Apache Ant(TM) version 1.10.12 compiled on October 13 2021

$ buck test //:yoga
PARSING BUCK FILES: FINISHED IN 0.0s
PARSING BUCK FILES: FINISHED IN 0.7s
Buck wasn't able to parse D:\repos\shmax\yoga\testutil\BUCK:
KeyError: '"CXX" is not defined in D:\\repos\\shmax\\yoga\\tools\\build_defs\\oss\\yoga_defs.bzl'
Call stack:
  File "D:\repos\shmax\yoga\.buckd\resources\2021.01.12.01\buck_server\buck_parser\profiler.py", line 507, in wrapped
    return func(*args, **kwargs)
  File "D:\repos\shmax\yoga\testutil\BUCK", line 1
    load("//tools/build_defs/oss:yoga_defs.bzl", "ANDROID", "CXX", "FBJNI_TARGET", "LIBRARY_COMPILER_FLAGS", "SOLOADER_TARGET", "subdir_glob", "yoga_cxx_library", "yoga_dep", "yoga_java_library")

This error happened while trying to get dependency '//testutil:testutil' of target '//:YogaTests'
```

I'm new to this project, so it's very likely I'm doing something wrong, but I have done my best to meet all the prerequisites laid out in the documentation. This particular error seems like it might be related to this recent commit from  @aniketmathur :

https://github.com/facebook/yoga/commit/477fedd1b64ad4ac45ef12ef1ea82966fdb7ceeb

Any thoughts?

PR #1132, opened 2022-02-18
Add missing reference to CXX to yoga_defs.bzl

shmax:
Fix for #1131 

This issue seems to have been introduced in https://github.com/facebook/yoga/commit/477fedd1b64ad4ac45ef12ef1ea82966fdb7ceeb

I don't know if this is a complete fix, as I'm still not having any luck getting the tests to run on Windows ("File not found: cl
"), but I thought I may as well get the conversation started.


NickGerleman:
https://github.com/facebook/yoga/pull/1166 removed Buck from the repo in general, since it hasn't been exercised. I am planning to follow up with a change that allows running C++ UTs via CMake. Though ideally we would expose tests for the JNI bindings to OSS, like we were some time ago.

PR #1138, opened 2022-04-06
Block Formatting and other fixes

alexei-spitfire:
Implemented block formatting, so we now support `display: block;` `inline-block;` and `inline-flex;`

This is done using a new version of the recursive layout algorithm `YGNodeBlockImpl()` which is heavily modified from the original `YGNodelayoutImpl()`. A wrapper function `YGLayoutNodeInternal()` determines whether a cached layout can be used, and if not it chooses one of these algorithms for the current node. Block formatting is disabled by default, and must be enabled using a YGConfig object, like so:
`YGConfigSetEnableBlockFormatting (yogaConfig.get(), true);` 

Other changes include support for fixed positioning and bugfixes to absolute positioning.

TODO:
Fix relative positioning (it's like "normal" static positioning, but an offset is applied at the end)
Implement vertical-align property for block formatting

### DIFF
--- a/yoga/Utils.h
+++ b/yoga/Utils.h
@@ -38,6 +38,7 @@
 
 struct YGCollectFlexItemsRowValues {
   uint32_t itemsOnLine;
+  bool isBlockNonInline;
   float sizeConsumedOnCurrentLine;
   float totalFlexGrowFactors;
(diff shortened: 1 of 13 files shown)


alexei-spitfire:
Accidental pr intended for another repo, please ignore!

PR #1141, opened 2022-04-13
Added suppression for warnings about unused parameters in 'publish<Type E>()' func

inobelar:


NickGerleman (approved):
LGTM

As a nit `[[maybe_unused]]` might be a clearer way to do the same thing.

PR #1144, opened 2022-04-29
【YogaKit】Fix two crashes. YogaKitSample upgrade to swift 5

Henrs:
1. if A view addSubviews(B), B view addSubviews(C),  sometimes we need A view addSubviews(C),   yoga  asset because of C had a owner,  when node insert a child node, i think we need to remove child node from owner first.
2. yoga maybe calculate a Nan value, but UIKit size can't accept a Nan Value, it will be cause unexpect crashes
3. YogaKitSample upgrade to swift 5

### DIFF
--- a/YogaKit/Source/YGLayout.m
+++ b/YogaKit/Source/YGLayout.m
@@ -489,10 +489,8 @@ static void YGApplyLayoutToViewHierarchy(UIView* view, BOOL preserveOrigin) {
           },
       .size =
           {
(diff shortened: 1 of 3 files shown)


NickGerleman:

Henrs:
> r

ok，Does version 2.0.0 mean that YogaKit is the final version？

NickGerleman:
Yes, 2.0 is the final version of YogaKit by Meta.

PR #1154, opened 2022-07-18
Rewrite CompactValue to avoid undefined behavior from the use of a union for type-punning 

htpiv:
C++ does not, pedantically, allow the use of unions for type-punning in the way that C does. Most compilers, in practice, do support it; however, recent versions of MSVC appear to have a bug that cause bad code to be generated due to this U.B. (see: https://developercommunity.visualstudio.com/t/Bad-code-generated-for-std::isnan-compil/10082631). This led to a series of issues in the react-native-windows project, see:
* https://github.com/microsoft/react-native-windows/issues/4122
* https://github.com/microsoft/react-native-windows/issues/8675

In C++20, the `<bit>` header and `bit_cast` function provide a pleasant API for type-punning. Since C++20 is not universally available, if the feature-test macro for `bit_cast` is not defined, memcpy is used instead.

### DIFF
--- a/yoga/CompactValue.h
+++ b/yoga/CompactValue.h
@@ -9,6 +9,9 @@
 
 #ifdef __cplusplus
 
+#ifdef __cpp_lib_bit_cast
+#include <bit>
+#endif
 #include "YGValue.h"
 #include "YGMacros.h"
 #include <cmath>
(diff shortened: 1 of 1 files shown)


yoga/CompactValue.h
```diff
@@ -9,6 +9,9 @@
 
 #ifdef __cplusplus
 
+#ifdef __cpp_lib_bit_cast
+#include <bit>
+#endif
```
rozele, line 14: Looks like internal builds are failing because std::memcpy requires `#include <cstring>`

yoga/CompactValue.h
```diff
@@ -157,13 +150,36 @@ class YOGA_EXPORT CompactValue {
...
+    static_assert(sizeof(u) == sizeof(f), "uint32_t and float must have the same size");
+    std::memcpy(&u, &f, sizeof(f));
+    return u;
+#endif
+  }
 
-  VISIBLE_FOR_TESTING uint32_t repr() { return payload_.repr; }
+  static float asFloat(uint32_t u) {
+#ifdef __cpp_lib_bit_cast
+    return std::bit_cast<float>(data);
```
rozele, line 172: Should this be:
```
    return std::bit_cast<float>(u);
```

Issue #1120, opened 2022-01-13
Enabling logging on Android crashes the app


matejdro:
# Report

# Issues and Steps to Reproduce

1. Create a YogaConfig with a logger
2. Create a YogaNode with that config
3. Attempt to layout it

# Expected Behavior

Logger should work

# Actual Behavior
Whole app crashes with

```
signal 6 (SIGABRT), code -1 (SI_QUEUE), fault addr --------
Abort message: 'JNI DETECTED ERROR IN APPLICATION: illegal class name 'Lcom/facebook/yoga/YogaLogLevel;'
        (should be of the form 'package/Class', [Lpackage/Class;' or '[[B')
        in call to FindClass
        from void com.facebook.ckage/Class', [Lpackage/Class;' or '[[B')
        in call to FindClass
        from void com.facebook.yoga.YogaNative.jni_YGNodeCalculateLayoutJNI(long, float, float, long[], com.facebook.yoga.YogaNodeJNIBase[])'
    x0  0000000000000000  x1  0000000000002c71  x2  0000000000000006  x3  0000007fdd736130
    x4  fefeff7edc724f04  x5  fefeff7edc724f04  x6  fefeff7edc724f04  x7  7f7f7f7fff7f7f7f
    x8  00000000000000f0  x9  0000007e4e49a118  x10 0000000000000000  x11 ffffff80fffffbdf
    x12 0000000000000001  x13 0000000019c11820  x14 0000007fdd734f40  x15 ffffffffffffffff
    x16 0000007e4e56fd20  x17 0000007e4e54a130  x18 0000007e5cace000  x19 0000000000002c71
    x20 0000000000002c71  x21 00000000ffffffff  x22 0000000000000000  x23 0000000000000000
    x24 0000007dba8b5ed6  x25 0000000000000001  x26 0000000000000000  x27 0000007dbb217000
    x28 b400007e5c3076c0  x29 0000007fdd7361b0
    lr  0000007e4e4fb938  sp  0000007fdd736110  pc  0000007e4e4fb968  pst 0000000000000000
```

matejdro:
@evancharlton this bug was not react-related though (we are not using React, only Yoga). Does this fix in another repository also address native Yoga?

evancharlton:
[...]

PR #1156, opened 2022-08-11
fix: Correctly resolve classes with FindClass(..)

evancharlton:
`JNIEnv`'s `FindClass(..)` function takes the classes in the standard
`foo/bar/Baz` class specification (unless they're special, like arrays).
Specifying them with `Lfoo/bar/Baz;` results in a
`ClassNotFoundException` being raised -- which is especially unhelpful
when intending to re-throw an exception.

Fixes: #1120

### DIFF
--- a/java/jni/YGJNIVanilla.cpp
+++ b/java/jni/YGJNIVanilla.cpp
@@ -147,7 +147,7 @@ static int YGJNILogFunc(
     if (*jloggerPtr) {
       JNIEnv* env = getCurrentEnv();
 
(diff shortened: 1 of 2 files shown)


sveinar:
@NickGerleman wrote in another PR that yoga has been without a full-time maintainer. 
@NickGerleman, would you be able to have a look at this PR as well?

It is bringing down the hostapp with a ClassNotFound exception when yoga wants to tell us what is really wrong instead of rethrowing it.

evancharlton:
Oh, nice - fixing it in the facebook/react-native repo automatically pushed the fix here.

I guess this PR can be closed!

PR #1159, opened 2022-09-10
fix: Check cpp_exceptions before throw exception

xiaoxiang781216:


NickGerleman:
Going to take #1006 for the same fix, since it is merge ready.

PR #1160, opened 2022-09-10
fix: flexWrap error when lineCount <= 1

xiaoxiang781216:


NickGerleman (changes requested):
Thanks for the PR!

Some of the logic here around cross-axis justification seems to be alive for the single line case, and there do seem to be specific branches in the existing code for where there is more than one line. Could you provide some more details on the error you are seeing, and why you think this is the right fix?

Contribution will also require signing the CLA the bot commented with.

PR #1161, opened 2022-09-29
Stale flex basis value used with cached values

rozele:
Summary:
We're occasionally seeing text truncation occur on Windows. It appears to occur when a cached main axis value is supplied to the flex basis computation for children, resulting in a stale flex basis value getting used.

This diff just contributes a unit test that reliably reproduces this case.

Differential Revision: D39929963


jeetiss:
I've experienced the same problem with `react-pdf` and `satori`
- https://github.com/vercel/satori/issues/393
- https://github.com/diegomura/react-pdf/issues/2182

Here is the [code sandbox](https://codesandbox.io/s/yoga-setmeasurefunc-misuse-v168ep) with `javascript` reproduction

PR #1165, opened 2022-10-06
Start Adding GitHub Actions

NickGerleman:
This change starts adding more coverage to GitHub Actions. Existing workflows are split up to be per-platform, and stale scripts, etc are removed.

We are currently limited a bit by issues with the build itself, but this still adds a good bit of coverage that readily works, and adds places to inject more.

Another option would have been to move these to CircleCI where we have more credits, or used docker images instead of manual setup steps. etc, The Yoga build and number of changes is very light though, so we don't really need the complexity yet.

Some TODOs:
1. Fix the Apple Builds (pod lint and pod install return errors seen by the community)
2. Add working Android UTs
3. Add C++ UTs
4. Add Apple Publish
5. Add version stamping


.github/actions/clang-format/action.yml
```diff
@@ -0,0 +1,24 @@
...
+    description: LLVM version to use # Should be kept roughly in sync with arcanist
+    required: false
+    default: 12
+
+runs:
+  using: "composite"
+  steps:
+    - name: Install
+      shell: bash
+      run: sudo apt-get install clang-format-${{ inputs.version }}
```
NickGerleman, line 16:
```suggestion
      run: sudo apt-get install -y clang-format-${{ inputs.version }}
```

.github/workflows/valiadate-android.yml
```diff
@@ -0,0 +1,30 @@
...
+
+  format:
+    name: Format
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: google-java-format
+        run: npx --yes google-java-format -i --glob=java/**/*.java
```
NickGerleman, line 30:
```suggestion
        run: npx --yes google-java-format -i --set-exit-if-changed --dry-run --glob=java/**/*.java
```

PR #1167, opened 2022-10-12
Fix memory leak

simonla:
image

`root_child0` be removed by`YGNodeRemoveChild(root, root_child0);`
so `YGNodeFreeRecursive(root);` can not free `root_child0`


motiz88:
Looks good, thanks @simonla!

PR #1173, opened 2022-11-15
Incorporate gap space into main axis overflow flag

NickGerleman:
Summary:
In https://github.com/facebook/react-native/issues/35351 we see incorrect child item height when the flex-wrap is enabled, the cross-axis is to be stretched, and main-axis overflow is caused by gap.

In YGDistributeFreeSpaceSecondPass, if we do not have overflow (determined by flexBasisOverflows), we have stretch cross-alignment, and we reason that nothing can add to main axis dimensions, we know we're a single line and want to take full cross dimensions. and can set YGMeasureModeExactly which uses parent dimensions. Guessing an optimization?

If we do have overflow, then we set YGMeasureModeAtMost to find minimum possible cross-axis dimensions instead.

`flexBasisOverflows` incorporates both computed flex basis, and margin, so it is more generally a flag for whether we will wrap. So we should incorporate gap spacing into it. E.g. it is also used for whether we should the match main axis parent dimension of the overall container. This change does just that, and renames the flag to `mainAxisOverflows`.

We will want to cherry-pick the fix for this into RN 0.71 since we have not yet introduced the community to the incorrect behavior, and we expect a lot of usage of flex-gap.

Differential Revision: D41311424

### DIFF
--- a/yoga/Yoga.cpp
+++ b/yoga/Yoga.cpp
@@ -2079,7 +2079,7 @@ static float YGDistributeFreeSpaceSecondPass(
     const float availableInnerCrossDim,
     const float availableInnerWidth,
     const float availableInnerHeight,
-    const bool flexBasisOverflows,
+    const bool mainAxisOverflows,
     const YGMeasureMode measureModeCrossDim,
     const bool performLayout,
     const YGConfigRef config,
@@ -2175,7 +2175,7 @@ static float YGDistributeFreeSpaceSecondPass(
         !YGNodeIsStyleDimDefined(
             currentRelativeChild, crossAxis, availableInnerCrossDim) &&
         measureModeCrossDim == YGMeasureModeExactly &&
-        !(isNodeFlexWrap && flexBasisOverflows) &&
+        !(isNodeFlexWrap && mainAxisOverflows) &&
         YGNodeAlignItem(node, currentRelativeChild) == YGAlignStretch &&
         currentRelativeChild->marginLeadingValue(crossAxis).unit !=
             YGUnitAuto &&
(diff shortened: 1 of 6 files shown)


yoga/Yoga.cpp
```diff
@@ -2884,7 +2884,8 @@ static void YGNodelayoutImpl(
 
   // STEP 3: DETERMINE FLEX BASIS FOR EACH ITEM
 
-  float totalOuterFlexBasis = YGNodeComputeFlexBasisForChildren(
+  // Computed basis + margins
+  const float totalOuterFlexBasis = YGNodeComputeFlexBasisForChildren(
```
jacobp100, line 2888: Because this function is only called once, should we change it to return the main axis size?

jacobp100:
I think what you’ve done makes sense in the current code

I’m confused why we do all this calculation at all. Don’t we already know if the children wrapped at this point - and is that not the same thing?

yoga/Yoga.cpp
NickGerleman: I was looking at that, but it isn't super clean, since the main purpose of the function is to mutate each node with its calculated axis. Imo it shouldn't be adding margin either, but that is already per child so doing it during the iteration makes sense to me.

NickGerleman:
> I’m confused why we do all this calculation at all. Don’t we already know if the children wrapped at this point - and is that not the same thing?

I think this place is the first place we can check. This is directly after we calculate flex basis per-item, which allows us to know if wrapping will happen. Then later we break into specific lines, and that calculation is item-by-item adding leading gap.

PR #1177, opened 2022-11-22
Migrate to emscripten

jeetiss:


javascript/sources/Node.cc
```diff
@@ -331,9 +331,8 @@ Value Node::getPadding(int edge) const {
       YGNodeStyleGetPadding(m_node, static_cast<YGEdge>(edge)));
 }
 
-Value Node::getGap(int gutter, ) {
-  return Value::fromYGValue(
-      YGNodeStyleGetGap(m_node, static_cast<YGGutter>(gutter)));
+float Node::getGap(int gutter) {
+  return YGNodeStyleGetGap(m_node, static_cast<YGGutter>(gutter));
```
jacobp100, line 335: I left these as `YGValues` in the hope that in the future, we add support for percentages. I just couldn't figure it out at the time

jeetiss:
@NickGerleman hello, I'm back

The bindings are ready so let's discuss the migration and the next steps.

- I just replaced `nbind` with `embind` and removed all `nbind` code. Is it okay or it would be better to maintain both versions while migrating?

- I wanna provide separate `dev` (with some additional assertions and checks) and `prod` (minified and with optimizations) bundles, so I need some bundler (rollup). What do you think about that? maybe you have suggestions here.

- I don't know what to do with CI. Can I setup GitHub action for testing like [this](https://github.com/jeetiss/exp-emscripten/blob/main/.github/workflows/test.yaml)  or it is not a big deal rn?

javascript/sources/Node.cc
jeetiss: this makes sense now, thanks for the clarification

but I have problems with types `YGNodeStyleGetGap` returns `float` that can't be converted with Value::fromYGValue

jeetiss: I think that is better to fix `YGNodeStyleGetGap` first, it should return `YGValue`

I can't pick the unit for this type now (should it be undefined or auto or digit?)

jacobp100: Ah ok - didn't realise the these functions were inconsistent. The unit is `YGUnitPoint`

NickGerleman:
This is incredible! I will take a close look soon. 

I think it is okay to fully replace nbind. It has been a long time since folks used it for Node native module versions of Yoga compared to packages like yoga-layout-prebuilt, so it is better imo to have something working, compared to full parity.

Re dev vs prod, I know Emscripten has options for what it emits (it will already minify), so I’m guessing the extra bits would be for the JavaScript frontend to it? If we can avoid the complexity, it might be better to distribute the package without running through another bundler, then the consuming project could bundle using the bundler they are already using.

I’d love it if we could add GitHub Actions validation, but I know it isn’t present right now. There are some workflows I added recently that could be pattern matched, and it would document how we are running tests against this. The website build is currently using the last published version of yoga-layout instead of the one in the repo too, so we don’t get validation from the website build job (tried to move this to workspaces recently but the website setup needs some serious renovation to support it).

jeetiss:
I got the warning 

```
warning: undefined symbol: _ZN8facebook4yoga24LayoutPassReasonToStringENS0_16LayoutPassReasonE (referenced by top-level compiled C/C++ code)
```

do you have any idea why it is here and maybe it can be fixed

NickGerleman:
> I got the warning
> 
> ```
> warning: undefined symbol: _ZN8facebook4yoga24LayoutPassReasonToStringENS0_16LayoutPassReasonE (referenced by top-level compiled C/C++ code)
> ```
> 
> do you have any idea why it is here and maybe it can be fixed

I think this means the build isn’t picking up [yoga/event/event.cpp](https://github.com/facebook/yoga/blob/acbffc8485bc230087ce23b2cb86551e1fa322d7/yoga/event/event.cpp).

javascript/sources/Node.cc
nicoburns: > I left these as YGValues in the hope that in the future, we add support for percentages. I just couldn't figure it out at the time

@jacobp100 I added in support for percentage gaps when I ported your gap changes to [Taffy](https://github.com/DioxusLabs/taffy) (a rust flexbox library originally based off of Yoga). Diff is [here](https://github.com/DioxusLabs/taffy/pull/248/files#diff-78a705bd422b6ec7ece906ca29973427ec10834766ca5bf1609459f0e4f539d5) if it helps you figure things out.

NickGerleman: That seems like a great help 🙂.

During H1 2023 I have some work on my plate to try to port some of the layout examples in WPT to Yoga compatible fixtures as well. Seems like a great resource for validating correctness of new layout props outside the context of browsers.

NickGerleman (commented):
I'm super grateful for this work. Given how the nbind version was long-term broken, we should be aggressive in taking improvements here.

If you don't mind, I'm going to try to push a commit to this PR adding the `yarn test` script you wrote to GitHub workflows. The previous lack of coverage for bindings let a lot of bad code slip into them, so if we have a working test script, it would help to both protect the change and demonstrate it is working when I import it 🙂.

javascript/Makefile
```diff
@@ -0,0 +1,32 @@
...
+		-s USE_CLOSURE_COMPILER=1 \
+		-s USE_ES6_IMPORT_META=0 \
+		-s MODULARIZE=1 \
+		-s ASSERTIONS=0 \
+		-s ALLOW_MEMORY_GROWTH=1 \
+		-s DYNAMIC_EXECUTION=0 \
+		-s TEXTDECODER=0 \
+		-s ENVIRONMENT='web' \
+		-s FILESYSTEM=0 \
+		-s MALLOC="emmalloc" \
```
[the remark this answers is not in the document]
NickGerleman: Testing locally:
emmalloc: 165kb
dlmalloc: 173kb

javascript/Makefile
```diff
@@ -0,0 +1,32 @@
+CC=emcc
+
+all: clean dir asm
+
+asm:
+	$(CC) sources/yoga/*.cpp sources/yoga/**/*.cpp sources/*.cc \
+		--bind \
+		-Isources \
+		-g0 \
+		-Os \
```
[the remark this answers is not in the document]
NickGerleman: Disabling exceptions and RTTI saves 10kb. LTCG does not lead to size savings, but is best practice for performance so I am tempted to think we should enable it (despite impact to build perf).

javascript/Makefile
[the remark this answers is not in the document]
jeetiss: Yes, I tested it too and it doesn't affect the perf

javascript/Makefile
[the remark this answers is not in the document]
jeetiss: yep, sure `emcc` build takes 2 times longer but it definitely makes sense

thanks for suggestion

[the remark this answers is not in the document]
jeetiss: 🚀

[Снимок экрана 2022-12-22 в 13 08 02](https://github.com/jeetiss/yet-another-yoga-layout-fork/pull/10)

NickGerleman:
I pushed a couple of changes here:
1. We have passing tests in GitHub Actions now 🎉
2. We build separate targets for asm.js and wasm.js, defaulting to asm.js for compat, but allowing platforms supporting wasm to use the wasm entrypoint from the export map (if using a modern bundler or Node). Anecdotally, the WASM build of tests ran in about half the time of the asm.js version.
3. Updated the babel versions, transforms. The old version didn't recognize Flow syntax to reexport all types, which I wanted to add to avoid duplication in the multiple entrypoints.

NickGerleman:
Here is the performance difference between asm.js and wasm being run by Node from the scenarios in the run-bench script. 
image

nicoburns:
> Here is the performance difference between asm.js and wasm being run by Node from the scenarios in the run-bench script.

Interesting. I'm not sure how representative these benchmarks are, but ~13ms for the "huge nested layout" was also what I was getting by using the `yoga-layout-prebuilt` package from NPM (although I was only measuring layout computation time and not including the tree setup time there).

I can't understand why "Align stretch in undefined axis" is so slow though. That only has 10 nodes (compared to 10,000 for the huge nested layout)! Admittedly it's not the same benchmark, but I was able to run a variant of the "huge nested layout" with depth of 1 instead of depth of 4 (so 10^1 = 10 rather than 10^4 = 10,000 nodes) in ~45 *micro* seconds.

One thing I did notice when benchmarking `yoga-layout-prebuilt` was that there seemed to be a fixed ~30ms penalty the first time I called `calculate_layout()` (can't remember exact name in yoga) in a given run (this happened even if other functions to setup nodes had already been called. I wonder if a similar penalty might be skewing the results here. I would say that the results of the 4 benchmarks are suspiciously similar.

NickGerleman:
I haven't tried to debug it to see if it is working as expected, but there is some mention in the script on skipping the first few iterations to avoid one-time costs, then averaging the rest.

Some of the tests I can see like "Align stretch in undefined axis" add 2000 child nodes under the root node, so there still is a lot there. That 2000 node count is the same for other tests, but "huge nested layout" is triply nested of 6 children it looks like.

nicoburns:
> Some of the tests I can see like "Align stretch in undefined axis" add 2000 child nodes under the root node, so there still is a lot there.

Ah, I can what I've done here. I'm reading the C benchmarks and assuming the JavaScript benchmarks are the same! The C benchmarks with the same name is only using 10 child nodes. These numbers make a lot more sense at a count of 2000 nodes.

NickGerleman:
Looking at the detail of Chrome not allowing synchronous compilation of WebAssembly files> 4KB, we should bite the bullet and change the bindings to require async initialization I think.

NickGerleman:
I think it would be reasonable to have a `yoga-layout/sync` documented in the readme which points to a synchronously obtained asmjs build. Technically, we could go further and have both sync and async asmjs and wasm builds, since Node doesn't have the same sync compilation limitation that Chrome does. So we could have a default async variant, then the sync one would be the same export map as the async one, minus browsers getting the asmjs version.

NickGerleman:
@jeetiss What do you think of a design where we:
1. Package 4x entrypoints, for the matrix of sync/async and asm/wasm
2. The default entrypoint provides a version where the user needs to call `await Yoga.load()` or similar to allow for async wasm compilation. Default to wasm for node, browser, but default to asm elsewhere or for old clients that don't understand export maps.
3. A `/sync` entrypoint provides the legacy sync API, but falls back to asm.js on browsers to allow it (at a size/performance cost), but we still let Node use sync wasm compilation

This change in general is a breaking API change compared to the last package, and means semver wise we ideally ought to do a `yoga@2.0.0` (hoping it doesn't create confusion as Yoga 2 is a bit overloaded lol). I still don't have a publishing pipeline set up, and might end up just starting with Maven/npm since the Apple/YogaKit build is not in a great state and I haven't built the context to really critically understand the set of problems and fixes for it.

After this change, but before we publish that new breaking version, it might make sense to make the change to `YGNodeStyleGetGap` to move to `YGValue` (@jacobp100).

Also want to apologize for commandeering this PR a bit 😅. Original goal was just to add GitHub actions/apply the flag changes, but I got a little carried away hacking at this and trying to understand everything here.

jeetiss:
@NickGerleman sounds awesome, I am glad that you are pushing this forward 🚀

PR #1189, opened 2022-12-09
Update Yoga.cpp

jacobp100:
Just to see test results on CI


jacobp100:
@cortinico are you able to give me approval to run the workflows?

NickGerleman:
> @cortinico are you able to give me approval to run the workflows?

I started this off. But the CI run in OSS doesn’t run the unit tests. I’m going to prioritize finally getting the GTest UTs running in OSS again before I take holiday off after the end of this week.

NickGerleman:
I will pull this so I can report UT results.

NickGerleman:
It looks like this change is passing all UTs 👍

NickGerleman:
It's... not the properly debuggable C++ build just yet, but FYI @jacobp100 @intergalacticspacehighway that tests generated from the fixtures should now be run in the "JavaScript / Test" workflows. Still working to get the rest of the suite enabled and a better local experience.

NickGerleman:
The test failure being shown now is because the automatically merged output with this change and main has both your fix, and the one from @intergalacticspacehighway, doing the same thing I think.

NickGerleman:
C++ UTs are now running in OSS 👍

Should be relatively easy to run and debug locally now too as well.

NickGerleman:
Got around to removing the gap-specific hack in https://github.com/facebook/yoga/pull/1380

Note that we do want to do this at the point of the last element, instead of the first, since we use `betweenMainDim` for positioning, and will otherwise incorrectly position the child after the first.

PR #1194, opened 2022-12-27
Add newer fixtures

NickGerleman:
Summary:
flexlayout has the same fixtures forked from Yoga, with some disabled, and some added. This syncs that list to Yoga, adding the fixes made to it along with new fixtures.

One new test absolute_child_with_max_height_larger_shrinkable_grandchild is failing, along with two fixtures where Chrome had recently changed its behaviors. These are commented out at the moment before I do something smarter with the test runner to allow more fixtures.

Differential Revision: D42242096

### DIFF
--- a/csharp/tests/Facebook.Yoga/YGAbsolutePositionTest.cs
+++ b/csharp/tests/Facebook.Yoga/YGAbsolutePositionTest.cs
@@ -26,7 +26,6 @@ namespace Facebook.Yoga
 
             YogaNode root_child0 = new YogaNode(config);
             root_child0.PositionType = YogaPositionType.Absolute;
-            root_child0.Start = 10;
             root_child0.Top = 10;
             root_child0.Width = 10;
             root_child0.Height = 10;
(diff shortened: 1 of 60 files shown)


nicoburns:
@NickGerleman Not the biggest issue, but the test `justify_content_colunn_max_height_and_margin` has a typo (should be `column` not `colunn`) reported to Taffy here https://github.com/DioxusLabs/taffy/pull/330. The issue wasn't introduced by this PR, but maybe it would be a good place to fix it?

nicoburns:
Looks like gtest allows you to disable tests by prefixing the test name with `DISABLED_`. Seems like it could be a good way to go for known failures? https://stackoverflow.com/questions/7208070/googletest-how-to-skip-a-test

NickGerleman:
> Looks like gtest allows you to disable tests by prefixing the test name with `DISABLED_`. Seems like it could be a good way to go for known failures? https://stackoverflow.com/questions/7208070/googletest-how-to-skip-a-test

Teaching the test generator to emit annotations to skip tests would be a good concrete step towards getting some of this cleaned up. Patterned after what was done for experimental features, we could maybe add an HTML attribute like `skip-test="true"` or `data-skip-test="true"`.

Then we would emit for GTest, add `test.skip` for Jest, etc, for the different test backeends.

PR #1207, opened 2023-01-03
Restructure the C# bindings for Yoga to be based on .NET Standard 2.0 and the new .NET 6.0+ target platform work.

jkoritzinsky:
Updated the C# bindings to build using the new simplified .NET SDK tooling. The C# projection now builds for .NET Standard 2.0 (to support Unity, .NET Framework, UWP, and modern .NET) and .NET 6.0 for iOS (to have the native calls correctly support calling into a statically-linked library).

The native changes were needed to get the C# tests passing on Windows x64 (tested locally).

This PR does not update any build, test, or package scripts. I can either add that to this PR or follow up with another PR.


NickGerleman (commented):
This is great stuff! Very excited to see the C# bindings work again.

Would be good to clean up some of the dead project configuration around here. E.g. looking at the new source tree, I can see a Xamarin sln referencing folders no longer present, along with a non-vcxproj build script that doesn't work/I'm guessing is no longer used. 

I can see there is updated code for Mono binding to YogaKit on iOS. From what I can tell the previous iOS build used that build script instead of depending on the vcxproj. Will those non-Windows builds still work? If not I would reccomend removing as well (I'm okay losing parity if we get something presently working and maintainable in return).

I think we should add some Github workflows here to excersise any builds/NUnit tests which are working, for the matrix we want to support. My preference would be to do that as part of the change, so we can demonstrate everything is working. The "validate-xxx" workflows should be relatively pattern matchable but I can also help with this, given the permutations we want to support.

NickGerleman:
Also feel free to ignore the JavaScript linting failure. Made a change internally without realizing Prettier wasn't turned on for Yoga in internal validation. Working on fixing that up soon so it won't happen again.

csharp/Yoga/Yoga.vcxproj
```diff
@@ -163,7 +163,7 @@
       <PrecompiledHeader>NotUsing</PrecompiledHeader>
       <WarningLevel>Level3</WarningLevel>
       <Optimization>Disabled</Optimization>
-      <PreprocessorDefinitions>_DEBUG;_WINDOWS;_USRDLL;YOGA_EXPORTS;FB_ASSERTIONS_ENABLED=0;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>DEBUG;_DEBUG;_WINDOWS;_USRDLL;YOGA_EXPORTS;FB_ASSERTIONS_ENABLED=0;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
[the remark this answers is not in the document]
NickGerleman: For extra credit, these are the flags I'm tentatively setting in the CMake Build https://github.com/facebook/yoga/blob/2327bad973991efc748b3a1b953fa2b0b5808b90/cmake/project-defaults.cmake

They're not too far off from MSVC defaults, so it's probably okay if they're not perfectly matched, but having a reference set of flags consistently followed might make things easier to reason about.

jkoritzinsky:
I've added a build and test GitHub Action at the `.github/workflows/validate-csharp.yml` path. It will validate the build, package, and test steps for windows-x64 and linux-x64 on Ubuntu. The linux build should probably build against an older version of glibc to make it more usable across linux distributions, but at least for now my main goal was getting something working.

jkoritzinsky:
I haven't hooked up testing for iOS yet as I'm not familiar enough with GitHub actions to know the best way to run tests against an iOS simulator or device.

NickGerleman:
> The linux build should probably build against an older version of glibc to make it more usable across linux distributions, but at least for now my main goal was getting something working.

I looked into that a bit for the CMake change I'm working on. Right now the repo is keeping things simple, with the GitHub hosted/managed runners. The oldest LTS Ubuntu supports is 18.04, with GCC 6. but GitHub only supports back to 20.04.

Ideas were:
1. Select different toolchain versions on top of a current install
2. Convert the `setup-xxx.yaml` files into a Dockerfile and start building our own images ala https://github.com/marketplace/actions/build-and-push-docker-images

There is lower hanging fruit though, so it's not worth thinking about too much yet.

csharp/Facebook.Yoga/Facebook.Yoga.csproj
```diff
@@ -1,5 +1,15 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;net6.0-ios</TargetFrameworks>
+    <IsPackable>true</IsPackable>
   </PropertyGroup>
+  
+  <ItemGroup>
+    <PackageReference Include="NuGetizer" Version="0.9.1" />
```
NickGerleman, line 8: TIL this exists 👀

.github/workflows/validate-csharp.yml
```diff
@@ -0,0 +1,77 @@
...
+    name: Build and Package [${{ matrix.os }}]
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [windows-latest]
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Setup
+        uses: ./.github/actions/setup-js
```
NickGerleman, line 42: `setup-js` installs the Emscripten toolchain, which we don't really need. But it seems like we could factor this to instead be a setup script for CS with:
1. "actions/setup-dotnet"
2. "dotnet workload restore" (current CI failures point to needing to add this)

```suggestion
        uses: ./.github/actions/setup-cs
```

.github/workflows/validate-csharp.yml
```diff
@@ -0,0 +1,77 @@
...
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+env:
+  FORCE_COLOR: 3
+
```
NickGerleman, line 12:
```suggestion
```

This env variable is Node-specific and can be removed.

.github/workflows/validate-csharp.yml
```diff
@@ -0,0 +1,77 @@
...
+  test:
+    name: Test [${{ matrix.os }}]
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Setup .NET
```
NickGerleman, line 23: See other comment about factoring this into `setup-cs`.

.github/workflows/validate-csharp.yml
```diff
@@ -0,0 +1,77 @@
...
+  build-native:
+    name: Build Native Components [${{ matrix.os }}]
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        include:
+          - os: windows-latest
+            rid: win-x64
+          - os: ubuntu-latest
+            rid: linux-x64
```
NickGerleman, line 65:
```suggestion
          - os: windows-latest
            rid: win-x64
          - os: windows-latest
            rid: win-arm64
          - os: ubuntu-latest
            rid: linux-x64
          - os: ubuntu-latest
            rid: linux-arm64
          - os: ubuntu-latest
            rid: android-arm64
          - os: macos-latest
            rid: ios-arm64
```

csharp/Facebook.Yoga/Facebook.Yoga.csproj
```diff
@@ -0,0 +1,15 @@
...
+    <TargetFrameworks>netstandard2.0;net6.0-ios</TargetFrameworks>
+    <IsPackable>true</IsPackable>
+  </PropertyGroup>
+  
+  <ItemGroup>
+    <PackageReference Include="NuGetizer" Version="0.9.1" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <SupportedRid Include="win-x64;linux-x64" />
```
NickGerleman, line 12:
```suggestion
    <SupportedRid Include="win-x64;win-arm64;linux-x64;linux-arm64;ios-arm64;android-arm64" />
```

csharp/Yoga/CMakeLists.txt
```diff
@@ -0,0 +1,69 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+cmake_minimum_required(VERSION 3.4.1)
```
NickGerleman, line 6:
```suggestion
cmake_minimum_required(VERSION 3.14)
```

Planning to bump the min-version across the board in the main CMake build so that we can use `target_link_options` and newer policies.

csharp/Yoga/CMakeLists.txt
```diff
@@ -0,0 +1,69 @@
...
+
+set(SOURCES YGInterop.cpp)
+
+if (MSVC)
+    list(APPEND SOURCES Yoga.rc)
+endif()
+
+file(GLOB yogajni_version_script
+        yogajni.version)
+
```
NickGerleman, line 57:
```suggestion
```

csharp/Yoga/CMakeLists.txt
```diff
@@ -0,0 +1,69 @@
...
+add_library(yoga SHARED ${SOURCES})
+
+target_include_directories(yoga PRIVATE
+    ${yogacore_DIR})
+
+target_link_libraries(yoga yogacore)
+
+if (NOT MSVC)
+    target_link_libraries(yoga -Wl,--gc-sections)
```
NickGerleman, line 66: AppleClang uses `--dead_strip` instead of `--gc-sections`. So I think this will cause the iOS build to fail.

```suggestion
    target_link_options(yoga PRIVATE
    # Discard unused sections
    $<$<CONFIG:RELEASE>:$<$<CXX_COMPILER_ID:Clang,GNU>:-Wl,--gc-sections>>
    $<$<CONFIG:RELEASE>:$<$<CXX_COMPILER_ID:AppleClang>:-Wl,-dead_strip>>)
```

csharp/Yoga/CMakeLists.txt
```diff
@@ -0,0 +1,69 @@
...
+        -Wall
+        -std=c++11)
+endif()
+
+set(SOURCES YGInterop.cpp)
+
+if (MSVC)
+    list(APPEND SOURCES Yoga.rc)
+endif()
+
```
NickGerleman, line 54: Not seeing the corresponding resource file?

csharp/Facebook.Yoga/Facebook.Yoga.csproj
```diff
@@ -0,0 +1,15 @@
...
+    <PackageFile Include="@(SupportedRid->'Facebook.Yoga.Native.%(Identity)')" PackFolder="Dependency" Version="$(Version)" />
```
NickGerleman, line 13:
```suggestion
    <PackageFile Include="@(SupportedRid->'Facebook.Yoga.Native.%(Identity)')" PackFolder="Dependency" Version="$(Version)" PackageLicenseExpression="MIT" />
```

csharp/Yoga/CMakeLists.txt
NickGerleman: Not required, but it wouldn't be a bad idea to set a DLL version in one.

.github/workflows/validate-csharp.yml
```diff
@@ -0,0 +1,77 @@
...
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Setup .NET
+        uses: actions/setup-dotnet@v3
+        with:
+          dotnet-version: 6.0.x
+      
+      - name: Restore Workloads
+        run: dotnet workload restore
```
NickGerleman, line 29:
```suggestion
        run: dotnet workload restore
        working-directory: csharp
```

.github/workflows/validate-csharp.yml
```diff
@@ -0,0 +1,77 @@
...
+
+      - name: dotnet pack
+        run: dotnet pack -c Release
+        working-directory: csharp
+
+  build-native:
+    name: Build Native Components [${{ matrix.os }}]
```
NickGerleman, line 57:
```suggestion
    name: Build Native Components [${{ matrix.os }}][${{ matrix.rid }}]
```

.github/workflows/validate-csharp.yml
jkoritzinsky: We can't do this for all of these platforms until we set up native builds for each platform. Rid here is mainly for target folder. The native build doesn't yet have cross-build infrastructure set up. If we could get the one entry-point that is C#-specific into the regular yoga code-base, then we could chain together the builds and let the regular yoga build handle all of the cross-building for yoga platforms and this step would just package the files into NuGet packages.

csharp/Facebook.Yoga/Facebook.Yoga.csproj
jkoritzinsky: I don't think you specify the license expression on the dependency. I'll add it to the project files for each package.

csharp/Yoga/CMakeLists.txt
jkoritzinsky: The resource file already existed, so I didn't make any changes to it. Just left it as-is.

.github/workflows/validate-csharp.yml
NickGerleman: Ah, got it.

.github/actions/setup-cs/action.yml
```diff
@@ -0,0 +1,13 @@
...
+  using: "composite"
+  steps:
+      - name: Setup .NET
+        uses: actions/setup-dotnet@v3
+        with:
+          dotnet-version: 6.0.x
+      
+      - name: Restore Workloads
+        run: dotnet workload restore
+        working-directory: csharp
```
NickGerleman, line 13:
```suggestion
        shell: bash
        working-directory: csharp
         
```

csharp/Yoga/CMakeLists.txt
```diff
@@ -0,0 +1,69 @@
...
+        # Enable warnings and warnings as errors
+        /W3
+        /WX
+        # Disable RTTI
+        $<$<COMPILE_LANGUAGE:CXX>:/GR->
+        # Use /O2 (Maximize Speed)
+        $<$<CONFIG:RELEASE>:/O2>)
+else()
+    add_compile_options(
+        -fno-omit-frame-pointer
```
NickGerleman, line 40:
```suggestion
        -fPIC
        -fno-omit-frame-pointer
```

csharp/Facebook.Yoga/Facebook.Yoga.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;net6.0-ios</TargetFrameworks>
```
NickGerleman, line 3: It looks like we need to put "Microsoft.ios.Sdk.net6" behind a condition.

csharp/Yoga/CMakeLists.txt
```diff
@@ -0,0 +1,70 @@
...
+        # Enable warnings and warnings as errors
+        /W3
+        /WX
+        # Disable RTTI
+        $<$<COMPILE_LANGUAGE:CXX>:/GR->
+        # Use /O2 (Maximize Speed)
+        $<$<CONFIG:RELEASE>:/O2>)
+else()
+    add_compile_options(
+        -fPIC
```
NickGerleman, line 40:
```suggestion
```

csharp/Yoga/CMakeLists.txt
```diff
@@ -0,0 +1,70 @@
...
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+cmake_minimum_required(VERSION 3.14)
+
+project(yoga)
+
+set(CMAKE_VERBOSE_MAKEFILE on)
+
```
NickGerleman, line 11:
```suggestion
set(CMAKE_POSITION_INDEPENDENT_CODE ON)

```

NickGerleman:
Good news:
image

NickGerleman:
This is the GitHub Actions restriction making things a bit of a pain. Not sure how contributions count wrt to how imports work instead of merging, but it might be possible to change that status to allow getting quick feedback without intervention. 

CE7C9277-6BC9-4B91-9D89-EDDE72D64C3C

NickGerleman:
Where possible I'd reccomend leaning on Ubuntu VMs instead of the Mac images. One is a VM in Azure and the other is from a pool of Mac minis maintained by GitHub/MSFT. The Macs are usually slower, have higher queue times, and flakier network. But it's the only option for when you really do need to build using their toolchain. So definitely can be used if needed, but not a "free" switch.

csharp/Yoga/Facebook.Yoga.Native.msbuildproj
```diff
@@ -0,0 +1,22 @@
...
+    <PackageId>Facebook.Yoga.Native.$(RuntimeIdentifier)</PackageId>
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
+    <PackageLicenseExpression>MIT</PackageLicenseExpression>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="NuGetizer" Version="0.9.1" />
+  </ItemGroup>
+
+  <Target Name="BuildNative" BeforeTargets="CoreBuild" Returns="@(NativeFileFullPath)">
+    <Exec Command="cmake -S . -B build -DCMAKE_INSTALL_PREFIX=binnative &amp;&amp; cmake --build build --target install --config $(Configuration)" />
```
NickGerleman, line 14:
```suggestion
    <Exec Command="cmake -S . -B build -DCMAKE_INSTALL_PREFIX=binnative -DCMAKE_BUILD_TYPE='$(Configuration)' &amp;&amp; cmake --build build --target install --config $(Configuration)" />
```

NickGerleman: I remember writing this somewhere earlier but might have closed the tab. Apologies if it is a duplicate.

NickGerleman:
@jkoritzinsky what do you think about merging this as-is? I think it is going to be a better reference than what we have right now, and we can leave tests disabled in the meantime.

jkoritzinsky:
I'm fine merging this as-is. I haven't had time to investigate the test failures.

NickGerleman:
Talking offline, we decided to go a separate route, of removing the bindings from being in tree. This was commited with https://github.com/facebook/yoga/commit/5e74e20a4df186827dce57ec663063be9898a0fb

PR #1213, opened 2023-01-15
Round pixel values of `intrinsicSize` with `YGRoundPixelValue`

stleamist:
This PR makes `intrinsicSize` return rounded pixel values using `YGRoundPixelValue` as `YGApplyLayoutToViewHierarchy` does.

This change prevents text truncation due to the low precision of C++ `float` when sizing the view containing `UILabel` using `intinsicSize`.

```swift
override func sizeToFit() {
  self.bounds.size = self.yoga.intrinsicSize
}
```

### DIFF
--- a/YogaKit/Source/YGLayout.m
+++ b/YogaKit/Source/YGLayout.m
@@ -320,11 +320,15 @@ YG_EDGE_PROPERTY(gap, Gap, Gap, YGGutterAll)
 }
 
 - (CGSize)intrinsicSize {
-  const CGSize constrainedSize = {
+  CGSize constrainedSize = {
       .width = YGUndefined,
       .height = YGUndefined,
   };
-  return [self calculateLayoutWithSize:constrainedSize];
+  constrainedSize = [self calculateLayoutWithSize:constrainedSize];
+  return (CGSize){
+      .width = YGRoundPixelValue(constrainedSize.width),
+      .height = YGRoundPixelValue(constrainedSize.height),
+  };
 }
 
 - (CGSize)calculateLayoutWithSize:(CGSize)size {


NickGerleman:
Change makes sense to me.

For record-keeping, could you add a before/after screenshot to the PR description?

For awareness, the automated tests for YogaKit, and build on latest toolchain have been broken for a while, so we will have reduced coverage for this change against any existing automated tests. It's been on the infrastructure todo list, but after fixing the CMake build/test experience, and UTs for JNI bindings/Android build. There are a lot of PRs fixing various bits of the build, but I haven't had the experience with the Apple build to understand the whole of what we should be doing to the existing build logic, vs if we should just axe it, follow the Cocoapods/SPM guide for building a new package + example, then follow the leads where it needs code-changes.

rozele:
Perhaps I'm missing something here, but as I understand, Yoga already returns values rounded to the pixel grid assuming you set the scale via YGConfigSetPointScaleFactor. Why do you need to round a value that should presumably already be rounded?

NickGerleman (changes requested):
See comments from Eric, and request for test plan

NickGerleman:
> Perhaps I'm missing something here, but as I understand, Yoga already returns values rounded to the pixel grid assuming you set the scale via YGConfigSetPointScaleFactor. Why do you need to round a value that should presumably already be rounded?

Just to check, is the idea here that:
1. Yoga already does its own rounding (`YGRoundToPixelGrid()`)  based on `pointScaleFactor`, when laying out a node.
2. The YogaKit-specific `YGRoundPixelValue()` and its usage should be removed, and we should rely on Yoga to do all of the rounding itself

I am seeing that YogaKit is already calling `YGConfigSetPointScaleFactor()` (though assuming the density never changes, which is... probably fine on iOS). So this would end up looking like just removing the existing usage I think.

NickGerleman:

PR #1228, opened 2023-02-18
typescript: Fix support for moduleResolution NodeNext

cc-bbohlender:
The `.js` file ending is expected for moduleResolution `NodeNext` in typescript. The changes in this PR should enable both the current `node` moduleResolution and the `NodeNext` moduleResolution in typescript.

Additionally, this fixes a typo for the definition of the `unsetMeasureFunc`.


javascript/src_js/index.d.ts
```diff
@@ -7,9 +7,9 @@
  * @format
  */
 
-import type {Yoga} from './wrapAsm';
+import type {Yoga} from './wrapAsm.js';
 
-export * from './generated/YGEnums';
-export * from './wrapAsm';
+export * from './generated/YGEnums.js';
```
NickGerleman, line 12: Is this type resolved correctly, importing from `wrapAsm.js` instead of `wrapAsm.d.ts`?

NickGerleman (commented):
This change tries to make it so that the TS here can be interpreted as either CommonJS or ESM.

I wonder if instead, we should add [`"type": "commonjs"`](https://www.typescriptlang.org/docs/handbook/esm-node.html#type-in-packagejson-and-new-extensions) to the package, so that it is only ever interpreted in a single way.

We don't have present tests over the TypeScript typings, for either configuration, so I imagine it might be better to only try to support one.

javascript/src_js/index.d.ts
XantreGodlike: Yes, types are binded to file

NickGerleman (changes requested):
It sounds like this was root caused to your fork having different `package.json` fields set?

cc-bbohlender:
> It sounds like this was root caused to your fork having different `package.json` fields set?

Yeah, think so too. We're discussing options
https://github.com/shuding/yoga-wasm-web/pull/15#issuecomment-1438122182

However, I also think that updating this package to modules and es6 should be done anyways, but maybe not so soon :)

javascript/src_js/wrapAsm.d.ts
```diff
@@ -156,7 +156,7 @@ export type Node = {
   setWidth(width: number | string): void,
   setWidthAuto(): void,
   setWidthPercent(width: number): void,
-  unsetMeasureFun(): void,
+  unsetMeasureFunc(): void,
```
jeetiss, line 159: @cc-bbohlender could you create another PR with this fix? it is not blocked

cc-bbohlender: @jeetiss done in here https://github.com/facebook/yoga/pull/1231

PR #1233, opened 2023-02-26
Typescript: improve enum types and setMargin type

cc-bbohlender:
Currently, it is impossible to write `node.setPositionType(2)` instead of `node.setPositionType(POSITION_TYPE_ABSOLUTE)`. I understand that the idea is to force explicit usage of the enums. However, declaring `type POSITION_TYPE_ABSOLUTE = 2 & ['POSITION_TYPE']` artificially limits use cases, where, for example, the state should be serialized typesafe.

Additionally, this PR fixes the incorrect typing of `setMargin(edge: Edge, margin: number)` by extending the type to `margin: number | string`

### DIFF
--- a/javascript/src_js/wrapAsm.d.ts
+++ b/javascript/src_js/wrapAsm.d.ts
@@ -135,7 +135,7 @@ export type Node = {
   setHeightPercent(height: number): void,
   setJustifyContent(justifyContent: Justify): void,
   setGap(gutter: Gutter, gapLength: number): Value,
-  setMargin(edge: Edge, margin: number): void,
+  setMargin(edge: Edge, margin: number | string): void,
   setMarginAuto(edge: Edge): void,
   setMarginPercent(edge: Edge, margin: number): void,
   setMaxHeight(maxHeight: number | string): void,


NickGerleman (changes requested):
> Currently, it is impossible to write node.setPositionType(2) instead of node.setPositionType(POSITION_TYPE_ABSOLUTE). I understand that the idea is to force explicit usage of the enums. However, declaring type POSITION_TYPE_ABSOLUTE = 2 & ['POSITION_TYPE'] artificially limits use cases, where, for example, the state should be serialized typesafe.

This was an intentional design decision, so that the underlying data type is opaque/an implementation detail. It also avoids issues like enum types being exposed on hover as entities like `0 | 1 | 2`, or being able to plug any enum into any method (it removes the type safety).

If you are wanting to serialize these enums externally, I would recommend adding something like a switch/case over whatever format you are using.

javascript/src_js/wrapAsm.d.ts
```diff
@@ -135,7 +135,7 @@ export type Node = {
   setHeightPercent(height: number): void,
   setJustifyContent(justifyContent: Justify): void,
   setGap(gutter: Gutter, gapLength: number): Value,
-  setMargin(edge: Edge, margin: number): void,
+  setMargin(edge: Edge, margin: number | string): void,
```
NickGerleman, line 138: This matches the other types here (converted from previous flowtype typings), but we could also replace this + others with a template literal type for better safety (added in TS 4.1, which is the oldest supported by DefinitelyTyped).

```suggestion
  setMargin(edge: Edge, margin: number | 'auto'  | `${number}%`): void,
```

NickGerleman:
> Using the enums from typescript would make sense here since they would be present when hovering/autocompleting and would look like: node.setAlignItems(Align.ALIGN_AUTO). The constants could still be exported via export ALIGN_AUTO = Align.ALIGN_AUTO to not break anything.

I think this is a good solution, with the caveat that the TypeScript enums are a runtime concept, and the underlying source is not currently TypeScript. I am not sure if it can be kosher to replicate the same structure in JS, then expose in typings as an enum, or if we really should be making the source TypeScript to do that.

It seems like `const enum` has well defined runtime semantics that could be matched (and usable from vanilla JS):
1. https://babeljs.io/docs/babel-plugin-transform-typescript#optimizeconstenums
2. https://www.typescriptlang.org/docs/handbook/enums.html#const-enums

NickGerleman (approved):

PR #1234, opened 2023-03-07
Python bindings

nicomt:
I implemented python bindings for a project at work and was wondering if there is interest to have to host this in the main repo.
The bindings are mostly based on the js emscripten native implementation with some minor changes to work with pybind11 and take advantage of Python's memory management.

I had to manually change a couple of lines in the test after they were generated but I assume that is okay because it seems to be the case for other languages too. 
4481b1ee08c9eddcc5e6fb5bb6844b054250ad17


nicomt:
Just so you know, the project builds using setuptools with the pybind11 extension to build the project.
If pip 10+ is installed, you should be able to run `pip install .` or `pip wheel .` from the root directory to install and build
There is more info here
https://pybind11.readthedocs.io/en/stable/compiling.html#pep-518-requirements-pip-10-required

NickGerleman:
> I implemented python bindings for a project at work and was wondering if there is interest to have to host this in the main repo.

This is some great work, I need to think a little bit more about this.

Bindings contributed to the main repo have historically languished after contribution. Especially when the build was not continuously tested (e.g. I would expect we exercise the build and tests as part of any new bindings as part of GitHub Actions). So, what has historically happened for e.g. the C# bindings is that they were contributed, then any changes to the core engine did not make bindings updates. If we are validating, it also means cross-cutting changes need to change code for more bindings as well.

These builds will also naturally degrade and need maintenance over time. E.g. as language versions get deprecated, OS images change, PyPI tokens get updated (and we would need to manage those as part of getting publish infra back up). So without that maintenance from us, we end up in a state where we have folks with expectations we are supporting this, but we may not have experience to be.

Historically, when the repo saw less attention, going through the main repo was a barrier to innovating on the bindings, or velocity to contribute to them.

On the plus side, it creates exposure to the bindings which are legitimately useful (contributions are shared to one place), and means they are easier to keep in sync with main Yoga.

I think a solution if we did want that level of support is if we clarify a support level on the bindings in the repo which are externally supported. So that we do have a centralized solution, but are more hands off with it, beyond hosting it in the single place. Though I am still not sure that would be much better than the decentralized solution.

What do you think of everything here?

NickGerleman:
Sorry for the very late response to this (was on vacation for several weeks). But I think my preference for this would be to keep this out of tree. I do like your suggestion of pointing to the project in the Yoga README or Yoga website though (and there are other bindings which might be worth pointing out as well).

ankit-gautam23 (commented):
I would like to ask to split PR so it will be easy for reviewing.

nicomt:
> I would like to ask to split PR so it will be easy for reviewing.

@ankit-gautam23 
I'm not sure I understand but maybe looking through the commits would be useful.
https://github.com/facebook/yoga/pull/1234/commits
Not sure if there is another way to split this that makes sense.

PR #1236, opened 2023-03-12
fix: types tweaks

jeetiss:
- format types with `prettier`
- apply suggestion from https://github.com/facebook/yoga/pull/1233#discussion_r1120807560


javascript/src_js/index.d.ts
```diff
@@ -7,9 +7,9 @@
  * @format
  */
 
-import type {Yoga} from './wrapAsm';
+import type { Yoga } from "./wrapAsm";
```
NickGerleman, line 10: I forgot that I had told ESLint to ignore `.d.ts` files. https://github.com/facebook/yoga/blob/c09405d58c624fd26d5e0a7ed98bd10c5cc4c05b/javascript/.eslintrc.js#L12

I did that at the time because I didn't want to spend the time then to set up `typescript-eslint` a TS dep, and tests against the types. But it would be great to fix that up eventually.

NickGerleman (approved):

PR #1286, opened 2023-05-10
Add `data-disabled` to test fixtures

NickGerleman:
Summary:
This can be marked in fixtures to skip a test without commenting it out. We add one more usage of this.

The same functionality existed (unused) before for `experiments`, which I changed to `data-experiments`.

Formatting of JS tests changed to be closer to what Prettier would output, and to remove usage of `Yoga.UNDEFINED` which doesn't existi and just resolves to `undefined` (this is converted to NaN by the wrapper layer).

Differential Revision: D45723003


nicoburns:
@NickGerleman This is awesome. Does this mean that https://github.com/facebook/yoga/pull/1194 is now unblocked? I have some work in progress to export Taffy's test fixtures to Yoga. I'll try to get that into a publishable state, but it might make sense to merge #1194 first, as Taffy has already imported versions of those fixtures and they would therefore be included in an export.

Issue #850, opened 2019-01-07
absolute child element percent length of left/top/right/bottom/width/height calcuated different compared with web


qcclcr:
# Report

- [x] I have searched [existing issues](https://github.com/facebook/yoga/issues) and this is not a duplicate

# Issues and Steps to Reproduce
   I found in yoga when absolute child element has percent length, it calculate base on parent‘s contentEdge but not padding Edge（which should be， https://www.w3.org/TR/CSS2/box.html#padding-edge
https://www.w3.org/TR/CSS2/visuren.html#propdef-left）。

  

As CSS defines：

Left，Like 'top', but specifies how far a box's left margin edge is offset to the right of the left edge of the box's **“containing block”**. For relatively positioned boxes, the offset is with respect to the left edge of the box itself.

10.1 Definition of "containing block"
The position and size of an element's box(es) are sometimes calculated relative to a certain rectangle, called the containing block of the element. The containing block of an element is defined as follows:

    1、The containing block in which the root element lives is a rectangle called the initial containing block. For continuous media, it has the dimensions of the viewport and is anchored at the canvas origin; it is the page area for paged media. The 'direction' property of the initial containing block is the same as for the root element.
    2、For other elements, if the element's position is 'relative' or 'static', the containing block is formed by the content edge of the nearest block container ancestor box.
    3、If the element has 'position: fixed', the containing block is established by the viewport in the case of continuous media or the page area in the case of paged media.
    4、**If the element has 'position: absolute'**, the containing block is established by the nearest ancestor with a 'position' of 'absolute', 'relative' or 'fixed', in the following way:
        a、In the case that the ancestor is an inline element, the containing block is the bounding box 
              around the padding boxes of the first and the last inline boxes generated for that element. In 
              CSS 2.1, if the inline element is split across multiple lines, the containing block is undefined.
        b、Otherwise, the containing block is formed by the **padding edge of the ancestor**.
        If there is no such ancestor, the containing block is the initial containing block.

# Expected Behavior
height of child is expected to be 1000px * 50% = 500px; and in web it is 500px

# Actual Behavior
but in yoga it's value is (1000px - 200px -200px) * 50% = 300px

# Link to Code
First Demo, if child is absolute , it is wrong
```javascript
<html>
<body>
  <div class="heightWidthPaddingVertical" style="position: relative; box-sizing:border-box; flex-direction:column; align-items:center; justify-content:center; width:750px; height:1000px; padding-top:200px;
[...]

PR #1290, opened 2023-05-11
Add errata "PercentAbsoluteOmitsPadding"

NickGerleman:
Summary:
This change default-enables a bugfix to Yoga's handling of absolute-positioned children sized using percentages. Yoga previously measured percentage of the box without padding, where the W3C spec, and browser behavior verified by test, use the padding box instead. See https://github.com/facebook/yoga/issues/850 for more details of the backing issue.

I have so far left this behind a `YGExperimentalFeature`, because it will change the dimensions of existing views which use absolute percentages with parent padding.

This change moves it from a `YGExperimentalFeature` to `YGErrata`. This means the conformant path is enabled by default, but users of `YGErrataClassic` and `YGErrataAll` will stay on the previous behavior.

Right now the name is `PercentAbsoluteOmitsPadding` as something that tries to explain the issue in lay-terms and be a little less long than the old one, but I am open to other ideas.

I also fixed a reversal of `height` and `width` in one place compared to the original PR, and enabled a significantly more edge-casey fix without the errata.

Differential Revision: D45763574


nicoburns:
Could we mark the test cases that are affected by this flag with a data attribute or similar (even if it doesn't actually do anything yet)? Seems that it would be useful to build up a test suite that explicitly tests the quirk modes. I guess ideally we would also include assertion values for those tests too (seeing as we can't just generate the quirky values using Chrome).

Also, TypeScript tests seem to include unrelated changes to formatting / variable naming...

NickGerleman:
> Seems that it would be useful to build up a test suite that explicitly tests the quirk modes.

Let me take a look, the original PR made just added a test for the new case, but will see if I can add one for the old.

> Also, TypeScript tests seem to include unrelated changes to formatting / variable naming...

This is a quirk of how changes made in Phabriactor are exported to GitHub. Each "commit" here is separately merged/managed internally, but depends on the previous commits. I would recommend looking only at the latest commit, and using the others as context (they had separate reviews).

PR #1301, opened 2023-05-25
Update version in Yoga.podspec

tkim-castlabs:
In `v1.19.0`, the Xcode compilation error was fixed. 
https://github.com/facebook/yoga/commit/f174de70afdde2492e8677bd0e716eb41bf64469

So `spec.version` should be updated.

### DIFF
--- a/Yoga.podspec
+++ b/Yoga.podspec
@@ -6,7 +6,7 @@
 
 Pod::Spec.new do |spec|
   spec.name = 'Yoga'
-  spec.version = '1.14.0'
+  spec.version = '1.19.0'
   spec.license =  { :type => 'MIT', :file => "LICENSE" }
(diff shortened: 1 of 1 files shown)


cipolleschi:
Hi @tkim-castlabs! Thank you so much for this contribution. However, this is not required. We are not publishing Yoga to Cocoapods anymore. We are shipping it with the React-Native framework, building it from source.

This change will not affect anything in the build system, unfortunately.

If you are observing the linked error, I suggest you to update your app as we fixed the issue in recent versions of the Framework.

NickGerleman:
Adding some more context, the above is true for React Native, which does have its own podspec, but builds Yoga locally.

We are hoping to do a proper OSS release of Yoga soon, but that will be bumping to a new version.

PR #1306, opened 2023-06-12
Mark YogaKit Podspec as Deprecated

NickGerleman:
Summary:
We are deprecating YogaKit as part of Yoga 2.0, with that version planned as the last release. Mark the podspec as deprecated.

We are also deprecating the Android ViewGroup in a similar way but from the googling I did there is not an equivalent way to do that for Maven artifacts.

Differential Revision: D46663201

### DIFF
--- a/YogaKit.podspec
+++ b/YogaKit.podspec
@@ -9,6 +9,7 @@ podspec = Pod::Spec.new do |spec|
   spec.license =  { :type => 'MIT', :file => "LICENSE" }
   spec.homepage = 'https://facebook.github.io/yoga/'
   spec.documentation_url = 'https://facebook.github.io/yoga/docs/'
+  spec.deprecated = true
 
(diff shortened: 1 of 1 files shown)


rtPag:
Can you already share what the replacement for YogaKit will be?

NickGerleman:
> Can you already share what the replacement for YogaKit will be?

We aren't currently planning a direct replacement. New experiences at Meta tend to be built with higher level frameworks over UIKit, like ComponentKit, or React Native.

Because we aren't spending the time to develop it further, the plan is to release YogaKit in its current state alongside the new Yoga bits in the 2.0 release we are doing later this month, but mark it deprecated. The last source revision will be present in 2.0 branch we will make, but we will remove it from the main branch after.

The revision in the branch should facilitate anyone who wants to fork the project, and I think there have been projects where folks have forked to add things like AppKit support. Once out of the Yoga repo, I hope we can encourage future development in one of those, instead of squatting on it to not do anything with it.

PR #1337, opened 2023-07-29
Allow negative padding (I known CSS rule is not allowed)

cntrump:
Negative padding support is an useful feature.

The padding of the parent element can offset the margin of the child element:

```
+----------++-----------+     Grid
|          ||           | padding: -2pt
|          ||           |
|    A     ||     B     |
|          ||           |    Item
|          ||           | margin: 2pt
+----------++-----------+
+----------++-----------+
|          ||           |
|          ||           |
|    C     ||     D     |
|          ||           |
|          ||           |
+----------++-----------+
```

### DIFF
--- a/yoga/YGConfig.cpp
+++ b/yoga/YGConfig.cpp
@@ -38,6 +38,14 @@ bool YGConfig::shouldPrintTree() const {
   return flags_.printTree;
 }
 
+void YGConfig::setAllowNegativePadding(bool allowNegativePadding) {
+  flags_.allowNegativePadding = allowNegativePadding;
+}
+
+bool YGConfig::allowNegativePadding() const {
+  return flags_.allowNegativePadding;
+}
+
 void YGConfig::setExperimentalFeatureEnabled(
     YGExperimentalFeature feature,
     bool enabled) {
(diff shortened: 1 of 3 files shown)


NickGerleman:
In this example, could we achieve the same thing by using 2pt gap instead of margin?

cntrump:
> In this example, could we achieve the same thing by using 2pt gap instead of margin?

How do I set child elements to have the same size ?

It is not work for me: root gap is 4 and child item width is 50%

NickGerleman:
https://snack.expo.dev/rnqKLMHqq

Red boxes touch the edge of their parent container (no margin to offset), but there is consistent spacing between items. 

```jsx
<View style={styles.exampleContainer}>
  <View style={styles.box} />
  <View style={styles.box} />
  <View style={styles.box} />
  <View style={styles.box} />
  <View style={styles.box} />
  <View style={styles.box} />
</View>

const styles = {
  exampleContainer: {
    backgroundColor: 'white',
    width: 250,
    gap: 4,
    flexDirection: 'row',
    flexWrap: 'wrap',
  },
  box: {
    backgroundColor: 'red',
    height: 50,
    width: '20%',
  }
};
```

image

PR #1339, opened 2023-08-01
Add SwiftPM support

cntrump:
Using yoga in Swift

```swift
import UIKit
import yoga

let YGUndefined = Float.nan

UIGraphicsBeginImageContextWithOptions(CGSize(width: 1, height: 1), true, 0)
let scale = UIGraphicsGetCurrentContext()?.ctm.a ?? 1
UIGraphicsEndImageContext()

let config = YGConfigNew()
YGConfigSetUseWebDefaults(config, true)
YGConfigSetPointScaleFactor(config, Float(scale))

let root = YGNodeNewWithConfig(config)
YGNodeStyleSetFlexDirection(root, YGFlexDirectionRow)
YGNodeStyleSetFlexWrap(root, YGWrapWrap)
YGNodeStyleSetWidth(root, 200)

for i in 0..<4 {
    let leaf = YGNodeNewWithConfig(config)
    YGNodeStyleSetAspectRatio(leaf, 1)
    YGNodeStyleSetWidthPercent(leaf, 50)
    YGNodeInsertChild(root, leaf, UInt32(i))
}

YGNodeCalculateLayout(root, YGUndefined, YGUndefined, YGNodeStyleGetDirection(root))

for i in 0..<4 {
    let leaf = YGNodeGetChild(root, UInt32(i))
    let left = YGNodeLayoutGetLeft(leaf)
    let top = YGNodeLayoutGetTop(leaf)
    let width = YGNodeLayoutGetWidth(leaf)
    let height = YGNodeLayoutGetHeight(leaf)

    print("leaf \(i): origin: \(left), \(top), size: \(width), \(height)")
}

let width = YGNodeLayoutGetWidth(root)
let height = YGNodeLayoutGetHeight(root)

print("root size: \(width), \(height)")

YGNodeFreeRecursive(root)
YGConfigFree(config)
```

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,0 +1,32 @@
+// swift-tools-version:5.0
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import PackageDescription
+
+let package = Package(
+    name: "yoga",
+    products: [
+        .library(name: "yoga", targets: [ "C" ])
+    ],
+    targets: [
+        .target(
+            name: "C",
+            path: ".",
+            sources: [
+                "yoga"
+            ],
+            publicHeadersPath: ".",
+            cxxSettings: [
+                .headerSearchPath(".")
+            ]
+        )
+    ],
+    cxxLanguageStandard: CXXLanguageStandard(rawValue: "c++17")
+)
--- a/Package.swift
+++ b/Package.swift
@@ -13,11 +13,11 @@ import PackageDescription
 let package = Package(
     name: "yoga",
     products: [
-        .library(name: "yoga", targets: [ "C" ])
+        .library(name: "yoga", targets: [ "core" ])
     ],
     targets: [
         .target(
-            name: "C",
+            name: "core",
             path: ".",
             sources: [
                 "yoga"
--- a/.github/workflows/validate-swiftpm.yml
+++ b/.github/workflows/validate-swiftpm.yml
@@ -1,0 +1,22 @@
+name: Swift
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+jobs:
+  test:
+    name: Build
+    runs-on: macos-latest
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Build Debug
+        run: swift build -c debug
+
+      - name: Build Release
+        run: swift build -c release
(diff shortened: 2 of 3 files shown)


yoga/BitUtils.h
```diff
@@ -11,7 +11,7 @@
 #include <cstdio>
 #include <cstdint>
 
-#include <yoga/YGEnums.h>
+#include "YGEnums.h"
```
NickGerleman, line 14: Meta's Buck build is set up so that public headers `{Yoga,YGEnums,YGMacros,YGValue}.h` must be accessed using `yoga/` [header namespace](https://buck.build/rule/cxx_library.html#exported_headers).

So we unfortunately cannot access them as relative paths, or via traversal.

It is a bit verbose, and maybe a bit breaking, but an option I had looked at was to fully move files like `yoga/Yoga.h` to `yoga/include/yoga/Yoga.h`.

Package.swift
```diff
@@ -0,0 +1,24 @@
...
+        .library(name: "yoga", targets: [ "yoga" ])
+    ],
+    targets: [
+        .target(
+            name: "yoga",
+            path: "yoga",
+            publicHeadersPath: "include"
+        )
+    ],
+    cxxLanguageStandard: CXXLanguageStandard(rawValue: "gnu++17")
```
NickGerleman, line 23: Could we keep in sync with https://github.com/facebook/yoga/blob/main/Yoga.podspec?

Its min flags are roughly in sync with the range of devices that Meta is testing and using.

Package.swift
```diff
@@ -0,0 +1,24 @@
+// swift-tools-version:5.0
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let package = Package(
```
NickGerleman, line 6: Could we add validation to GitHub Actions for this? E.g. for CocoaPods build, `pod spec lint` will build the pod and check the podspec is valid.

yoga/include/YGEnums.h
```diff
@@ -0,0 +1 @@
+../YGEnums.h
```
NickGerleman, line 1: New files need the copyright header present in the other header files.

NickGerleman (commented):
Thank you for making this PR 😃.

I left a couple comments related to C++ header organization changes. I think we should continue using header namespace for public headers, within Yoga.

cntrump:
I found a simple way to support SwiftPM, PR updated.

yoga/BitUtils.h
cntrump: OK, I see that. Now I just add tow new files to support SwiftPM.

Package.swift
cntrump: Changed to `c++17`.

yoga/include/YGEnums.h
cntrump: OK, copyright added to new files.

Package.swift
cntrump: Added validate-swiftpm.yml

NickGerleman (approved):

PR #1342, opened 2023-08-09
Update standalone JS docs

rolandcrosby:
Excited to see some activity around this project! The move from Flow to TypeScript broke one of the links in the documentation, which this PR fixes. I went ahead and updated the example code as well, since it looks like some things moved around since the switch to the async loader.


NickGerleman (approved):
Thank you for updating this. Yoga’s website is in need of quite a bit of love right now.

PR #1359, opened 2023-09-05
Enable `-Wconversion`

NickGerleman:
Summary:
This enables clang warnings around potentially unsafe conversions, such as those with mismatched signedness, or ones which may lead to truncation.

This should catch issues in local development which create errors for MSVC (e.g. Dash), who's default `/W3` includes warnings akin to `-Wshorten-64-to-32`.

This full set of warnings here is a tad spammy, but probably more useful than not.

Changelog: [Internal]

Reviewed By: yungsters

Differential Revision: D48954777


yoga/node/LayoutResults.h
```diff
@@ -48,17 +53,15 @@ struct LayoutResults {
   CachedMeasurement cachedLayout{};
 
   YGDirection direction() const {
-    return getEnumData<YGDirection>(flags, directionOffset);
+    return static_cast<YGDirection>(flags_.direction);
   }
 
   void setDirection(YGDirection direction) {
-    setEnumData<YGDirection>(flags, directionOffset, direction);
+    flags_.direction = static_cast<uint8_t>(direction);
```
NickGerleman, line 60:
```suggestion
    flags_.direction = static_cast<uint8_t>(direction) & 0x03;
```

Issue #1324, opened 2023-07-09
fix Yoga.cpp for Xcode14.3


McAlexDudu:
Hi, when I update to Xcode14.3, and the Swift is 5.8, some gramma seems to not work(as follow). I replace '|' to "||" seems to work, so hope to check out the problem. Thanks a lot!
<img width="944" alt="截屏2023-07-09 22 27 11" src="https://github.com/facebook/yoga/assets/16278257/df90578f-87fe-4972-9fff-1ceda26c8334">

NickGerleman:
This was fixed by https://github.com/facebook/yoga/commit/f174de70afdde2492e8677bd0e716eb41bf64469

This change is available in the Yoga 2.0 beta packages we just published, or by source. If you are seeing this because you are using an older version of React Native, please upgrade to 0.69 or later.

PR #1363, opened 2023-09-09
Fixed minor syntax error

hetanthakkar:
I've corrected the operator syntax error. This error disrupts the execution of command: 'npx react-native run-ios.'

Screenshot 2023-09-08 at 8 04 49 PM


NickGerleman:
See https://github.com/facebook/yoga/issues/1324

The issue you are seeing was actually fixed years ago in Yoga, and you are seeing it because you are using a new version of XCode with an older version of React Native.

This is still fine to fix, but is not what is causing your error.

PR #1391, opened 2023-09-14
Add linker flags to workaround iOS 14 runtime crash

### LINKED ISSUE #1390: EXC_BAD_ACCESS in `new yoga::Config(getDefaultLogger())` when building for iOS 14 using Xcode 15
# Report

- Build Yoga 2.0 (or 1.19, same issue) in Xcode 15, targeting a iOS 14.5 Simulator
- Run the project

# Expected Behavior

Should not crash

# Actual Behavior

Crashes

image

# Stack

```
Thread 1 Queue : com.apple.main-thread (serial)
#0	0x0000000000000000 in 0x00000000 ()
#1	0x0000000108bed2d8 in ::YGConfigNew() at /Users/<redacted>/SourcePackages/checkouts/yoga/yoga/Yoga.cpp:182
```

# Other

- No issues when building with Xcode 14.2, this is a new issue with the Xcode 15 toolchain.

- No issues when building for iOS 15 or higher.

clausjoergensen:
Resolves #1390

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -27,6 +27,8 @@ let package = Package(
                 .headerSearchPath(".")
             ],
             linkerSettings: [
+                // Binaries using symbols with a weak definition crash at runtime on iOS 14/macOS 12 or older
+                // when using the Xcode 15 toolchain (FB13097713)
                 .unsafeFlags(["-Wl", "-ld_classic"])
             ]
         )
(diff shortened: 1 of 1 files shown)


NickGerleman (approved):

cortinico (approved):

clausjoergensen:
I actually found another issue here. Turns out `-ld_classic` isn't available in older versions of Xcode, so this isn't safe to ship as part of the Package.swift :(

Instead consumers of Yoga should add this themselves if they use Xcode 15 and support iOS 14 and below.

PR #1408, opened 2023-09-28
Fixed issue with first line element gap handling.

jlmip:
If the first element of a line is not contributing (e.g. position absolute), an additional gap will be added to the line, because the first gap element of the line is never identified (wrong start index). 
Fix: raise the index of the first line element until we find an element that is contributing to the line.


NickGerleman:
Thanks for the contribution. Could you add a test case to YGGapTest.html?

PR #1422, opened 2023-10-13
Support "align-content: space-evenly"

nicoburns:
### Changes made
- Regenerated tests (as some aspect ratio tests seem to be out of date compared to the fixtures)
- Added SpaceEvenly variant to the "Align" enums (via enums.py)
- Implemented `align-content: space-evenly` alignment in CalculateLayout.cpp
- Added generated tests `align-content: space-evenly`
- Updated NumericBitfield test to account for the fact that the Align enum now requires more bits (this bit could do with being reviewed as I am not 100% certain that it's valid to just update the test like this).

### Changes not made
- Any attempt to improve the spec-compliance of content alignment in general (e.g. I think https://github.com/facebook/yoga/pull/1013 probably still needs to happen)

### DIFF
--- a/gentest/fixtures/YGAlignContentTest.html
+++ b/gentest/fixtures/YGAlignContentTest.html
@@ -1,9 +1,38 @@
-<div id="align_content_flex_start" style="width: 130px; height: 100px; flex-wrap: wrap; flex-direction: row; align-content: flex-start;">
+<!-- ALIGN CONTENT: FLEX-START -->
+
+<div id="align_content_flex_start_nowrap" style="width: 140px; height: 120px; flex-direction: row; align-content: flex-start;">
+  <div style="width: 50px; height: 10px;"></div>
+  <div style="width: 50px; height: 10px;"></div>
+</div>
+
+<div id="align_content_flex_start_wrap" style="width: 140px; height: 120px; flex-wrap: wrap; flex-direction: row; align-content: flex-start;">
   <div style="width: 50px; height: 10px;"></div>
   <div style="width: 50px; height: 10px;"></div>
   <div style="width: 50px; height: 10px;"></div>
   <div style="width: 50px; height: 10px;"></div>
   <div style="width: 50px; height: 10px;"></div>
+</div>
+
+<!-- This tests the case where `flex-wrap: wrap` but the content does not actually wrap -->
+<div id="align_content_flex_start_wrap_singleline" style="width: 140px; height: 120px; flex-wrap: wrap; flex-direction: row; align-content: flex-start;">
+  <div style="width: 50px; height: 10px;"></div>
+  <div style="width: 50px; height: 10px;"></div>
+</div>
+
+<div id="align_content_flex_start_wrapped_negative_space" style="display: flex; flex-direction: column; width: 320px; height: 320px; border-width: 60px;">
+  <div style="display: flex; flex-direction: row; flex-wrap: wrap; align-content: flex-start; justify-content: center; height: 10px;">
+    <div style="width: 80%; height: 20px; flex-shrink: 0;"></div>
+    <div style="width: 80%; height: 20px; flex-shrink: 0;"></div>
+    <div style="width: 80%; height: 20px; flex-shrink: 0;"></div>
+  </div>
+</div>
+
+<div id="align_content_flex_start_wrapped_negative_space_gap" style="display: flex; flex-direction: column; width: 320px; height: 320px; border-width: 60px;">
+  <div style="display: flex; flex-direction: row; flex-wrap: wrap; align-content: flex-start; justify-content: center; height: 10px; gap: 10px;">
+    <div style="width: 80%; height: 20px; flex-shrink: 0;"></div>
+    <div style="width: 80%; height: 20px; flex-shrink: 0;"></div>
+    <div style="width: 80%; height: 20px; flex-shrink: 0;"></div>
+  </div>
 </div>
 
 <div id="align_content_flex_start_without_height_on_children" style="width: 100px; height: 100px; flex-wrap: wrap; flex-direction: column; align-content: flex-start;">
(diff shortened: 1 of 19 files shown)


NickGerleman:
> Updated NumericBitfield test to account for the fact that the Align enum now requires more bits (this bit could do with being reviewed as I am not 100% certain that it's valid to just update the test like this).

This should be okay. In `Style.h`, `minimumBitCount()` is used to avoid hardcoding. We could do the same in the tests, but I'm in the process of refactoring to remove usage of `NumericBitfield` anyway (plain old bitfields are a lot easier to reason about when debugging or making changes).

tests/generated/YGAspectRatioTest.cpp
```diff
@@ -80,28 +80,28 @@ TEST(YogaTest, aspect_ratio_does_not_stretch_cross_axis_dim) {
 
   ASSERT_FLOAT_EQ(0, YGNodeLayoutGetLeft(root_child0_child0));
   ASSERT_FLOAT_EQ(0, YGNodeLayoutGetTop(root_child0_child0));
-  ASSERT_FLOAT_EQ(300, YGNodeLayoutGetWidth(root_child0_child0));
-  ASSERT_FLOAT_EQ(197, YGNodeLayoutGetHeight(root_child0_child0));
+  ASSERT_FLOAT_EQ(285, YGNodeLayoutGetWidth(root_child0_child0));
+  ASSERT_FLOAT_EQ(187, YGNodeLayoutGetHeight(root_child0_child0));
```
NickGerleman, line 84: > Regenerated tests (as some aspect ratio tests seem to be out of date compared to the fixtures)

@rozele added this specific fixture yesterday with https://github.com/facebook/yoga/commit/50ecd981414b85e786ffec8110154dd2caf82a9e.

It it disabled right now regardless, but I am curious where the difference comes from. @rozele @nicoburns do you know what Chrome versions/platforms created these results?

yoga/algorithm/CalculateLayout.cpp
```diff
@@ -2021,6 +2021,18 @@ static void calculateLayoutImpl(
...
+            currentLead +=
+                remainingAlignContentDim / static_cast<float>(lineCount + 1);
+            if (lineCount > 1) {
+              crossDimLead =
+                  remainingAlignContentDim / static_cast<float>(lineCount + 1);
+            }
+          } else {
+            currentLead += remainingAlignContentDim / 2;
+          }
+          break;
```
NickGerleman, line 2035: Am I following this logic correctly?

`currentLead` is the leading position of the current line (incremented by each line height), then `crossDimLead` is the per-line lead added to each line after.

When there is not free space left, we fall back to centering. This is the specified behavior in the [flexbox spec](https://www.w3.org/TR/css-flexbox-1/#valdef-align-content-space-around) for `space-around`, where the later [box alignment](https://www.w3.org/TR/css-align-3/#distribution-values) spec instead says `space-around` and `space-evenly` have a fallback alignment of "safe center"?

NickGerleman: > e.g. I think https://github.com/facebook/yoga/pull/1013 probably still needs to happen)

Yuck, I see. We are contributing `crossDimLead` to not just distribution of the line, but then also using it for alignment within the line.

Is there a reason for `if (lineCount > 1)` here? Or was it just around for `Align::SpaceAround` to avoid division by zero, if the lead will never be used after the next line (ignoring the bug where it contributes to `align-items`).

gentest/fixtures/YGAlignContentTest.html
```diff
@@ -54,6 +54,25 @@
...
+<div id="align_content_spaceevenly_wrap_singleline" style="width: 140px; height: 120px; flex-wrap: wrap; flex-direction: row; align-content: space-evenly;">
+  <div style="width: 50px; height: 10px;"></div>
+  <div style="width: 50px; height: 10px;"></div>
+</div>
+
+<div id="align_content_spaceevenly_nowrap" style="width: 140px; height: 120px; flex-direction: row; align-content: space-evenly;">
+  <div style="width: 50px; height: 10px;"></div>
+  <div style="width: 50px; height: 10px;"></div>
+</div>
+
```
NickGerleman, line 75: Could we also add a test for when we have negative free space left, since that has special logic?

NickGerleman (commented):
Great job, and great contribution 🙂. I left a couple quick comments/questions, but this otherwise looks solid.

tests/generated/YGAspectRatioTest.cpp
nicoburns: I'm getting these results on Chrome 117 and 118 (just updated) on  macOS (arm64)

gentest/fixtures/YGAlignContentTest.html
nicoburns: Ah, yes. I have imported Taffy's tests for these cases (we also have one that tests with negative space and `gap`), and for the same cases for the other align-content modes.

yoga/algorithm/CalculateLayout.cpp
nicoburns: > currentLead is the leading position of the current line (incremented by each line height), then crossDimLead is the per-line lead added to each line after.

I think that's correct. Although, I believe the effect of the code adding to `currentLead` here is to set the "lead" for the *first* line.

> Is there a reason for if (lineCount > 1) here?

I don't think so. I just left it there on a principle of not wanting to touch anything. I think you're probably right about it being there to avoid a divide by zero. Although it seems like it might be better not to run alignment at all if there are zero lines (Taffy actually skips the entire Flexbox algorithm if there are no children as it reduces to relatively trivial calculations involving the container node's styles)

> When there is not free space left, we fall back to centering. 

Yep

> This is the specified behavior in the [flexbox spec](https://www.w3.org/TR/css-flexbox-1/#valdef-align-content-space-around) for space-around, where the later [box alignment](https://www.w3.org/TR/css-align-3/#distribution-values) spec instead says space-around and space-evenly have a fallback alignment of "safe center"?

I believe the "safe" alignment modes are opt-in. So these would probably be separate enum variants (or a boolean on top of the alignment) in Yoga.

NickGerleman (approved):

tests/generated/YGAspectRatioTest.cpp
rozele: > I'm getting these results on Chrome 117 and 118 (just updated) on macOS (arm64)

I just pulled down this PR and re-ran gentest, it's reverting these changed.

Chrome - Version 118.0.5993.70 (Official Build) (arm64)

NickGerleman: I get the same as the changes this PR makes (arm64 macOS 117)... Something's odd here.

NickGerleman: Maybe chrome running an A/B test?

nicoburns: So I pasted the test HTML into the template and opened it up in Chrome, and I think I know what's going here. It's because there is a div with `overflow: scroll`, and I have set my OS to "always show scrollbars", whereas @rozele, I'm assuming you have it set to one of the settings where scrollbars don't take up space. If I re-run the gentests with scrollbars set to overlay scrollbars then I get the same results as comitted.

nicoburns: Interestingly, the *reason* I have my OS set to "always show scrollbars" is because:

- Taffy now takes the space taken up by scrollbars when `overflow: scroll` is set
- The only way to test that properly using gentests is to ensure that scrollbars are taking up space when you generate the tests
- So Taffy's gentest runner now has an assert to check for that and refuses to run if you don't have space-taking scrollbars enabled.

I'm guessing Yoga doesn't reserve space for scrollbars? (which kinda makes sense given that it's mobile focussed - although I imagine it could cause issues for react-native-windows)

Want to me to remove that commit from this PR?

rozele: Hmm, maybe that's why `overflow: scroll` wasn't supported in gentest before my diff 😅

I'd hate to say we shouldn't use `overflow: scroll` in fixtures arbitrarily, especially since they can certainly be a source of bugs, but maybe having fixtures for them is a bad idea (unless there's a way we can make gentest detect the scrollbar setting and warn / switch to the "correct" default before running)

NickGerleman: Feel free to keep it in. We can try to figure out the right solution for this after (I think we might be able to avoid the ambiguity in the fixture).

On reserving space for scrollbar on non-mobile, I think WinUI default has shy scrollbars that don't take up space until hovered over, and then they go over content, https://learn.microsoft.com/en-us/windows/apps/design/controls/scroll-controls

I think in the other cases, Scrollbar being present shrinks the available space of the layout, but isn't handled explicitly by Yoga.

NickGerleman: Edit: looks like there is a `--hide-scrollbars` Chrome option we can pass to Chromedriver, to make this deterministic, and match existing behavior.

nicoburns: > unless there's a way we can make gentest detect the scrollbar setting and warn / switch to the "correct" default before running

This is definitely possible (you can test for the setting in the browser by creating a node with `overflow: scroll` set and checking for it's width), but it might require a bit of work to thread the error message back through to the ruby test runner script so that it could be displayed nicely to the user.

I definitely think it would be good to this kind of error reporting improvements at some point, as currently the gentest script has a nasty habit of just silently crashing (sometimes having written half the tests!) if anything goes wrong (like an unexpected property value).

nicoburns: > Edit: looks like there is a `--hide-scrollbars` Chrome option we can pass to Chromedriver, to make this deterministic, and match existing behavior. I tacked this onto the imported PR for now.

I guess that works. I think I dismissed that for Taffy because I wanted the opposite: "force show scrollbars", and that doesn't seem to exist.

PR #1424, opened 2023-10-16
Rename ambiguous getLeading/Trailing... functions in Node.cpp

joevilches:
Summary: See D50344874

Differential Revision: D50344874


nicoburns:
A huge +1 for this change. I can read those new names and know exactly what they refer to. Whereas the old ones were indeed ambiguous to me.

PR #1430, opened 2023-10-17
Pass '--hide-scrollbars' flag to chrome when generating tests

nicoburns:
I have verified that this works (both that it doesn't crash, and that it causes me to generate the same assertions as everyone else, even with the "always show scrollbars" OS setting enabled).

@NickGerleman If I was following a "classic OSS workflow" I would rebase my space-evenly PR on top of this one and remove the commit updating the aspect ratio tests from this PR. Will Phabricator / the sync scripts deal with that properly, or would it be easier for me to just put this PR "on top" of the other one and just have a commit reverting the test changes? I can see you've already imported the other PR, and have been building work on top of it.

### DIFF
--- a/gentest/gentest.rb
+++ b/gentest/gentest.rb
@@ -14,7 +14,7 @@ browser = Watir::Browser.new(:chrome, options: {
     "browser" => "ALL",
     "performance" => "ALL"
   },
-  args: ['--force-device-scale-factor=1', '--window-position=0,0']
+  args: ['--force-device-scale-factor=1', '--hide-scrollbars', '--window-position=0,0']
 })
 
 Dir.chdir(File.dirname($0))


NickGerleman:
> @NickGerleman If I was following a "classic OSS workflow" I would rebase my space-evenly PR on top of this one and remove the commit updating the aspect ratio tests from this PR. Will Phabricator / the sync scripts deal with that properly, or would it be easier for me to just put this PR "on top" of the other one and just have a commit reverting the test changes? I can see you've already imported the other PR, and have been building work on top of it.

I think either should work. The import flow looks like commits from GitHub PRs being replicated, but targeting the different libraries in the FB monorepo instead of the OSS repo. Then we enforce that changes which effect the source of certain projects (e.g. RN in OS mirrors Yoga) must be exported back to GitHub, to run RN's OSS CI as well.

As part of building everything in the monorepo, I actually needing to tack on a spanning RN change to the commit, where RN switches on Yoga align value, and we need to add the new case at the same time as the enum to avoid warnings. This looks like making a change on top of the internal commit, then the system, when it merges, will push the commit to GitHub, with the same original author (including any edits).

I so far amended this change on top of yours, and re-ran generation. If you are okay with that, we could commit it as part of your change.

NickGerleman:
Was merged inline with https://github.com/facebook/yoga/commit/0d28b283e2c9931fe9f73f737c8fd9e58f922876#diff-54d80dd80005f12c993c43e178f20b79ef64a86ae8712e1e10f6c409a1dd7506

PR #1434, opened 2023-10-20
Add errata supporting changes to position: static

joevilches:
Summary:
X-link: https://github.com/facebook/react-native/pull/41130

I will use this errata to gate my changes that actually make position: static behave like the web. We have future plans to make position: relative the default again but users could still have declared certain nodes as position: static, so I think this is needed regardless.

Differential Revision: D50506915


NickGerleman:
FYI @nicoburns in case Taffy wants to support `position: "static"` in the future (esp https://github.com/facebook/yoga/pull/1434/commits/2f42c9d33e709e686bb508ed7058c3b0acac117a which adds fixtures).

nicoburns:
Oo, yeah this has been requested (https://github.com/DioxusLabs/taffy/issues/212), and is definitely needed if one wants to render a "default web div".

@joevilches I'd be interested in what your implementation strategy for this is. In particular, have you thought about:

-  How this interacts with caching? In particular, in the incremental relayout case where one of the node's `position` styles has changed.
- Whether there needs to be some kind of output of parent-child linkage between containing blocks and their non-direct-child contained nodes (for consumers of Yoga/Taffy)?

Also, I think some `px` tests would be good.

NickGerleman:
Caching is a good callout. Right now some of the refactoring is to pass containing block when recursing during layout, instead of just parent dimensions. Containing block dimensions might need to be added as cache keys, like owner size is right now.

nicoburns:
> Caching is a good callout... Containing block dimensions might need to be added as cache keys, like owner size is right now.

Yeah, that would probably work. Actually, I figured that what we call "parent size" in Taffy might be called "owner size" in Yoga precisely because for this use case. I *think* that the containing block dimensions could probably be used *instead* of the "parent size" in the cache key (bearing in mind that for non-absolutely positioned nodes the parent is the containing block). Which would be nice, to avoid bloating the cache size.

NickGerleman:
The history of “owner” is in this commit: https://github.com/facebook/yoga/commit/f0edefdbb741f22aad77551dcedef59baab7a63f

The history here is that the React Native Fabric renderer keeps a Yoga node per ShadowNode, but Fabric ShadowNodes are immutable after being laid out. https://reactnative.dev/architecture/render-pipeline

This clashes with Yoga’s existing public API, which is oriented around nodes being mutated throughout their lifetime. Fabric took a dependency on Yoga internals, to build this different model, with some of it making its way into Yoga itself. I’ve been thinking about how we can move that boundary into something public in the future.

Out of probably a couple dozen of Yoga using frameworks in our monorepo, Fabric is the only one using the private APIs. All others are now set up to only have visibility to public headers.

nicoburns:
Hmm... a node being able to have more than one parent doesn't sound like a fantastic idea. I'm guessing this is to allow for some kind of copy-on-write semantics on the Fabric side? I'd need to look at the code, but off the top of my head it seems to me that:

- Either old trees are relayouted after a newer tree is created. Which sounds like it wouldn't be great for perf due to cache thrashing, and should perhaps lead to the Yoga node being cloned too.
- Or they're not, in which case I don't understand why old ShadowNodes would need to retain their references to a yoga nodes at all. The only reason I can think is to access to the computed layout. And I'm thinking maybe this could just be copied into the ShadowNode.

Long term, it would probably be good to make Yoga's algorithms generic over the node type they operate over, and make them operate directly on ShadowNodes in Fabric.

NickGerleman:
Copy-on-write is correct. Fabric treats a node owned by a different parent as one that cannot be mutated, so we lazily create new writable Yoga nodes during layout.

Layout information is copied to the ShadowNode, but original Yoga node needs to be cloned when creating new ShadowNode revision. This also clones the nodes layout cache. Then Fabric will dirty on node prop change. Shadow trees are allowed to be created on multiple threads, and multiple revisions can exist, derived from a single ancestor revision. 

There are a bunch of ideas on how to refactor this eventually.

joevilches:
@nicoburns Yeah what Nick said is the gist of it. Basically just plan to pass down containing block information in the calculate layout functions, haven't planned much else beyond that. Caching was not on my mind initially but I will keep an eye out for it after the discussion here. I plan on getting started with these things this week, we shall see how much progress I make :)

PR #1438, opened 2023-10-26
Backout "Add tests for absolute positioning of children with padding in the parent"

joevilches:
Summary:
Backing out D50670457
Github PR is: https://github.com/facebook/yoga/commit/4f98bfe40a72aed0d5888d9d5c5d9c2340da7f4f

That PR started failing our workflows due to a build error in gtest. It seems unlikely that it is due to that change but backing it out to see if it fixes that issue.

https://github.com/facebook/yoga/actions/runs/6657567207/job/18095847672

Differential Revision: D50708256


NickGerleman:
@joevilches I think this is likely due to a change in the ubuntu-latest image.

The error is happening in headers for GCC stdlibc++. The clang build should instead be using LLVM libc++.

I think we could try to follow the solution in https://stackoverflow.com/questions/76859275/error-compiling-a-cpp-containing-stdchrono-errorstatic-constexpr-unsigned-fra in project-defaults.cmake, conditionally when compiling with Clang.

NickGerleman:
Here’s some information I found about the stdlib included by default. https://stackoverflow.com/questions/19774778/when-is-it-necessary-to-use-the-flag-stdlib-libstdc

NickGerleman:
Looks tentatively like https://github.com/facebook/yoga/pull/1440 should fix.

PR #1440, opened 2023-10-27
Use libc++ in Ubuntu Clang Build

NickGerleman:
An ubuntu-latest image update caused build failures, where GTest importing chrono headers caused an error inside of stdlib++ when being compiled by Clang. 

This switches to explicitly asking for libc++ (LLVM stdlib) in the reference Ubuntu Clang build instead of stdlibc++ (GCC stdlib).

We don’t force this in CMake logic, to not force a specific stdlib for users compiling Yoga alongside other libraries. E.g. https://github.com/facebookexperimental/libunifex/issues/86


NickGerleman:
Actually this might be better to set as part of the environment, so that cmake users don’t get this as something kinda opinionated.

cortinico (approved):

PR #1446, opened 2023-11-02
Fix issue where absolute children of row-reverse containers would inset on the wrong side

joevilches:
Summary:
NickGerleman pointed out that my recent changes to fix the slew of row-reverse problems in Yoga actually ended up regressing some parts. Specifically, absolute children of row-reverse containers would have their insets set to the wrong side. So if you set left: 10 it would apply it to the right.

Turns out, in `layoutAbsoluteChild` there were cases where we were applying inlineStart/End values to the flexStart/End edge, which can never be right. So I changed the values to also be flexStart/End as the fix here.

Differential Revision: D50945475

### DIFF
--- a/gentest/fixtures/YGFlexDirectionTest.html
+++ b/gentest/fixtures/YGFlexDirectionTest.html
@@ -208,3 +208,195 @@
     <div style="width: 10px;"></div>
   </div>
 </div>
+
+<div id="flex_direction_row_reverse_inner_pos_left" style="height: 100px; width: 100px;">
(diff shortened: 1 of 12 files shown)


nicoburns:
One thing I've found to be super-helpful to get good test coverage of issues like this without having to create a million tests is to use uneven padding/border/margin/inset sizes. For examples `<div style="left: 1px; right: 2px; top: 3px; bottom: 4px;">`. If you are applying styles incorrectly then this test will fail, without you needing to create a separate test for each side.

PR #1472, opened 2023-11-14
Use containing block to adjust absolute child position

joevilches:
Summary:
This change has most of the logic needed for supporting `position: static`. We do two things here that fix a lot of the broken static test:

1) We pass in the containing node to `layoutAbsoluteChild` and use it to properly position the child in the case that insets are defined.
2) We rewrite the absolute child's position to be relative to it's parent in the event that insets are defined for that child (and thus it is positioned relative to its CB). Yoga's layout position has always be relative to parent, so I feel it is easier to just adjust the coordinates of a node to adhere to that design rather than change the consumers of yoga.

The "hard" part of this algorithm is determining how to iterate the offset from the containing block needed to do this translation described above. That is handled in `layoutAbsoluteDescendants`.

Differential Revision: D51224327

### DIFF
--- a/gentest/fixtures/YGStaticPositionTest.html
+++ b/gentest/fixtures/YGStaticPositionTest.html
@@ -1,11 +1,11 @@
 <!-- The top level divs in each test are needed so that each div overlays each
      other and the top measurement is accurate -->
-<div id="static_position_insets_have_no_effect_left_top" data-disabled="true">
+<div id="static_position_insets_have_no_effect_left_top">
     <div style="width: 100px; height: 100px; position: static; top: 50px; left: 50px;">
     </div>
 </div>
 
-<div id="static_position_insets_have_no_effect_right_bottom" data-disabled="true">
+<div id="static_position_insets_have_no_effect_right_bottom">
     <div style="width: 100px; height: 100px; position: static; bottom: 50px; right: 50px;">
     </div>
 </div>
--- a/java/tests/com/facebook/yoga/YGStaticPositionTest.java
+++ b/java/tests/com/facebook/yoga/YGStaticPositionTest.java
@@ -26,7 +26,6 @@ public class YGStaticPositionTest {
   @Parameterized.Parameter public TestParametrization.NodeFactory mNodeFactory;
 
   @Test
-  @Ignore
   public void test_static_position_insets_have_no_effect_left_top() {
     YogaConfig config = YogaConfigFactory.create();
     config.setExperimentalFeatureEnabled(YogaExperimentalFeature.ABSOLUTE_PERCENTAGE_AGAINST_PADDING_EDGE, true);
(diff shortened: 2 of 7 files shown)


nicoburns:
This approach mostly makes sense to me. The only thing I would comment is that you might want to consider how this code would end up being structured in a future where Yoga has multiple layout algorithms (Grid, Block, etc) rather than just Flexbox. In that case, an absolutely positioned node's "static position" (it's position when not determined by "inset") needs to be positioned according the layout algorithm of it's direct parent rather than that of it's containing block (and IIRC for Block layout there are differences in auto margins even when "inset" is set).

I'm wondering whether it would make sense to always have parents lay out their absolute children (even if they're not the containing block). And just keep track of the relevant containing block and the offset of the current node relative to it (pass then as input parameters the recursive layout function) as you recurse down the tree.

I think in the "open world" data model that Yoga has (where the entire tree is accessible at all times) then your existing approach would still work, it'd just need to switch on layout algorithm for each child which might potentially be a little awkward code organisation wise?

(Taffy has taken on an additional constraint that layout algorithm only access styles and set layouts for their direct children in order to make it easier to integrate with UI toolkits - which precludes the approach you are taking)

joevilches:
@nicoburns Yeah I have been thinking about how a lot of these changes will work if we ever tried to do different types of displays besides flex. There are a ton of cases in the code where the assumption of flex is baked into the logic, so it would be quite the mess implementing it. Right now, the `layoutAbsoluteChild` function does take into account the flexbox algorithm (which is mainly dependent on the style of the parent) even though it is the containing node that is actually calling that function. It is relying on the fact that we layout absolute children at the end of our own layout and that we layout in a depth-first fashion from the root. So it might not be too bad regardless of what layout algorithm we are using, but I bet there is still quite a bit of support needed.

 >just keep track of the relevant containing block and the offset of the current node relative to it (pass then as input parameters the recursive layout function) as you recurse down the tree

I actually did this at first and switched to this implementation. The issue we (me and @NickGerleman) discovered when doing this is that Yoga writes the layout position relative to the parent in all cases. So anyone consuming Yoga will be interpreting those values as offsets from parent. On web the layout position is relative to the containing block, so it is a bit different. We thought it would be too risky/big of a task to change how consumers of Yoga should interpret the layout position, so we need to ensure that we "convert" the position to be relative to the parent after applying insets. To do this, we either have 2 options:

1. After layout, iterate over the tree and convert specific nodes to be relative to their parent
2. What we did here - let the CB be the one who lays out its absolute descendants

We steered away from 1, which was the original solution I chose because it leaves some node's layout in this weird in-between state where the position is in the wrong "coordinate system" for some amount of time. If we wanted to not adjust the position after-the-fact, yet still have it so that parents layout their absolute children, we would run into a problem: the parent's position relative to *its* parent is not yet determined (and so on, up until the root). So there would be no way to do this conversion from position-relative-to-CB to position-relative-to-parent that we need. As a result we landed on option 2.

A bonus is caching seems to just work out of the box, which I believe would not be the case if we chose a different approach. Also the code to is much more contained to one place (`layoutAbsoluteChild`) instead of passing that information to every function that eventually calls `calculateLayoutInternal` at some point.

PR #1491, opened 2023-12-01
fix Container min-height produces incorrect size when flex-wrap is enabled

### LINKED ISSUE #1300: Container `min-height` produces incorrect size when `flex-wrap` is enabled
# Report

I use `yoga-layout` asm bindings.
When I Create a container node where `width` is exact, but only set `min-height`, and enable `flexWrap`:
```js
const container = Node.create()

container.setWidth(200)
container.setMinHeight(200)
container.setFlexWrap(WRAP_WRAP)
container.setFlexDirection(FLEX_DIRECTION_ROW)
```

and add child nodes so that it goes to the next line.

```js
const node1 = Node.create()
const node2 = Node.create()

node1.setWidth(120)
node1.setHeight(120)
node2.setWidth(120)
node2.setHeight(120)

container.insertChild(node1, 0)
container.insertChild(node2, 0)

container.calculateLayout()
const height = container.getComputedHeight()
```

# Expected Behavior
The computed `height` property should be 240(120 + 120)

# Actual Behavior
The computed `height` property is 400.

Seems like the container's `min-height` property is applied to each wrapped row.

# Notes
I tried both with and without web defaults.
This also happens on wrapped columns.

Please let me know if I'm missing something. Thanks!

goolAdapter:
try to fix #1300

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -11,7 +11,10 @@
       "request": "launch",
       "program": "${workspaceFolder}/tests/build/yogatests",
       "cwd": "${workspaceFolder}",
-      "preLaunchTask": "Build Unit Tests"
+      "preLaunchTask": "Build Unit Tests",
+      "args": [
+        "--gtest_filter=YogaTest.*" // Replace * with the name of your test case
+      ]
     },
     {
       "name": "Debug C++ Unit tests (vsdbg)",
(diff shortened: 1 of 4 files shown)


.vscode/launch.json
```diff
@@ -11,15 +11,21 @@
       "request": "launch",
       "program": "${workspaceFolder}/tests/build/yogatests",
       "cwd": "${workspaceFolder}",
-      "preLaunchTask": "Build Unit Tests"
+      "preLaunchTask": "Build Unit Tests",
+      "args": [
+        "--gtest_filter=YogaTest.*" // Replace * with the name of your test case
```
NickGerleman, line 16: I don't think all the tests start with this

goolAdapter:
I've seen someone else come up with a better proposal. I withdraw my request. Thank you

PR #1493, opened 2023-12-04
* fix Container min-height produces incorrect size when flex-wrap and…

cattenblue:
* fix Container min-height produces incorrect size when flex-wrap and Height is Auto


NickGerleman:
Could you help by providing some more detail on the root cause, and why this change fixes it?

NickGerleman:
Ah, that makes sense. If we have multiple lines, a line is allowed to be less than the min height of the parent container.

Looks like the algorithm is specified by https://www.w3.org/TR/css-flexbox-1/#algo-cross-line

> If the flex container is [single-line](https://www.w3.org/TR/css-flexbox-1/#single-line-flex-container), then clamp the line’s cross-size to be within the container’s computed min and max [cross sizes](https://www.w3.org/TR/css-flexbox-1/#cross-size).

Let me read how we normally handle this for multi-line. Need to see if we can get rid of clamping entirely, or if we do it somewhere else again later.

cattenblue:
Thank you for your response, and I look forward to your revisions.

NickGerleman:
> Thank you for your response, and I look forward to your revisions.

https://github.com/facebook/yoga/commit/464d1668baeb2e448c12f3cc95990924972c2532 should be merged with a fix for this (but that change caused another slight issue that https://github.com/facebook/yoga/pull/1524 should address).

PR #1498, opened 2023-12-07
Create node version of gentest script

joevilches:
Summary:
The only instance of ruby in this repository is `gentest.rb` used to generate test cases from html fixtures. This is quite annoying as ruby is not the most popular compared to something like Node and it does not integrate into the rest of our stack. I changed this to use Node.js instead. Instead of `watir` we now use `selenium-webdriver`. `watir` is backed by Selenium so I do not expect anything to change.

Next commits will add command line options, clean up gentest.rb and its references, and change the README

Differential Revision: D51874433

### DIFF
--- a/gentest/gentest-driver.ts
+++ b/gentest/gentest-driver.ts
@@ -1,0 +1,89 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow strict-local
+ * @format
+ */
+
+import * as fs from 'node:fs/promises';
+import {format} from 'node:util';
+import * as path from 'path';
+import * as process from 'node:process';
+import {Builder, logging} from 'selenium-webdriver';
+import {Options} from 'selenium-webdriver/chrome.js';
+
+await (async () => {
+  const options = new Options();
+  options.addArguments(
+    '--force-device-scale-factor=1',
+    '--window-position=0,0',
+    '--hide-scrollbars',
+  );
+  options.setLoggingPrefs({
+    browser: 'ALL',
+    performance: 'ALL',
+  });
+  const driver = await new Builder()
+    .forBrowser('chrome')
+    .setChromeOptions(options)
+    .build();
+
+  process.chdir(path.dirname(process.argv[1]));
+
+  const fixtures = await fs.readdir('./fixtures');
+  for (const fileName of fixtures) {
+    const fixture = await fs.readFile('fixtures/' + fileName, 'utf8');
+    const fileNameNoExtension = path.parse(fileName).name;
+    console.log('Generate', fileNameNoExtension);
+
+    // TODO: replace this with something more robust than just blindly replacing
+    // start/end in the entire fixture
+    const ltrFixture = fixture
+      .replaceAll('start', 'left')
+      .replaceAll('end', 'right')
+      .replaceAll('flex-left', 'flex-start')
+      .replaceAll('flex-right', 'flex-end');
+
+    const rtlFixture = fixture
+      .replaceAll('start', 'right')
+      .replaceAll('end', 'left')
+      .replaceAll('flex-right', 'flex-start')
+      .replaceAll('flex-left', 'flex-end');
+
+    const template = await fs.readFile('test-template.html', 'utf8');
+    const f = await fs.open('test.html', 'w');
+    await f.write(
+      format(template, fileNameNoExtension, ltrFixture, rtlFixture, fixture),
+    );
+    await f.close();
+
+    await driver.get('file://' + process.cwd() + '/test.html');
+    const logs = await driver.manage().logs().get(logging.Type.BROWSER);
+
+    await fs.writeFile(
+      `../tests/generated/${fileNameNoExtension}.cpp`,
+      eval(logs[0].message.replace(/^[^"]*/, '')),
+    );
+
+    await fs.writeFile(
+      `../java/tests/com/facebook/yoga/${fileNameNoExtension}.java`,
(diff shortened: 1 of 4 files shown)


nicoburns:
If you use playwright-core instead of selenium-webdriver then it will manage browser versions for you, and would also remove the need for users to install chromedriver.

gentest/gentest-driver.ts
```diff
@@ -0,0 +1,89 @@
...
+  });
+  const driver = await new Builder()
+    .forBrowser('chrome')
+    .setChromeOptions(options)
+    .build();
+
+  process.chdir(path.dirname(process.argv[1]));
+
+  const fixtures = await fs.readdir('./fixtures');
+  for (const fileName of fixtures) {
```
nicoburns, line 37: It would be nice if the behaviour when the generation script (script run within the browser tab) throws an error could be improved. Specifically, it would be nice if no tests were updated in that case (and ideally that the error, along with the file/fixture that caused it, was printed to the console).

gentest/gentest-driver.ts
```diff
@@ -0,0 +1,89 @@
...
+    );
+    await f.close();
+
+    await driver.get('file://' + process.cwd() + '/test.html');
+    const logs = await driver.manage().logs().get(logging.Type.BROWSER);
+
+    await fs.writeFile(
+      `../tests/generated/${fileNameNoExtension}.cpp`,
+      eval(logs[0].message.replace(/^[^"]*/, '')),
+    );
```
nicoburns, line 69: Looks like this doesn't delete old tests where the test fixture has been deleted or renamed? I think that was being handled in the old script by deleting the entire directory prior to writing out the new tests. IMO the best order would be:

- Run all fixtures in the browser (keeping results in memory)
- If all run without error:
    - Delete existing test directories
    - Write out all tests in parallel

gentest/gentest-driver.ts
```diff
@@ -0,0 +1,89 @@
+/**
```
nicoburns, line 1: This script needs instructions on how to run it (especially given that it's a TS file that won't run with vanilla node):

- Minimum node version
- That you need to run NPM install
- That you need chromedriver (if you do - I already had it installed so can't be sure)
- The run command (`node --loader babel-register-esm gentest-driver.ts` ?). Perhaps this could be included in the package.json as a script. And then the instructions could just be `npm run gentest` or similar.

gentest/gentest-driver.ts
```diff
@@ -0,0 +1,89 @@
...
+      format(template, fileNameNoExtension, ltrFixture, rtlFixture, fixture),
+    );
+    await f.close();
+
+    await driver.get('file://' + process.cwd() + '/test.html');
+    const logs = await driver.manage().logs().get(logging.Type.BROWSER);
+
+    await fs.writeFile(
+      `../tests/generated/${fileNameNoExtension}.cpp`,
+      eval(logs[0].message.replace(/^[^"]*/, '')),
```
nicoburns, line 68: I believe this could use `JSON.parse` instead of `eval` (which would probably be preferable both from a security and code readability perspective?)

gentest/gentest-driver.ts
```diff
@@ -0,0 +1,89 @@
...
+  process.chdir(path.dirname(process.argv[1]));
+
+  const fixtures = await fs.readdir('./fixtures');
+  for (const fileName of fixtures) {
+    const fixture = await fs.readFile('fixtures/' + fileName, 'utf8');
+    const fileNameNoExtension = path.parse(fileName).name;
+    console.log('Generate', fileNameNoExtension);
+
+    // TODO: replace this with something more robust than just blindly replacing
+    // start/end in the entire fixture
```
nicoburns, line 43: Probably the fixtures should be changed to use `inset-inline-start` and `inset-inline-end` instead of left/right? I don't think those existed when these tests were created.

nicoburns (commented):
Excited to see this new runner. I am very much considering adopting it for Taffy. Partly because it's a lot faster than our runner (which has to compile a Rust script before it runs, and runs each fixture in a fresh browser tab rather than batching them). And partly because there are a bunch of other derivatives of Yoga/Taffy (mostly ports/bindings to other languages) that could potentially use this as well, and the main benefit of our runner (that it is written in Rust, the same as our library) doesn't seem as beneficial in that context.

I've left you a few review comments :)

(CC: @alice-i-cecile)

gentest/gentest-driver.ts
joevilches: A followup me an @NickGerleman were thinking about was to make it so that there is minimum logic being run on the browser. Right now that is where all the complicated code generation comes from because that is where we inspect the browsers layout. Ideally we port that over to being driven off of the browser if our driver library has such capabilities (not very familiar with all that you can do). Right now, because all of that logic is on the browser and we just read the console.logs, it is hard to do anything elegant around error handling and debugging.

gentest/gentest-driver.ts
joevilches: ooo good callout. I can take a look at that

gentest/gentest-driver.ts
joevilches: Yeah in some of the later commits in this stack I update the readme for instructions on how to get this to work wrt to dependencies. We also have scripts in place: `yarn gentest` and `yarn gentest-validate` which abstracts that out that loader.

As for chromedriver, according to https://www.selenium.dev/documentation/webdriver/troubleshooting/errors/driver_location/#use-the-latest-version-of-selenium, it appears that selenium now handles those deps for you like playwright. I have not tested this yet though, I'll give it a try today

gentest/gentest-driver.ts
joevilches: There are the other things like `margin-start` etc. I think we need a pretty wholistic system to replace things like that

PR #1503, opened 2023-12-09
Sign generated tests

joevilches:
Summary: This diff makes it so that our driver will sign all of the generated files to help ensure that they are not edited by hand. Next I will add CI to actually verify the signature

Differential Revision: D51966201


nicoburns:
In Taffy we run the test generation in CI and then use git to check for changes files and fail the check if there are any. IMO this is a better approach than signing the tests, as it also catches the case where someone has edited a fixture and has forgotten to run the test generator at all (in which case the test signature would be valid but would no longer match the fixture)

joevilches:
I am also implementing a check for that, just not published yet. The signing has some extra internal benefits including changing the cursor in VS code so it is obvious the code should not be edited by hand, and internal CI checks from for free if you have a signature

PR #1556, opened 2024-01-16
Remove static-behaves-like-relative errata

joevilches:
Summary:
X-link: https://github.com/facebook/react-native/pull/42315

Since we aim to ship static to all users of yoga (not just XPR), we need to remove the errata that is gating most of the features. This should be a non breaking change. To ensure that, I added a new errata which, if on, will use the inner size of the containing node as the containing block. This is how it has been for a while and resolving this is risky and time consuming so for the time being we will stick with that.

Differential Revision: D52706161


nicoburns:
It's annoying that `static` was added to the API without actually being implemented, but I wonder if an errata is actually needed here. If a consumer of Yoga is aware of the need to enable this errata then could they not just as easily modify their style translation code to map `static` to `relative` when setting the style on the Yoga node?

NickGerleman:
> It's annoying that `static` was added to the API without actually being implemented, but I wonder if an errata is actually needed here. If a consumer of Yoga is aware of the need to enable this errata then could they not just as easily modify their style translation code to map `static` to `relative` when setting the style on the Yoga node?

Yeah, we don't want it. It was added as part of staging, since RN Fabric would use `static` to mean some special logic wrt stacking contexts. And we wanted to be able to work on static, before changing RN default behavior around stacking contexts.

PR #1576, opened 2024-02-09
Support for (de)serializing measure funcs

joevilches:
Summary:
In addition to all the state that gets set on the node that is easy to serialize - like floats, enums, bools, etc - we also need to serialize measure functions. This is because these functions take a nontrivial amount of time up during layout and we should capture that. Also, they are important to the ability to truly replay layout as it was captured as the results of the measure functions determine many of the steps the layout algorithm takes.

Capturing this is rather tricky however, but I think I found a solution that is relatively simple and non-error prone. Essentially, since we are capturing the entire tree and virtually every input that goes into the flexbox algorithm, we *should* be able to replay layout exactly as it was captured. This means that the order in which measure functions are called *should* be the same. If this is the case, then all we need to do to capture the measure functions is store their input, output, and duration in a big array. During deserialization we just keep track of an index and use that to determine which measure function we should call. That is the premise behind what happens in this diff. In theory the algorithm could change and the capture would be wrong but it is easy enough to recapture again.

In order to capture you need to insert a method exposed by CaptureTree.h into the client measure func, which is kind of annoying but not that bad. In future diffs I will put a macro in place to make this even easier.

I also add our first capture! Which is of a large react native desktop app

Differential Revision: D53581121


nicoburns:
> This means that the order in which measure functions are called should be the same

If the idea here is to capture real-world ui trees for performance benchmarking then I don't think you ought to be assuming that. Some of the most performance-impacting changes will be so precisely because they are impacting the number of measure function measurements. I guess if the idea is just to capture a trace for analysis then it makes sense, but not if you want to be able to rerun the capture against newer versions of Yoga.

Have you considered capturing the React element tree instead (or as well)? I feel like it would be pretty easy to create a Yoga tree from a capture of that.

joevilches:
@nicoburns 

Yes this is for performance benchmarking. That is a valid point. Any way of serializing measure funcs runs into this issue I believe and the idea would just be to recapture the UI, which admittedly would get annoying.

Coincidentally, the captures I have were actually taken on a version of Yoga that was around a year old, but replayed on current Yoga and the measure func calls match up exactly (from what I checked). So maybe it is not too common of a problem. 

Can you expand on what you mean by capturing React element tree? As in a big json of the components + props that matter wrt Yoga? That would still not support the measure functions unless we replayed on react native which I guess is a possibility.

NickGerleman:
> Have you considered capturing the React element tree instead (or as well)? I feel like it would be pretty easy to create a Yoga tree from a capture of that.

I think the closest analog to this would be serializing either the React Component Tree, or the React Native ShadowTree.

It is common practice today to run product code in a Node JS environment, running React through a synthetic renderer (usually, for snapshot testing, or basic component interaction). This is isolated from the whole native stack though. When a real RN renderer is used, React tells React Native how to construct it's ShadowTree from its fibers. Then the ShadowTree itself has:

1. Native component props, derived from React props
2. Native component state, updatable from multiple threads, e.g. in response to native events on UI thread
3. Output layout metrics, that must be computed and set during the commit
4. Underlying immutable state on the particular (user-definable) ShadowNode, created during construction or commit. E.g. the base ShadowNode providing Yoga layout keeps a Yoga node here.

When we look at what is takes to run this in a replicable environment, it starts to look a lot like running a fuller RN instance 🙂, and perf-testing that. Admittedly, there are forms of RN more embeddable than running a full mobile app, and we could be taking more advantage of them. Definitely a lot more involved than the Yoga tree though.

NickGerleman:
> Coincidentally, the captures I have were actually taken on a version of Yoga that was around a year old, but replayed on current Yoga and the measure func calls match up exactly (from what I checked). So maybe it is not too common of a problem.

I think we want to effectively key these based on constraints, instead of order. I.e. a measure function can replay a previous constraint it was asked for.

That doesn't get everything (esp thinking about a close future with min-content causing us to need to retake these), but I think it should be rare enough that the actual constraint set changes.

joevilches:
@NickGerleman by constraints do you just mean the input? So available width/height and width/height measure mode? If so, there seem to be several instances where same input maps to different outputs. For example: 
Screenshot 2024-02-12 at 10 21 32

The output-width here differs despite same input

NickGerleman:
That seems strange to me. Unless underlying content changes, the same text node under same constraints should lead to same measure. Are these all from the same commit and node?

joevilches:
They appear to be from different nodes (at least the memory address of the yoga node at the time of the capture is different). They are for sure on different commits, I clear the vec containing the inputs during each call to calculate layout.

To test this I just iterated over the list on each capture and compared inputs (available height/width and width/height mode).

nicoburns:
I agree that measurements ought to be keyed by input constraints. And would add that timings (how long the measure function takes to run) ought to be captured in addition to outputs.

I would also point you to:

- Taffy's [test measure function](https://github.com/DioxusLabs/taffy/blob/main/tests/fixtures.rs) which has a small measure function (~30 LoC) that can produce reasonable height/width outputs for a given text input by assuming that each character has a fixed width. We use this in our generated tests along with the [Ahem font](https://github.com/Kozea/Ahem) which actually has fixed-sized characters in order to be able to do generated tests that include text (which is by far the easiest way to test min-content vs. max-content constraints).
- Taffy's ["cosmic_text" example](https://github.com/DioxusLabs/taffy/blob/main/examples/cosmic_text.rs) which uses the [cosmic-text](https://github.com/pop-os/cosmic-text) library to actually shape and layout arbitrary text with real fonts.

I feel like a mock measure function based on raw text captured from an app along with either:
- something like cosmic text that can properly shape it and lay it out (with realistic performance because it's actually doing the work), but is standalone and doesn't rely on the underlying OS.
- something like the simpler text measure function where the timings are faked based on an analysis of captured timings

might be a good approach for benchmarks.

PR #1594, opened 2024-03-11
Fix TypeScript link

jamonholmgren:
Link was wrong, fixed it.

* [Old link](https://github.com/facebook/yoga/blob/main/javascript/src/wrapAssembly.d.ts)
* [New link](https://github.com/facebook/yoga/blob/main/javascript/src/wrapAssembly.ts)


NickGerleman:
FYI that we’re planning to switch this over to “website-next” something hopefully later this week.

The old one will still build and get auto published, but the whole website might soon get replaced (with some new documentation also explaining how to actually use Yoga 😅).

PR #1631, opened 2024-03-21
Add JavaScript bindings for YGHasNewLayout

### LINKED ISSUE #681: [JS] JavaScript Bindings missing for YGNodeHasNewLayout
I'm wondering what others think about adding in some flags to mark which nodes were changed during the last calculateLayout. I believe a didChange and didTreeChange would be useful for walking the tree after the layout has been calculated and applying layout values to other systems.

### didChange
Determines if any of the values in a node's calculated layout are different from before
### didTreeChange
Determines if any of the node's descendants have changed during the last layout

I see these as the opposite of isDirty and isTreeDirty (which is basically what's requested here https://github.com/facebook/yoga/issues/425), so I suspect they could all be merged into bitflags to keep the memory footprint down. And I believe the did*Change flags should be easy to compute while calculating the layout since you'll have access to previous and new values at the same time to know if didChange is true, in which case you can set the parent's didTreeChange flag at the same time, so it's all O(1). 

### Use case
When I mark a node dirty and eventually do a calculateLayout, I then need to do a reflow event where I copy the layout from the nodes to the underlying system, but often most nodes haven't changed. I'd like to be able to walk down only the paths in the tree that have values that have changed and need to be transferred over. I think this could provide a huge performance improvement for any similar systems. Right now I have to keep an old version of all layouts around to compare with, and I have to walk over ALL nodes and compare them first to see if I need to copy values over.

Opinions?

WillsonHaw:
Summary: Adds JavaScript bindings for YGHasNewLayout which introduces
two new node methods: `hasNewLayout()` and `markLayoutSeen()`.

Closes #681

### DIFF
--- a/javascript/src/Node.cpp
+++ b/javascript/src/Node.cpp
@@ -421,6 +421,14 @@ bool Node::isDirty(void) const {
   return YGNodeIsDirty(m_node);
 }
 
+void Node::markLayoutSeen() {
+  YGNodeSetHasNewLayout(m_node, false);
+}
+
+bool Node::hasNewLayout(void) const {
+  return YGNodeGetHasNewLayout(m_node);
+}
+
 void Node::calculateLayout(double width, double height, int direction) {
   YGNodeCalculateLayout(
       m_node, width, height, static_cast<YGDirection>(direction));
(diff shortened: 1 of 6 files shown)


NickGerleman:
Looks great!

Issue #1637, opened 2024-03-29
Top Level Await


bbohlender:
Hey, congrats on launching yoga 3.0 🥳 

I am building [@react-three/uikit](https://github.com/pmndrs/uikit) and I've just upgraded it to yoga 3.0.
Unfortunately, for many bundlers/frameworks top-level-await requires additional configuration (vite requires setting target to esnext in the vite config) or simply doesn't work (nextjs 14 with app router).

I like the idea of having top-level-await for this library but I think the tooling is not there yet. I think an additional entry point without top-level-await would be appropriate while the tooling catches up (I can make a PR for that if reasonable).

NickGerleman:
> I like the idea of having top-level-await for this library but I think the tooling is not there yet. I think an additional entry point without top-level-await would be appropriate while the tooling catches up (I can make a PR for that if reasonable).

I think having a secondary entrypoint without TLA might make sense, as a fallback to those who cannot use it. I tried to consolidate the flavors compared to previous version (where it got a bit out of hand), but TLA is probably going to be the biggest thing preventing 3.0 from working automagically in a lot of setups.

It does Seem like TLA is getting there. 95% browser support, all stable versions of Node, standardized in ES2022 (but, Vite defaults to `es2020`). But as you noted, common toolchains out there still aren't set up to support it by default.

Later down the line is ESModule WASM integration, which has historically explicitly depended on TLA, but that [might have changed](https://github.com/WebAssembly/esm-integration/commit/9f6136d664818e9a6e92caf0d2e2515cc8ae7bcd).

bbohlender:
Fixed with #1638 
@NickGerleman, thanks for merging 🙏 
Is there an expected release timeline for, e.g., 3.0.3, so I can move a few users of uikit to the latest version that uses yoga 3.0 ? :)

NickGerleman:
@bbohlender I will push out a new release in the next couple days

NickGerleman:
Yoga 3.0.3 should be published now

bbohlender:
@NickGerleman thank you very much for pushing this out so fast.
I've just finished moving uikit to `yoga-layout/load`.

Unfortunately, the use of `exports` resulted in discovering what is, in my opinion, a typescript bug. Essentially typescript prefers to resolve `xy.z.js` to `xyz.ts` even if `xyz.js` exists only if using the `exports` field. The following screenshots show that even version `3.0.2` throws an error if I add
```
"exports": {
  ".": "./src/index.js"
}
[...]

PR #1638, opened 2024-03-30
Entry point without TLA for js package

bbohlender:
Follow up on #1637 

TLDR: tooling for TLA is not there yet; An additional entry point without top-level-await is appropriate

- adds ./load entry to js package
- uses .js file extensions to prevent requiring [allowImportingTsExtensions](https://www.typescriptlang.org/tsconfig#allowImportingTsExtensions)


NickGerleman:
> uses .js file extensions to prevent requiring [allowImportingTsExtensions](https://www.typescriptlang.org/tsconfig#allowImportingTsExtensions)

We are already doing this in the babel transform during packing. 

IMG_0512

NickGerleman:
It looks like the Jest environment didn’t like the extension change.

The model in the repo right now is that when Yoga is published, its package entry point is changed to be written to babel generated JS, along with rewriting imports. But still include original sources as well.

NickGerleman (changes requested):
Current change seems to be breaking the dev-time builds and scripts which transform the typescript on the fly.

bbohlender:
Sorry for my confusion

NickGerleman (approved):

NickGerleman:
@bbohlender would you be able to give this a quick mention in the package readme?

bbohlender:
@NickGerleman, for sure! Is that, what you had in mind?

NickGerleman:
That looks great. Thanks for this PR!

PR #1639, opened 2024-04-02
Update align-content handling of overflow

NickGerleman:
Summary:
Gentest tests started failing because Chrome changed behavior of overflowed align-content container. Spec says should fallback to "safe center", which is really just "start", instead of previous "center" behavior. This changes behavior accordingly.

There is one bit where I think we are doing the wrong thing wrt alignment of flex start vs start (which we don't support yet), but couldn't repro a failing chrome test.

Differential Revision: D55617689


nicoburns:
> There is one bit where I think we are doing the wrong thing wrt alignment of flex start vs start (which we don't support yet), but couldn't repro a failing chrome test.

I believe Chrome is also falling back to flex-start currently (see: https://github.com/servo/servo/pull/31724). I agree with you that this seems to violate [the spec](https://drafts.csswg.org/css-align/#distribution-values), but the spec is currently somewhat inconsistent as it does specify `flex-start` as the fallback for some alignment modes.

nicoburns:
CSSWG issue https://github.com/w3c/csswg-drafts/issues/10154

PR #1643, opened 2024-04-12
Implement Percentage support for gap styles

realsoelynn:
Summary:
Changelog [Internal]:
- Added percentage value for flex layout gap
- Wired up to pass proper available width and height to implement this feature

Differential Revision: D56002340


nicoburns:
We have some tests for this in Taffy's test suite (search for `gap_percentage` here: https://github.com/DioxusLabs/taffy/tree/main/test_fixtures/flex). In particular you may want to add tests for the case where the container's size isn't a fixed size, which can require re-resolving the gap size once the container size is computed.

realsoelynn:
> We have some tests for this in Taffy's test suite (search for `gap_percentage` here: https://github.com/DioxusLabs/taffy/tree/main/test_fixtures/flex). In particular you may want to add tests for the case where the container's size isn't a fixed size, which can require re-resolving the gap size once the container size is computed.

Thank you so much for the feedback on this. I am looking at how to improve our test cases as well as visualization aspect of it.

gentest/fixtures/YGGapTest.html
```diff
@@ -171,3 +171,72 @@
...
+  <div style="flex: 1;"></div>
+  <div style="flex: 1;"></div>
+  <div style="flex: 1;"></div>
+</div>
+
+<div id="row_gap_percent_wrapping_with_mixed_flexible_content" style="flex-direction: row; width: 300px; height: 300px; gap: 10%;">
+  <div style="width: 10px;"></div>
+  <div style="flex: 1;"></div>
+  <div style="width: 10%;"></div>
+</div>
```
realsoelynn, line 233: @nicoburns Thanks for the test case suggestion.

PR #1647, opened 2024-04-17
Expose Gap Percentage to ReactNative

realsoelynn:
Summary:
Expose the Gap Percent from Yoga to RN Layer

Changelog: [Internal]
- Enable flex gap percentage value for RN.

Differential Revision: D56160597


realsoelynn:
Not required any more

PR #1674, opened 2024-07-01
Add config version, and invalidate layout on config change

acoates-ms:
This is a continuation of the previous PR: https://github.com/facebook/react-native/pull/45047

I made the change more generic for allowing any kind of config change to invalidate layout.

### DIFF
--- a/tests/YGScaleChangeTest.cpp
+++ b/tests/YGScaleChangeTest.cpp
@@ -112,3 +112,66 @@ TEST(YogaTest, errata_config_change_relayout) {
 
   YGConfigFree(config);
 }
+
+static uint32_t _measureCallCount = 0;
+static YGSize _measureCustom(
+    YGNodeConstRef /*node*/,
+    float width,
+    YGMeasureMode /*widthMode*/,
+    float height,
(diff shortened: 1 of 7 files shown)


NickGerleman (changes requested):
This looks good! But I think it introduces subtle bug that breaks layout caching in Fabric.

In current Fabric impl, we lazily clone any committed ShadowNodes when we know we may need to write new layout information to them. This currently means new inline Yoga node and Yoga config in the ShadowNode. The new config is associated with cloned node via `YGNodeSetConfig`, and right now this only dirties the node if meaningful config values have changed.

After this change, if we have new config, derived from previous one, they may be structurally equal, but have a different revision count. So the process of building new config, derived from previous, may still lead to invalidation. Or, in the reverse, a config revision ID is no longer valid if a new config is assigned, and we should clear it (though we do dirty anyways).

I think, we can address this, by adding code around the existing equality check (IIRC in `Node::setConfig()`), to update kept revision ID to match new config revision ID if they are equal, or clear if not.

NickGerleman (approved):
LGTM

PR #1681, opened 2024-07-11
update:yoga-caching-algorithm

admirsaheta:
## Summary:

The pull request optimizes the `canUseCachedMeasurement` function in the Yoga library by reducing redundant computations, combining conditions where possible, and enhancing readability and performance. These changes aim to improve the efficiency of the function without altering its overall implementation or behavior.

## Changelog:

[GENERAL][CHANGED] - Optimized the `canUseCachedMeasurement` function to reduce redundant computations and improve performance.

## Test Plan:

1. **Unit Tests**: Ran existing unit tests to ensure the optimized function produces the same results as before.
2. **Performance Tests**: Conducted performance tests to measure the execution time of the `canUseCachedMeasurement` function before and after the optimization.
3. **Code Review**: Reviewed the changes to ensure they maintain the original logic and correctness of the function.

The tests confirmed that the optimized function is more efficient while producing the same correct outputs.

### DIFF
--- a/yoga/algorithm/Cache.cpp
+++ b/yoga/algorithm/Cache.cpp
@@ -57,63 +57,49 @@ bool canUseCachedMeasurement(
     const float marginColumn,
     const yoga::Config* const config) {
   if ((yoga::isDefined(lastComputedHeight) && lastComputedHeight < 0) ||
-      ((yoga::isDefined(lastComputedWidth)) && lastComputedWidth < 0)) {
+      (yoga::isDefined(lastComputedWidth) && lastComputedWidth < 0)) {
     return false;
   }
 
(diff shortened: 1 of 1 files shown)


efstathiosntonas:
I’ve patched rn 0.74.3 with this change and I’ve encountered weird layout from time time on both iOS and Android.

The layout was tending to move upwards on the screen, eg a flex:1 view could move all the way up and only the bottom part was visible. 

In another case  on Android only I had 3 elements in a flex direction row alignitems center and the 2 right last elements were moved upwards again whille the left one was positioned fine. 

Don’t know if something is missing on my end (maybe there are other yoga commits on 0.75 branch that accompany this one).  I thought it worths mentioning it. 

cc: @NickGerleman

PR #1690, opened 2024-08-24
Add support for using Yoga from swift on non Darwin platforms. 

iainsmith:
Specify that enums are closed for correct swift export on non darwin platforms.

---

Swift exports plain c enums as constants. On Darwin the yoga macros hit the NS_ENUM case which generates a swift enum, however on linux NS_ENUM is not defined.

See [This post] for a some more background. https://belkadan.com/blog/2021/10/Swift-Regret-Unannotated-C-Enums/

Tested on a couple of CI builds:
- [Failing build](https://github.com/iainsmith/YogaSwift/actions/runs/10478362350/job/29021559312)
- [Passing build](https://github.com/iainsmith/YogaSwift/actions/runs/10493885336/job/29068883339) swift build passes on this, but swift test is failing still for unrelated reasons.

Which results in a compiler error for the missing switch cases.
```swift
[303/309] Compiling YogaSwift YGNodeView.swift
/home/runner/work/YogaSwift/YogaSwift/Sources/YogaSwift/YGNode.swift:50:56: error: type 'YGDirection' has no member 'LTR'
48 |     width: Float = YGValueUndefined.value, height: Float = YGValueUndefined.value
49 |   ) {
50 |     YGNodeCalculateLayout(self.__node, width, height, .LTR)
   |                                                        `- error: type 'YGDirection' has no member 'LTR'
51 |   }
52 | }
```

### DIFF
--- a/yoga/YGMacros.h
+++ b/yoga/YGMacros.h
@@ -49,7 +49,7 @@
 #define YG_ENUM_END(name)
 #else
 #define YG_ENUM_BEGIN(name) enum name
-#define YG_ENUM_END(name) name
+#define YG_ENUM_END(name) __attribute__((enum_extensibility(closed))) name
 #endif
 
 #ifdef __cplusplus


NickGerleman (changes requested):
This breaks the build on non-Clang platforms.

yoga/YGMacros.h
```diff
@@ -49,7 +49,7 @@
 #define YG_ENUM_END(name)
 #else
 #define YG_ENUM_BEGIN(name) enum name
-#define YG_ENUM_END(name) name
+#define YG_ENUM_END(name) __attribute__((enum_extensibility(closed))) name
```
NickGerleman, line 52: A couple of these are flag enums, which would also need to be declared I think

iainsmith:
> This breaks the build on non-Clang platforms.

Makes sense. I'll push a change to wrap this in a __clang__ check.

yoga/YGMacros.h
iainsmith: @NickGerleman do we want this for `YGErrata & YGExperimentalFeature` or are there any others?

iainsmith: Just realised that enums.py specifies that only `Errata` is a bitset enum.

NickGerleman: That sounds good to me!

iainsmith: @NickGerleman , I took a quick stab at this on [this branch](https://github.com/facebook/yoga/compare/main...iainsmith:yoga:flag_enums?expand=1), but I've run into a compiler error I'm not sure how to approach.

```
~/Developer/personal/yoga/yoga/../yoga/YGEnums.h:61:5: 
error: enumeration value 'YGErrataClassic' is out of range of flags in enumeration type 'YGErrata' [-Werror,-Wflag-enum]
   61 |     YGErrataClassic = 2147483646,
      |     ^
~/Developer/personal/yoga/yoga/../yoga/YGEnums.h:62:5: 
error: enumeration value 'YGErrataAll' is out of range of flags in enumeration type 'YGErrata' [-Werror,-Wflag-enum]
   62 |     YGErrataAll = 2147483647)
```

Do you have thoughts on the best way to handle that?

Issue #1678, opened 2024-07-04
0 layoutWidth and 0 layoutHeight after repeated layout calculation


michaeltroger:
# Report

- [x] I have searched [existing issues](https://github.com/facebook/yoga/issues) and this is not a duplicate

# Issues and Steps to Reproduce
General setup:
- Any Android app
- Yoga library 3.1.0 from Maven (issue happens since 2.0.0), debug version
- using Yoga's Java API

Actual Yoga API usage making it reproducible:
1. Injecting a YogaConfig (usually an optional argument) when creating the YogaNode
2. Caching the created YogaNode instances
3. Resetting the existing instances before re-applying the properties like width/height
4. Calculating the layout
5. Adding some artificial delay before repeating from 3

The issue happens already with a single YogaNode and setting only width/height (see also sample app), but the issue is not limited to this scenario. I have reasons to believe that it happens more often the bigger the tree is, but I thought a single YogaNode might make debugging the issue easier. If the tree is bigger then individual Nodes are affected but not necessarily all Nodes of the tree.

# Expected Behavior
I would expect that consecutive calls with the same input always result in the same output.

# Actual Behavior
After some iterations (repeating the resetting of the YogaNode, the applying of properties and the calculation of the layout), `layoutWidth` and `layoutHeight` deliver 0 values.
It doesn't happen after a fixed amount of repetitions. It's mostly influenced through the delay between the calls and also the context in which the code is executed matters.
E.g. I had the issue in the provided sample app after:
- 300 iterations with 800 milliseconds delay in between
- 110 iterations with 600 milliseconds delay in between
- 150 iterations with 400 milliseconds delay in between
- 200 iterations with 200 milliseconds delay in between
- 350 iterations with 100 milliseconds delay in between
- 6000 iterations with 5 milliseconds delay in between

One needs to play around with these delays though. In a company internal Demo app it needs **only 5 iterations with a delay of 5 seconds** in between, while I see no issue with such delay in the provided sample app at all.

Coming to the actual production impact: I originally found the issue when rendering a layout with a root and 2 children boxes, each of them having another child box. **After re-calculating the layout ~5 times, some boxes started to disappear and the debugging led me to the 0 values.**

We were using Yoga 1.19.0 until now and the same code in Yoga 1 returns approximately after the same iterations, floating point numbers that are not entirely correct, but if rounded you wouldn't see any issue. 
For a size of 200x200, it returns 199.99997x199.99997 after 350 iterations in the sample app.

[...]

PR #1703, opened 2024-09-25
Fix YogaConfig getting garbage collected #1678

michaeltroger:
Fixes [#1678](https://github.com/facebook/yoga/issues/1678)

As described in the linked Issue, the problem is that the `YogaConfig` can get garbage collected by the JVM, while a `YogaNode` is still referring to it. This at some point leads to unexpected behaviour (0 values for `layoutWidth`/`layoutHeight`). The change coming with this PR makes sure the `YogaConfig` can not get garbage collected while it's used by a `YogaNode`.

Demo project to confirm the fix https://github.com/michaeltroger/yogabug

Kudos to @rtPag, who helped identifying the issue.


NickGerleman (approved):
LGTM!

Issue #1707, opened 2024-09-26
Podfile.lock changed everytime when installed on different machines


sheltonsuen:
# Report

- [x] I have searched [existing issues](https://github.com/facebook/yoga/issues) and this is not a duplicate

# Issues and Steps to Reproduce
***Replaces this with steps to repro your issue.***
## Env
* Mac version: 14.7 (23H124)
* Xcode version: 15.4 (15F31d)
* ruby: 3.2.3
* cocoapods: 1.15.2
* react native: 0.73.6
* node: 20.17.0

## Steps to reproduce
* `yarn install` then `pod install`

# Expected Behavior
***Describe what you expected would happen.***
* the `Podfile.lock` should keep consistent on different machines

# Actual Behavior
***Describe what actually happened.***
* the `Podfile.lock` changed on different machines

# Link to Code
***If you have some code that maintainers can clone/test for themselves, bugs can be resolved much faster. Please paste a link here.***

***When applicable, use this [fiddle](https://jsfiddle.net/emilsjolander/jckmwztt/) to post a web repro.***

sheltonsuen:
<img width="1395" alt="image" src="https://github.com/user-attachments/assets/37022875-2782-4582-a875-c2d72689b3d6">
The diff of Podfile.loc

sheltonsuen:
close as wrong repo to refer

PR #1709, opened 2024-09-26
fix: Podfile.lock changed everytime when installed on different machines

sheltonsuen:


sheltonsuen:
This PR is related to issue #1707, our team encountered this issue, after search around, we found this line of code can solve our problem, I'm not sure this is the root cause, hoping this can provide any insights on this issue

NickGerleman:
Are you hitting this when using React Native? Note that it has its own (very similar) podspec, different from the one in this repo: https://github.com/facebook/react-native/blob/main/packages/react-native/ReactCommon/yoga/Yoga.podspec

But the underlying cause here is that different Ruby versions sort things differently

sheltonsuen:
> Are you hitting this when using React Native? Note that it has its own (very similar) podspec, different from the one in this repo: https://github.com/facebook/react-native/blob/main/packages/react-native/ReactCommon/yoga/Yoga.podspec
> 
> But the underlying cause here is that different Ruby versions sort things differently, so they may create different results depending on machine if you aren't using consistent versions.

Aha, I thought this is the underlying yoga of RN, thanks for point me out 

By the way, the ruby version between out team member already keep consistent with each other (ruby 3.2.3), so I also wondering what's the root cause

PR #1726, opened 2024-10-14
Add support for `display: contents` style

j-piasecki:
## Summary

This PR adds support for `display: contents` style by effectively skipping nodes with `display: contents` set during layout.

This required changes in the logic related to children traversal - before this PR a node would be always laid out in the context of its direct parent. After this PR that assumption is no longer true - `display: contents` allows nodes to be skipped, i.e.:

```html
<div id="node1">
  <div id="node2" style="display: contents;">
    <div id="node3" />
  </div>
</div>
```

`node3` will be laid out as if it were a child of `node1`.

Because of this, iterating over direct children of a node is no longer correct to achieve the correct layout. This PR introduces `LayoutableChildren::Iterator` which can traverse the subtree of a given node in a way that nodes with `display: contents` are replaced with their concrete children.

A tree like this:
```mermaid
flowchart TD
    A((A))
    B((B))
    C((C))
    D((D))
    E((E))
    F((F))
    G((G))
    H((H))
    I((I))
    J((J))

    A --> B
    A --> C
    B --> D
    B --> E
    C --> F
    D --> G
    F --> H
    G --> I
    H --> J

    style B fill:#050
    style C fill:#050
    style D fill:#050
    style H fill:#050
    style I fill:#050
```

would be laid out as if the green nodes (ones with `display: contents`) did not exist. It also changes the logic where children were accessed by index to use the iterator instead as random access would be non-trivial to implement and it's not really necessary - the iteration was always sequential and indices were only used as boundaries.

There's one place where knowledge of layoutable children is required to calculate the gap. An optimization for this is for a node to keep a counter of how many `display: contents` nodes are its children. If there are none, a short path of just returning the size of the children vector can be taken, otherwise it needs to iterate over layoutable children and count them, since the structure may be complex.

One more major change this PR introduces is `cleanupContentsNodesRecursively`. Since nodes with `display: contents` would be entirely skipped during the layout pass, they would keep previous metrics, would be kept as dirty, and, in the case of nested `contents` nodes, would not be cloned, breaking `doesOwn` relation. All of this is handled in the new method which clones `contents` nodes recursively, sets empty layout, and marks them as clean and having a new layout so that it can be used on the React Native side.

Relies on https://github.com/facebook/yoga/pull/1725

## Test plan

Added tests for `display: contents` based on existing tests for `display: none` and ensured that all the tests were passing.


yoga/node/Node.h
```diff
@@ -31,6 +31,128 @@ namespace facebook::yoga {
...
+        // if no node without display: contents was found, try to backtrack
+        if (currentNode->style().display() == Display::Contents) {
+          next();
+        }
+      }
+
+      const Node* node_;
+      size_t childIndex_;
+      size_t currentNodeIndex_{0};
+      Backtrack backtrack_;
```
NickGerleman, line 131: Most of the time it will be an empty vector, so no heap allocation, but I wonder if this could refer to e.g. single parent iterator (and that iterator could point to parent), such that iteration only allocates memory on the stack.

yoga/node/Node.h
```diff
@@ -31,6 +31,128 @@ namespace facebook::yoga {
...
+      }
+    }
+
+    Iterator end() const {
+      return Iterator(nullptr, -1);
+    }
+
+   private:
+    const Node* node_;
+  };
```
NickGerleman, line 154: Does this all work out if we add following below? As I'm sure you ran into, missing or incorrect signatures can lead to cryptic errors within standard library, so it's helpful to have something like this which can show whether iterator is valid implementation.

`static_assert(std::input_iterator<LayoutableChildren::Iterator>);`

yoga/node/Node.h
```diff
@@ -298,6 +438,7 @@ class YG_EXPORT Node : public ::YGNode {
   Style style_;
   LayoutResults layout_;
   size_t lineIndex_ = 0;
+  size_t contentsChildren_ = 0;
```
NickGerleman, line 441: nit: this reads as a list of children
```suggestion
  size_t contentsChildrenCount_ = 0;
```

NickGerleman (commented):
This is really excellent! Thanks for all the work you've done here. I know this has gone through a lot of iteration to get to this point, and I think we've landed on a really solid way to implement this in Yoga + RN.

I can see it incorporates https://github.com/facebook/yoga/pull/1725 as part of it. I might want to split while importing, to land that part first, in case it leads to any unexpected breakage.

It also looks like this generates some MSVC /W4 specific warnings. We should fix those before merging. https://github.com/facebook/yoga/actions/runs/11328004846/job/31512544728?pr=1726

yoga/node/Node.h
j-piasecki: Done in https://github.com/facebook/yoga/pull/1726/commits/6b6acde221894d38d5956f46f461fb2633c31915

yoga/node/Node.h
j-piasecki: Done in https://github.com/facebook/yoga/pull/1726/commits/5028c3e1f7692ec14e866fbefced9a4042afdd6e

yoga/node/Node.h
```diff
@@ -31,6 +31,128 @@ namespace facebook::yoga {
 
 class YG_EXPORT Node : public ::YGNode {
  public:
+  class LayoutableChildren {
```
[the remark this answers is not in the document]
j-piasecki: Done in https://github.com/facebook/yoga/pull/1726/commits/73053201e6cb5dc54876a3b9474055be03375c2a

yoga/node/Node.h
j-piasecki: I don't think this is easily achievable. I see two main problems with it:
- incrementing the iterator would have to return a new instance with a pointer to the previous one when going deeper into the tree, we would need to prolong the lifetime of the previous iterator somehow, so that it's valid when actually backtracking.
- modifying the "parent" iterator would possibly break backtracking logic. To prevent that, each iterator would need to keep a deep copy of all ancestors.

NickGerleman (approved):
LGTM. I will break apart the absolute positioning change from this and land first just to be safe (will try to merge Wednesday after a lot of branches cut).

yoga/node/LayoutableChildren.h
```diff
@@ -0,0 +1,142 @@
...
+    const T* node_;
+    size_t childIndex_;
+    size_t currentNodeIndex_{0};
+    Backtrack backtrack_;
+
+    friend LayoutableChildren;
+  };
+
+  LayoutableChildren(const T* node) : node_(node) {
+    static_assert(std::input_iterator<LayoutableChildren<T>::Iterator>);
```
NickGerleman, line 118: Apparently there was a C++ 20 standard defect that led older stdlib to require these to be default constructible 😮. https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2325r3.html

Going to locally patch in a default ctor that just initializes to nullptr...

```
buck-out/v2/gen/fbsource/aaebd9ffd9ca7aaa/xplat/yoga/__yogaFbcode__/buck-private-headers/yoga/node/Node.h:150:12: note: in instantiation of member function 'facebook::yoga::LayoutableChildren<facebook::yoga::Node>::LayoutableChildren' requested here
    return LayoutableChildren(this);
           ^
buck-out/v2/gen/fbsource/aaebd9ffd9ca7aaa/xplat/yoga/__yogaFbcode__/buck-private-headers/yoga/node/LayoutableChildren.h:118:19: note: because 'LayoutableChildren<Node>::Iterator' does not satisfy 'input_iterator'
    static_assert(std::input_iterator<LayoutableChildren<T>::Iterator>);
                  ^
fbcode/third-party-buck/platform010/build/libgcc/include/c++/trunk/bits/iterator_concepts.h:634:30: note: because 'facebook::yoga::LayoutableChildren<facebook::yoga::Node>::Iterator' does not satisfy 'input_or_output_iterator'
    concept input_iterator = input_or_output_iterator<_Iter>
                             ^
fbcode/third-party-buck/platform010/build/libgcc/include/c++/trunk/bits/iterator_concepts.h:614:5: note: because 'facebook::yoga::LayoutableChildren<facebook::yoga::Node>::Iterator' does not satisfy 'weakly_incrementable'
        && weakly_incrementable<_Iter>;
           ^
fbcode/third-party-buck/platform010/build/libgcc/include/c++/trunk/bits/iterator_concepts.h:597:36: note: because 'facebook::yoga::LayoutableChildren<facebook::yoga::Node>::Iterator' does not satisfy 'default_initializable'
    concept weakly_incrementable = default_initializable<_Iter>
                                   ^
fbcode/third-party-buck/platform010/build/libgcc/include/c++/trunk/concepts:143:37: note: because 'facebook::yoga::LayoutableChildren<facebook::yoga::Node>::Iterator' does not satisfy 'constructible_from'
    concept default_initializable = constructible_from<_Tp>
                                    ^
fbcode/third-party-buck/platform010/build/libgcc/include/c++/trunk/concepts:139:30: note: because 'is_constructible_v<facebook::yoga::LayoutableChildren<facebook::yoga::Node>::Iterator>' evaluated to false
      = destructible<_Tp> && is_constructible_v<_Tp, _Args...>;
```

yoga/node/Node.h
```diff
@@ -144,6 +146,24 @@ class YG_EXPORT Node : public ::YGNode {
     return children_.size();
   }
 
+  const LayoutableChildren getLayoutChildren() const {
```
NickGerleman, line 149: The other children already do this, so makes sense to keep with the pattern (no change needed/requested), but it is bizarre that the methods which return mutable references to children are marked const.

NickGerleman:
Edge case I didn't think of (not blocker), what happens if the root node has `display: contents`?

PR #1728, opened 2024-10-16
Added document how to build Yoga for Java

darekbx:
Added documentation detailing how to build the Yoga library for Java/Android.

### DIFF
--- a/java/BUILD_JAVA.md
+++ b/java/BUILD_JAVA.md
@@ -1,0 +1,99 @@
+## Building Yoga native lib for Java
+Instruction how to build Yoga native lib on local MacOs/Linux machine.
+
+### Prerequisites
+Required tools:
+* ninja
+* cmake
+* g++
+
+### Steps
+
+1. Modify `cmake/project-defaults.cmake` file by commenting out `-Werror`:
+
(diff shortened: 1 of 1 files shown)


NickGerleman:
Writing instructions for how to patch the current build logic to build JNI part of library for macOS doesn’t seem very durable.

For Android we rely on Android Gradle Plugin’s NDK/CMake integration, and just distribute binaries.

If we wanted to support building Java locally more gracefully, we probably want a way to have Gradle without Android bits hook CMake, or potentially just go the route of Gradle C++ libraries more venally. https://docs.gradle.org/current/userguide/cpp_library_plugin.html

Separately, out of curiosity, what warnings as errors did you run into? We have a good number of different builds with high warning level in OSS CI, and internally at Meta, so I’m curious what gap we still have.

lunaleaps:
Can't speak on the actual build instructions, but wondering why the README wasn't used for these instructions.

darekbx:
@NickGerleman We got some errors and warnings while building Yoga for Java on MacOs with Apple CPU. We are creating custom implementation of Flexbox layout for Android Compose.

PR #1729, opened 2024-10-24
Fix for nodes with `display: contents` not being cleaned in some cases

j-piasecki:
## Summary

Fixes a case where a node with `display: contents` would not be cleaned up in some cases. This was caused by it being called after some early returns handling different quick paths. This PR moves the call to `cleanupContentsNodesRecursively` earlier so that it's always called.

### DIFF
--- a/yoga/algorithm/CalculateLayout.cpp
+++ b/yoga/algorithm/CalculateLayout.cpp
@@ -1316,6 +1316,10 @@ static void calculateLayoutImpl(
           flexColumnDirection, direction, ownerWidth),
       PhysicalEdge::Bottom);
 
+  // Clean and update all display: contents nodes with a direct path to the
+  // current node as they will not be traversed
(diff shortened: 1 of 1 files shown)


NickGerleman:
What was the chain where we were mutating child before clone?

j-piasecki:
> What was the chain where we were mutating child before clone?

The problem here wasn't mutating before cloning, but leaving a node marked as dirty after the layout has finished.

The exact case in which I found this was a node with a single `display: contents` child which needs to be a leaf. Then in the parent node [this](https://github.com/facebook/yoga/blob/b0b842d5e75d041e3af7e0ac55abfb8929fbbf21/yoga/algorithm/CalculateLayout.cpp#L1339) condition is true, so `cleanupContentsNodesRecursively` doesn't get called and the child node is never visited and cleaned. I assume the same will happen in the other paths with an early return here.

PR #1741, opened 2024-11-13
Revert "Extra log for case where availableHeight is undefined and sizing mode != max content (#1687)"

rozele:
Summary:
Reverting https://github.com/facebook/yoga/pull/1687 as it appears to regress Yoga performance anywhere from 10-33%.

## Changelog

[Internal]

Differential Revision: D65863569

### DIFF
--- a/yoga/algorithm/CalculateLayout.cpp
+++ b/yoga/algorithm/CalculateLayout.cpp
@@ -10,7 +10,6 @@
 #include <cfloat>
 #include <cmath>
 #include <cstring>
-#include <string>
 
 #include <yoga/Yoga.h>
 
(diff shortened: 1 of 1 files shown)


NickGerleman:
@rozele @joevilches out of curiosity were the numbers here derived from the Yoga benchmark in ship mode? (Like fbcode/mode/opt?).

Yoga internal logging functions can take variadic argument list in macro to lazy evaluate this if we wanted, and this is definitely unnecessary work, but I would guess no allocations because we run into short string optimization.

NickGerleman:
Oh, no I missed that we were appending the long C string too… Doing a lot of that does seem dicey… either way. Helpful to add the context (it might be internal, where I can’t see right now on GitHub during PTO)

Issue #1739, opened 2024-11-12
Combining align-content and align-items cause incorrect behavior


tienlocbui1110:
# Report

- [x] I have searched [existing issues](https://github.com/facebook/yoga/issues) and this is not a duplicate

# Issues and Steps to Reproduce
After upgrading Yoga from version 1.19.0 to 3.1.0, I encountered unexpected behavior when using both align-content and align-items. The layout output seems incorrect in the new version. I tested the same layout on Chrome, and Yoga 1.19.0 aligns correctly with Chrome’s behavior, which matches my expectations.

Here is the test code from [Yoga Playground](https://www.yogalayout.dev/playground):
```
<Layout config={{useWebDefaults: true}}>
  <Node style={{
    width: 300,
    height: 300,
    gap: 20,
    flexWrap: "wrap",
    alignContent: "stretch",
    alignItems: "flex-end"
  }}>
    <Node style={{width: 50, height: 50}} />
    <Node style={{width: 50, height: 100}} />
    <Node style={{width: 50, height: 50}} />
    <Node style={{width: 50, height: 50}} />
    <Node style={{width: 50, height: 50}} />
  </Node>
</Layout>
```

# Expected Behavior
You can see the expected behavior in this [JSFiddle](https://jsfiddle.net/q27sx9o5/):
![expected](https://github.com/user-attachments/assets/5ddfacb3-1a97-4c99-81f3-409837206ee2)

# Actual Behavior
The actual output in Yoga 3.1.0 does not match the expected result:
![actual](https://github.com/user-attachments/assets/73f2ec6f-c3c9-4235-956c-53dcaf0121f1)

NickGerleman:
Fixed in Yoga 3.2.0 with above PR 
<img width="1175" alt="image" src="https://github.com/user-attachments/assets/d88c79d7-c4c8-4687-8355-d11f3a4ba56b">

PR #1742, opened 2024-11-14
Fix: Correct Layout Behavior for Combined align-content and align-items

phuccvx12:
This pull request addresses the issue where combining align-content and align-items properties resulted in incorrect layout behavior in Yoga version 3.1.0, as reported in [Issue #1739](https://github.com/facebook/yoga/issues/1739).

# Changes Made:

Alignment Logic Update: Modified the alignment calculations to ensure that the combination of align-content and align-items properties produces the expected layout, consistent with CSS Flexbox standards and previous Yoga versions.

Test Cases Added: Introduced new test cases to cover scenarios involving various combinations of align-content and align-items properties to prevent future regressions.

# Testing:

All existing tests pass successfully.

New test cases confirm that the layout behaves as expected when align-content and align-items are used together.

# Impact:

This fix ensures that layouts using both align-content and align-items properties render correctly, aligning with the behavior observed in Yoga version 1.19.0 and standard web browsers.


NickGerleman:
Thank you for the PR! I’m away until next week, where I could look closer, but @joevilches or @rozele may be able to look in the meantime. We have some existing coverage around here, and this adds more tests, which is great!

NickGerleman:
Triggering the CI, it looks like the generated tests might be out of date with the fixture. That usually means it was run on an earlier revision, or more rarely, that the used version of Chrome in CI comes up with different results (Chrome does sometimes change behavior in edge cases to align closer to the spec).

joevilches:
@phuccvx12 I ran `yarn gentest` locally around 7 hours ago and did not see any tests need to change. I also recently added a change that is also failing. Likely an issue with the CI I can take a look

joevilches:
@phuccvx12  I tried again today and was able to repro the Chrome changes. #1746 should fix this action

PR #1743, opened 2024-11-14
Fix the nodes' owners not being updated when `display: contents` is used

j-piasecki:
In https://github.com/facebook/yoga/pull/1729 I moved the cleanup of `display: contents` nodes to happen before all the early returns, but that change also made it be called **before** `cloneChildrenIfNeeded`. It actually needs to be called after `cloneChildrenIfNeeded` to make sure that children of `display: contents` nodes are properly owned.

It also needs to be called in every short-path, so it's being called in four places in this PR. Please let me know whether it's ok or not.


NickGerleman:
I’m away until next week, but could look closer then. @joevilches @rozele could you take a look in the meantime?

PR #1744, opened 2024-11-16
Guard against seg fault if replaceChild index is out of bounds

joevilches:
Summary:
We are seeing some seg faults after the new display: contents logic was added. This is either because we are passing in an out of bounds index - in which case we try to read `display_` from protected memory. Or, the `Node *` was deleted at some point without removing it from this array. I think its the out of bounds issue mainly because I am not sure where this deletion would occur

I added an if to revert to the legacy, undefined behavior in this case. This is not ideal and we should find the root cause that is calling into this function improperly but for now it stops apps from crashing on the `replaceChild` call

Changelog: [Internal]

Reviewed By: rozele

Differential Revision: D66038645

### DIFF
--- a/yoga/node/Node.cpp
+++ b/yoga/node/Node.cpp
@@ -117,14 +117,20 @@ void Node::setMeasureFunc(YGMeasureFunc measureFunc) {
 }
 
 void Node::replaceChild(Node* child, size_t index) {
-  auto previousChild = children_[index];
-  if (previousChild->style().display() == Display::Contents &&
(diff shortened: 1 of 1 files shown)


NickGerleman:
@joevilches how much of an emergency is this? Silently allowing shadowtree or other Yoga node corruption to happen later just moves the blame of the problem, instead of fixing anything. We should be hard asserting here, like we do in some other places by using “at()” instead of subscript operator.

PR #1753, opened 2024-11-26
Update YGNodeStyleGetGap to return YGValue

heoblitz:
Gap can be styled using both `points` and `percentages`, but YGNodeStyleGetGap currently returns a float value.

To maintain alignment with the `padding` and `margin` functionalities and allow it to be handled in bridging code, this function has been updated to return YGValue.

### DIFF
--- a/capture/NodeToString.cpp
+++ b/capture/NodeToString.cpp
@@ -225,17 +225,17 @@ static void serializeTreeImpl(
     appendEdges<&YGNodeStyleGetPosition>(
         j, "position", node, defaultNode.get());
 
-    appendFloatIfNotDefault(
+    appendYGValueIfNotDefault(
         j["style"],
         "gap",
         YGNodeStyleGetGap(node, YGGutterAll),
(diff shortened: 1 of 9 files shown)


NickGerleman (approved):
Oof. Thanks for noticing this.

It will end up being a breaking change, but we just cut a branch, and I think it would be better to do this then to introduce a new overlap for having the right behavior.

heoblitz:
@NickGerleman I accidentally closed the PR while rebasing.

I've combined the previous commits into one. Please take a look when you have time. Thank you.

NickGerleman:
We seem to be crashing with `terminating due to uncaught exception of type facebook::yoga::vanillajni::YogaJniException: std::exception`

Seems like we still need to update method descriptor for `registerNatives`? `{"jni_YGNodeStyleGetGapJNI", "(JI)F", (void*)jni_YGNodeStyleGetGapJNI},` (it says it's returning a float right now).

heoblitz:
Oh, I missed that. I've corrected it properly.

Issue #1760, opened 2024-12-02
[JavaScript] - `"node"` module resolution


nyan-left:
# Report
- [x] I have searched existing issues and this is not a duplicate

# Issues and Steps to Reproduce

When using Node module resolution, imports from subpaths like `yoga-layout/load` fail:

```ts
import { Overflow } from 'yoga-layout/load'  // fails
```

This occurs because the package.json uses exports that require bundler module resolution
(https://github.com/facebook/yoga/blob/main/javascript/package.json#L15):
```json
"exports": {
    ".": "./src/index.ts",
    "./load": "./src/load.ts"
}
```

Many projects and libraries need to maintain compatibility with Node module resolution to support their users' existing codebases and build setups.
While direct application developers can potentially switch to "bundler" module resolution as a quick workaround, this becomes problematic for library authors building on top of yoga-layout. These library authors need to maintain the widest possible compatibility for their downstream users, making it difficult to require specific module resolution strategies.

Bundling yoga-layout as part of our library is not a viable solution as we want to leverage the existing npm ecosystem - allowing users to receive yoga-layout updates independently, maintain their own version management, and use standard module resolution patterns they're familiar with.

# Expected Behavior

Imports from `yoga-layout/load` should resolve correctly when using `node` module resolution, ensuring broader compatibility across different projects and build configurations.

# Actual Behavior

Get TypeScript error:

```
error TS2307: Cannot find module 'yoga-layout/load' or its corresponding type declarations.
```

NickGerleman:
`node16` resolution should also work (with the caveats that option has more generally). That has been the resolution for Node 16 in `@tsconfig/bases` [since May 2023](https://github.com/tsconfig/bases/commit/47d3c29c650091c9e2eaad27bd1133db63f0fd22) (which is a bit more recent than I expected).

Yoga took a dependence on ES modules for 3.0 to simplify the number of configurations, but that means type checker and bundler need to have some awareness of them. We removed the dependency on top-level await, because it was a bit too breaking, but the ecosystem is pretty increasingly depending on ES module semantics and resolution, so this is a place where projects really should try to put in the work to get a toolchain supporting them.

We export the main entrypoint outside package.json exports, but reaching into `yoga-layout/dist/src/load` would probably be the best way to get things working with old module resolution in the meantime.

nyan-left:
Thanks for the detailed response. I agree that upgrading to modern module resolution is ideal, though it's not always possible in current projects.

[...]

PR #1764, opened 2024-12-06
Fix/typescript 4 node module resolution

nyan-left:
Closes https://github.com/facebook/yoga/issues/1760

This PR adds support for TypeScript 4.x and `"node"` module resolution by fixing import extension handling. Previously, the package only worked with TypeScript 5.x due to `.ts` extensions in declaration files causing `TS2691` errors in earlier versions.

## Changes
- Changed imports from `.ts` to `.js`, configured in:
  - `Jest` - `moduleNameMapper`
  - `Docusaurus/Webpack` - `extensionAlias`
- Added `tsconfig.load.json` for the `/load` entry point
- Updated `just` build process:
  - Temporarily rewrites `load.ts` imports to point to `dist/src` 
  - Generates artifacts and restores source references
- Updated `package.json` to include `load.*` files

### DIFF
--- a/javascript/just.config.cjs
+++ b/javascript/just.config.cjs
@@ -84,7 +84,7 @@ task(
       declarationDir: 'dist',
     }),
     babelTransformTask({src: 'src', dst: 'dist/src'}),
-    transformLoadTask(),
+    transformLoadEntryPointTask(),
     'prepack-package-json',
   ),
 );
@@ -98,29 +98,16 @@ function recursiveReplace(obj, pattern, replacement) {
     }
   }
 }
-
-function transformLoadTask() {
+function transformLoadEntryPointTask() {
   return async () => {
-    const js = {
-      content:
-        'export * from "./dist/src/load.js";\n//# sourceMappingURL=load.js.map\n',
-    };
-    await writeFile('load.js', js.content);
+    await tscTask({
+      project: 'tsconfig.load.json',
+    })();
 
-    const sourceMap = {
-      version: 3,
-      file: 'load.js',
-      names: [],
-      sources: ['./src/load.ts'],
-      sourcesContent: [js.content],
-      mappings: 'AAAA,cAAc,YAAY',
-    };
-    await writeFile('load.js.map', JSON.stringify(sourceMap));
-
-    const dts = {
-      content: 'export * from "./dist/src/load.js";\n',
-    };
-    await writeFile('load.d.ts', dts.content);
+    await babelTransformTask({
+      src: 'load.ts',
+      dst: '.',
+    })();
   };
 }
 
--- a/javascript/just.config.cjs
+++ b/javascript/just.config.cjs
@@ -113,7 +113,7 @@ function transformLoadTask() {
       names: [],
       sources: ['./src/load.ts'],
       sourcesContent: [js.content],
-      mappings: 'AAAA,cAAc,eAAe',
+      mappings: 'AAAA,cAAc,YAAY',
     };
     await writeFile('load.js.map', JSON.stringify(sourceMap));
 
--- a/javascript/just.config.cjs
+++ b/javascript/just.config.cjs
@@ -84,6 +84,7 @@ task(
       declarationDir: 'dist',
     }),
     babelTransformTask({src: 'src', dst: 'dist/src'}),
+    transformLoadTask(),
     'prepack-package-json',
   ),
 );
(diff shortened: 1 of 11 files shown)


javascript/just.config.cjs
```diff
@@ -98,6 +99,31 @@ function recursiveReplace(obj, pattern, replacement) {
...
 
+function transformLoadTask() {
+  return async () => {
+    const js = {
+      content:
+        'export * from "./dist/src/load.js";\n//# sourceMappingURL=load.js.map\n',
+    };
+    await writeFile('load.js', js.content);
+
+    const sourceMap = {
```
NickGerleman, line 110: Having this step hardcode the entrypoint, and a sourcemap for it, doesn't seem very maintainable or resilient to future changes (especially since all the tests use the esm version).

I think we would want to include this new file in existing transformation and declaration generation instead.

NickGerleman (changes requested):
The `.js` renaming seems to break resolution in the website build when Webpack tries to find nonexistent .js file.

```
Module not found: Error: Can't resolve './wrapAssembly.js' in '/vercel/path0/javascript/src'
```

Also seems to break resolution in test runner. 

Let's hold off on doing anything too inorganic to support TypeScript < 5.0, since it's no longer in support.

javascript/package.json
```diff
@@ -19,7 +19,9 @@
   "files": [
     "dist/binaries/**",
     "dist/src/**",
-    "src/**"
+    "src/**",
+    "load.js",
+    "load.js.map"
```
NickGerleman, line 24: I think this is missing `load.d.ts`

nyan-left: Missed a commit, thanks for spotting this.

nyan-left:
> The `.js` renaming seems to break resolution in the website build when Webpack tries to find nonexistent .js file.
> 
> ```
> Module not found: Error: Can't resolve './wrapAssembly.js' in '/vercel/path0/javascript/src'
> ```
> 
> Also seems to break resolution in test runner.
> 
> Let's hold off on doing anything too inorganic to support TypeScript < 5.0, since it's no longer in support.

These two are resolved with a bit of config for [jest](https://github.com/facebook/yoga/pull/1764/files#diff-0dccf577be79696283965bb52ade209c345c2941acb0d41ab9ea602073ec77e0) and [docusaurus/weback. ](https://github.com/facebook/yoga/pull/1764/files#diff-10f68cce4f632eff26e4811bde30e1dece41a54b17251ce00b8248508b992866)

nyan-left:
@NickGerleman 

Regarding linting, given the small nature of the `/load.ts`, are you happy with ignoring it in eslint?

I'm not too keen on hard coding the tsconfig paths, and afaik a more modular approach like [this](https://github.com/typescript-eslint/typescript-eslint/issues/7383) doesn't currently exist.

NickGerleman:
Changing the resolvers in each of these is clever 🙂.

It is also the kind of debt that I would like to avoid for TS versions out of support. It means that, any time in the future, if someone adds a `.ts` import, to some files, but not others, we silently break compatibility, with no indications or tests in the like. I would not like us to change resolution of specific parts of the repo to treat `.ts` files as `.js`, for the sake of versions which are out of support.

I'm less concerned with the changes to avoid the need for ESM resolution, since they are still relevant in supported TS versions, and are less likely to break with future code changes.

It looks like now, ESLint is failing, because `load.ts` is using a separate `tsconfig` from the rest of the repo. We could specialize this specific file in the ESLint config, or we could maybe keep everything in the same project, and just add a step to copy the declaration for this entrypoiny?

nyan-left:
> It is also the kind of debt that I would like to avoid for TS versions out of support. It means that, any time in the future, if someone adds a .ts import, to some files, but not others, we silently break compatibility, with no indications or tests in the like. 

I think we can solve this via `"allowImportingTsExtensions": false` in `javascript/tsconfig.json`. The implication of this would be that we need to tweak generated tests to import from `.js` or give jest it's own tsconfig. Alternative could also be some eslint extensions setup. Both, `lint` and `tsc` happen within the `validate-js` ci job, so that'll be caught. 


Regarding resolvers, it's the reccommended approach by the webpack team. Looks like the only reason it's not the default in webpack already is because webpack doesn't support ts "out of the box". Perhaps one to reporo and make an issue with docusaurus to see their stance.

https://github.com/webpack/webpack/issues/13252#issuecomment-1170290316
https://github.com/webpack/enhanced-resolve/pull/351#issue-1285504009


Jest suggest using `ts-jest` in TypeScript projects. While I've not personally tested `.js` extension resolution in it, I'd assume this would without setup. 
https://github.com/jestjs/jest?tab=readme-ov-file#using-typescript

nyan-left:
Moving the load.d.ts works for me! :)

nyan-left:
Hi @NickGerleman,

I'm wondering if you have had any further thoughts regarding the `allowImportingTsExtensions` approach to address your concerns. TypeScript 4 still has millions of weekly downloads 😓 , and we can prevent any silent breakage through compile-time checks.

NickGerleman:
My stance is that it’s not worth future complexity to add support for TypeScript 4, especially as other popular libraries are dropping it as well (not to mention DT no longer testing it). https://github.com/uuidjs/uuid/pull/837

It’s the same reason we don’t do things like keep around workarounds to support Visual Studio 2015, or support ancient toolchains for microcontrollers (though folks have asked for both). We end up will all these bespoke and usually untested configurations, and no long term path to remove what was added to support them. So we have to cut a line somewhere, where “support level” of the tool chain is usually the best proxy.

nyan-left:
Thanks for your thoughts. Closing the PR - will explore `patch-package` or forking instead.

PR #1768, opened 2024-12-11
Fix test util to measure text properly based on flex direction

joevilches:
Summary: Depending on the flex direction text will either be capped to the measured size or to the longest word, so I added that functionality

Differential Revision: D67106199


nicoburns:
@joevilches I don't think this is right. I believe the measure function should still return the size floored by the longest word size, but the parent container should override that size due to the max-width style (which probably is different for flexbox row vs. column due to Flexbox having a weird quirk around min/max sizes in the main axis). As a general principle, the min- or max- content size of box is never affected by the parent's display mode (although the resultant size of the box may be).

joevilches:
@nicoburns that makes sense thanks for commenting. I am curious how Yoga would yield the same results as chrome in this case, since it did seem like at some point we would need to measure with some smaller width than the longest word to get the correct height. I can take a look again, maybe the issue is something else.

nicoburns:
@joevilches I think I previously misunderstood this code (not helped by the fact I was trying to read it on my phone!). The width you are using does seem broadly correct to me with default styles. Although the way we deal with this in Taffy is by using (what Yoga calls) an "Exact" sizing constraint in the column case (depending on `align-items`)

I think the reason this is (/seems to be) different for columns vs. rows is because `align-items` defaults to `stretch`. Which will mean that with default styles the min-content width will overridden for a column. I believe this wouldn't apply with `align-items` overridden to `start` or similar.

The following HTML:

```html
<div style="display: flex;width: 50px;flex-direction: column;">
        <div>
            orheg irehg ireh girehg iorhioh iroegh 
            jhnreuighuirebghuierhguierbuigbreuibuebgiuerbgiuberuigberuibguirebg
            righierhg ihe irogh eriohg iorehg ioerhg iorehg 
        </div>
</div>
```

Results in:

Screenshot 2024-12-24 at 14 19 23

But if you add an `align-items: start` style to the outer div then you get:

Screenshot 2024-12-24 at 14 19 13

PR #1779, opened 2025-01-21
Karim/exclude non yoga search paths

karim-alweheshy:
in rules_swift_package_manager this is necessary to exclude these from search path for public headers that will fail to compile eventually


NickGerleman:
Could you help provide a bit more context? What error are you seeing?

We have a modulemap for generated umbrella header for Swift usage with the list of public header: https://github.com/facebook/yoga/blob/main/yoga/module.modulemap Wonder if `rules_swift_package_manager` might not be picking this up?

includes within the target shouldn't ever reference anything outside the `yoga` directory then.

karim-alweheshy:
> Could you help provide a bit more context? What error are you seeing?
> 
> We have a modulemap for generated umbrella header for Swift usage with the list of public header: https://github.com/facebook/yoga/blob/main/yoga/module.modulemap Wonder if `rules_swift_package_manager` might not be picking this up?
> 
> includes within the target shouldn't ever reference anything outside the `yoga` directory then.

Because the public headers search path is set to be on the whole repo, rules_swift_package_manager compiles all public headers from all folders. This compilation fails because imported symbols in these headers are not part of the source.

Excluding the other folders enables rules_swift_package_manager to use the public header only in the `yoga` folder and not the rest

NickGerleman:
Describing the set of public headers, for the purpose of creating an umbrella header, is what the modulemap is for, at least when consuming as a Clang module. Or is `rules_swift_package_manager` combining headers for some non Clang module purpose? 

Getting the SwiftPM package to work with Bazel seems like good goal, but this behavior seems funky compared to what SwiftPM itself does/expects. Trying to understand more, to know if maybe this is a bug with `rules_swift_package_manager` that should be reported.

karim-alweheshy:
> Describing the set of public headers, for the purpose of creating an umbrella header, is what the modulemap is for, at least when consuming as a Clang module. Or is `rules_swift_package_manager` combining headers for some non Clang module purpose? Having a copy of the error would be helpful here.
> 
> Getting the SwiftPM package to work with Bazel seems like good goal, but this behavior seems funky compared to what SwiftPM itself does/expects. Trying to understand more, to know if maybe this is a bug with `rules_swift_package_manager` that should be reported.

There is a difference between how rules_spm work vs SPM itself

I want to double click on this issue specifically, do you agree that the public headers search path for the `core` target should exclude the mentioned folders?

NickGerleman:
Sorry for the long delay in response!

It's unfortunate that `rules_spm` is not respecting modulemaps. I'm a bit surprised there isn't a way to tell it to. I would definitely recommend creating an issue for them, if there really isn't a way to even explicitly tell it to use Yoga's modulemap.

We can add workarounds to Yoga for this, so long as we have a way to check that they aren't broken in the future. E.g. if we had a CI rule that could check that things worked, I would feel a lot better about this. One-off changes which may break in the future with unrelated changes, I have been more reluctant to add though.

PR #1793, opened 2025-03-02
allow for loading wasm without unsafe wasm execution

### LINKED ISSUE #1767: Request to Remove or Address `unsafe-eval` Usage in Yoga Layout
Hi Yoga Layout Team,

I am currently using `@react-pdf/renderer` (version `^4.1.5`) in my project, which depends on `yoga-layout`. During development, I've encountered issues related to the usage of `unsafe-eval` within `yoga-layout`.

### Problem
The inclusion of `unsafe-eval` poses significant security risks and violates stricter Content Security Policies (CSPs), which are becoming increasingly common in modern web applications. This makes it challenging to use `yoga-layout` in secure environments without relaxing CSP rules, which isn't always an acceptable solution.

### Steps to Reproduce
1. Integrate `@react-pdf/renderer` (which uses `yoga-layout`) into a project.
2. Implement a CSP policy that disallows `unsafe-eval`.
3. Observe the resulting errors when attempting to use `yoga-layout`.

Let me know if additional details, examples, or assistance with refactoring would be helpful. Thank you for all the effort you’ve put into maintaining this essential library!

Best regards,  
Michele

cramt:
fixes #1767, by allowed you to not bundle the wasm as a base64 encoded string which is dynamically loaded at eval time

### DIFF
--- a/javascript/CMakeLists.txt
+++ b/javascript/CMakeLists.txt
@@ -29,6 +29,12 @@ set(COMPILE_OPTIONS
 
 add_compile_options(${COMPILE_OPTIONS})
 
+link_libraries(embind)
+
+add_library(yogaObjLib OBJECT ${SOURCES})
+
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/binaries)
+
 add_link_options(
     ${COMPILE_OPTIONS}
     "SHELL:--closure 1"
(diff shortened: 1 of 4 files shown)


javascript/CMakeLists.txt
```diff
@@ -45,15 +51,27 @@ add_link_options(
...
-
-set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/binaries)
+add_link_options(
+    ${COMPILE_OPTIONS}
+    "SHELL:--closure 1"
+    "SHELL:--memory-init-file 0"
+    "SHELL:--no-entry"
+    "SHELL:-s ALLOW_MEMORY_GROWTH=1"
+    "SHELL:-s ASSERTIONS=0"
+    "SHELL:-s DYNAMIC_EXECUTION=0"
```
NickGerleman, line 65: From what I learned in the other threads, `DYNAMIC_EXECUTION=0` might be making the bindings quite a bit slower, if we still end up using it anyways.

Might be worth removing this argument from the base64 variant, or seeing what happens if we add `EMBIND_AOT` to both (might require bumping emsdk here https://github.com/facebook/yoga/blob/1b7d2c8d48e83fb136eccc5215faeb8d54da9c3e/javascript/just.config.cjs#L140)

javascript/CMakeLists.txt
```diff
@@ -45,15 +51,27 @@ add_link_options(
...
 
-link_libraries(embind)
+add_executable(yoga-wasm-separate-esm $<TARGET_OBJECTS:yogaObjLib>)
 
-add_library(yogaObjLib OBJECT ${SOURCES})
-
-set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/binaries)
+add_link_options(
+    ${COMPILE_OPTIONS}
+    "SHELL:--closure 1"
```
NickGerleman, line 60: Could we avoid duplicating the bits here that are not different?

PR #1802, opened 2025-03-19
LayoutData - added explicit default fields values initialization

inobelar:
Explicit defaults add 'strictness' and take the guesswork out of what's going on in places like this:

https://github.com/facebook/yoga/blob/6455a848a76f433bdb48b2640b7f4644774c76fd/yoga/algorithm/CalculateLayout.cpp#L2345

### DIFF
--- a/yoga/event/event.h
+++ b/yoga/event/event.h
@@ -43,7 +43,7 @@ struct LayoutData {
   int cachedMeasures = 0;
   int measureCallbacks = 0;
   std::array<int, static_cast<uint8_t>(LayoutPassReason::COUNT)>
-      measureCallbackReasonsCount = {{0}};
+      measureCallbackReasonsCount;
 };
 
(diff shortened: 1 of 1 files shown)


yoga/event/event.h
```diff
@@ -36,14 +36,14 @@ enum struct LayoutPassReason : int {
...
-  int measureCallbacks;
+  int layouts = 0;
+  int measures = 0;
+  uint32_t maxMeasureCache = 0;
+  int cachedLayouts = 0;
+  int cachedMeasures = 0;
+  int measureCallbacks = 0;
   std::array<int, static_cast<uint8_t>(LayoutPassReason::COUNT)>
-      measureCallbackReasonsCount;
+      measureCallbackReasonsCount = {{0}};
```
NickGerleman, line 46: nit: this is already default initialized

```suggestion
      measureCallbackReasonsCount;
```

NickGerleman (approved):

Issue #1574, opened 2024-02-07
Why can't we round node width/height directly in roundLayoutResultsToPixelGrid()?


kennethpu:
# Report

- [x] I have searched [existing issues](https://github.com/facebook/yoga/issues) and this is not a duplicate

In [PixelGrid.roundLayoutResultsToPixelGrid()](https://github.com/facebook/yoga/blob/a37565f70d74ecc1a312a62dc48091fd2c73136e/yoga/algorithm/PixelGrid.cpp#L65C6-L65C35) it appears that we are calculating a node's rounded width/height as follows:

1. Calculate the start position of the node.
```
const double absoluteNodeLeft = absoluteLeft + nodeLeft;
```
2. Using the start position and computed width, calculate the end position of the node.
```
const double absoluteNodeRight = absoluteNodeLeft + nodeWidth;
```
3. Calculate rounded node width/height as the difference between the rounded start/end positions of the node.
```
node->setLayoutDimension(
    roundValueToPixelGrid(
        absoluteNodeRight, 
        pointScaleFactor, 
        (textRounding && hasFractionalWidth), 
        (textRounding && !hasFractionalWidth)
    ) -
    roundValueToPixelGrid(
        absoluteNodeLeft, 
        pointScaleFactor, 
        false, 
        textRounding
     ),
     Dimension::Width);
```
It's not clear to me why we need to calculate rounded width/height in this way as opposed to rounding the width/height values directly, and it also seems like this may introduce rounding errors as noted in this previously opened issue (https://github.com/facebook/yoga/issues/901). Curious if the [proposed solution](https://github.com/facebook/yoga/pull/902/commits/19035c90e9107226cc53d91f6722ee331339624a) on that issue has merit or if there are other issues not being considered. 

As an iOS developer utilizing a fork of Yoga (really appreciate this project btw!) I'm also curious if you foresee any issues with tweaking the rounding logic on our fork to more closely mirror [Apple's strategy](https://developer.apple.com/documentation/coregraphics/1456348-cgrectintegral) of rounding down origin values and rounding up size values to the nearest pixel boundary. Are there other parts of the layout algorithm we would need to update for consistency?

NickGerleman:
I would love to change how layout rounding works in terms of how it's executed, but haven't put a lot of thought into it's underlying rounding logic. We have seen real bugs from it.

A goal of this style of rounding is to not introduce "gaps", so it makes sense for the start position reference to be the previous rounded pixel I think.

[...]

PR #1860, opened 2025-10-14
Fix layout bug caused by floating point rounding error

### LINKED ISSUE #749: Layout bug caused by floating point rounding error
# Report

- Layout multiple layers of views, and put a label as leaf node.
- Run it in a 3x resolution screen, like iPhone X or one of the Plus models.

# Expected Behavior
The text on the label should be displayed without problem

# Actual Behavior
The text got truncated.
screen shot 2018-04-08 at 20 30 26

# Link to Code
Example project to reproduce the issue https://github.com/leafduo/yoga-rounding

image

The problem was caused by the accumulation of floating point rounding error in complex layout. The view tree in the example has a height of 4, and causes floating point rounding errors to accumulate. In `YGRoundValueToPixelGrid`, the `fractial`variable is 0.999877929 and is considered not a whole integer. In the end, the width of the label is one pixel short, causing the text truncation.

This issue is not presented when `pointScaleFactor` is 3 (not 1 or 2), and perhaps it's because 1) 3 is bigger and will cause more error, and 2) multiply and division by 2 is pretty accurate because the binary nature of computer.

Using `double` instead of `float` to represent points, or raising the threshold in `YGFloatsEqual` will make the text fully displayed in the example, but not solve this problem completely. It will still exist in a more complex layout.

I think maybe we can store scaled value or use a real fraction representation?

matthieugicquel:
Closes #749 and possibly #1574

The current layout algorithm sometimes rounds down by a small value the size of text nodes, which makes the renderer (at least iOS) skip a line of text, resulting in buggy layout like this:

482035993-9f93908b-7068-49eb-aae2-759b55294ced

I don't have a deep enough understanding of the algorithm to know if this fix is "correct", but what I can say is that:
- it didn't break existing tests
- we didn't get any bug reports while using it in a production app

I only applied the change to the height calculation as I only encountered the issue on height, but I suppose width could benefit from a similar change.

### DIFF
--- a/tests/YGRoundingFunctionTest.cpp
+++ b/tests/YGRoundingFunctionTest.cpp
@@ -181,3 +181,27 @@ TEST(YogaTest, raw_layout_dimensions) {
 
   YGConfigFree(config);
 }
+
+TEST(YogaTest, roundLayoutResultsToPixelGrid_height_rounding_up) {
+  YGConfigRef config = YGConfigNew();
+  YGConfigSetPointScaleFactor(config, 3);
+
+  YGNodeRef node = YGNodeNewWithConfig(config);
+  YGNodeStyleSetPositionType(node, YGPositionTypeAbsolute);
+
+  // These are values extracted from a debugging session in a real iOS app
+  YGNodeStyleSetPosition(node, YGEdgeLeft, 38.333333969116211);
+  YGNodeStyleSetPosition(node, YGEdgeTop, 1970.3333333432674);
+  YGNodeStyleSetWidth(node, 339.66665649414063);
+  YGNodeStyleSetHeight(node, 96);
+  YGNodeSetNodeType(node, YGNodeTypeText);
+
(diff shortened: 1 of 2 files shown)


dylanclarke890:
I tested this and though it works I also managed to fix the issue by doing the pixel calculations with doubles instead of floats which 'feels' more correct to me, I think this issue simply stems from lack of precision with floats; when subtracting to get the width/height the actual value drifts because of the combined inaccuracies of both floats. Having `RoundValueToPixelGrid` return a double instead of casting to a float early keeps the correct precision for the maths *then* you can cast to a float without losing information.

IgorTodorovskiIBM:
Can we create a new release of yoga? see https://github.com/facebook/yoga/issues/1859

matthieugicquel:
This PR in react-native seems to fix the same issue (it's been merged then reverted) https://github.com/facebook/react-native/pull/54260

fbatiga:
Any news on this MR ?

PR #1894, opened 2026-02-25
CSS Grid 2/9: Grid layout algorithm

intergalacticspacehighway:
## Summary
- Core grid layout computation and dispatch integration
- Includes `AutoPlacement.h`, `GridLayout.h/cpp`, `TrackSizing.h`
- `CalculateLayout.cpp` grid dispatch, `CMakeLists.txt` updates

Split from #1865. Depends on #1893. Merge after 1/9.


yoga/algorithm/grid/GridTrack.h
```diff
@@ -0,0 +1,23 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#pragma once
+
+#include <yoga/style/GridTrack.h>
+
+namespace facebook::yoga {
+
+struct GridTrack : GridTrackSize {
```
NickGerleman: Could we document, what each of these mean? Why we keep this extra state/the difference between `GridTrack`/`GridTrackSize`? From naming, etc, it's not immediately clear.

yoga/algorithm/grid/AutoPlacement.h
```diff
@@ -0,0 +1,585 @@
...
+#include <cstdint>
+#include <map>
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
+
+namespace facebook::yoga {
+
+struct OccupancyGrid {
+  std::unordered_map<int32_t, std::vector<std::pair<int32_t, int32_t>>>
```
NickGerleman: `std::unordered_map` is quite expensive, heap allocating a node per inserted item. How many items is this usually be? Would it be better as a flat vector of items/binary searched vector? (sidenote, this is what `std::flat_map` is kinda meant to solve).

NickGerleman (commented):
Been looking through this, for about an hour.

I think we still need to chunk this change into steps, of smaller concerns, to be able to give it a real review (vs trivial nits) which I feel like I ought to do, if we want to maintain this long term. It's been tricky for me to understand how everything connects. That might be an indication, it's worth documenting some of the newly added structures, if what they're doing isn't intuitive.

Could we build this up, in small parts of spec/behavior/structures at a time?

intergalacticspacehighway:
> Could we build this up, in small parts of spec/behavior/structures at a time?

Sure @NickGerleman. yeah i thought this might be needed 😅. Will split it and also address above mentioned points in them.

intergalacticspacehighway:
@rozele I will split this PR further into chunks. Thanks for importing 1/N.

PR #1960, opened 2026-05-15
ci: declare workflow-level `contents: read` on 6 workflows

arpitjain099:
Pins the default `GITHUB_TOKEN` to `contents: read` on 6 workflows in `.github/workflows/` that don't call a GitHub API beyond the initial checkout.

## Why

CVE-2025-30066 (March 2025 `tj-actions/changed-files` supply-chain compromise) exfiltrated `GITHUB_TOKEN` from workflow logs. Pinning per workflow caps runtime authority irrespective of the repo or org default, gives drift protection if the default ever widens, and is credited per-file by the OpenSSF Scorecard `Token-Permissions` check.

YAML validated locally with `yaml.safe_load` on each touched file.

### DIFF
--- a/.github/workflows/validate-android.yml
+++ b/.github/workflows/validate-android.yml
@@ -8,6 +8,9 @@ on:
       - 'release-*'
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
   build:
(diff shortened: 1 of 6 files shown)


cortinico (approved):

PR #1963, opened 2026-05-15
Fix Kotlin property access syntax in Yoga Java test files

CalixTang:
Summary: Recommit of D105191609, which was reverted (likely because github export hadn't completed when it tried to land)

Differential Revision: D105335609


cortinico (approved):

PR #1975, opened 2026-06-12
Fix text truncation caused by pixel-grid rounding of near-integer text nodes

stleamist:
## Summary

Fixes a text truncation bug in `roundLayoutResultsToPixelGrid()`: when a text node's scaled dimension is close to a whole number (`hasFractionalWidth == false`), the trailing edge was force-floored. Because the `inexactEquals` tolerance (0.0001) can apply asymmetrically to the two edges of the same node, the leading edge may snap up while the trailing edge gets floored, producing a final width smaller than the measured text width — so the text renders with an ellipsis.

## Real-world case

Observed with a right-aligned `UILabel` laid out via FlexLayout on an @3x display (393pt-wide device). The measured text width was `54.666656`, positioned at `absoluteNodeLeft = 163.666641`:

| Value | × 3 (scale) | distance from whole number | within 0.0001 tolerance? | result |
|---|---|---|---|---|
| `nodeWidth = 54.666656` | 163.999968 | 0.000032 | yes → `hasFractionalWidth = false` | `forceFloor` enabled |
| `absoluteNodeLeft = 163.666641` | 490.999923 | 0.000077 | yes | snaps **up** to 491 |
| `absoluteNodeRight = 218.333297` | 654.999891 | 0.000109 | **no** | floored **down** to 654 |

Final width = (654 − 491) / 3 = **54.333…** < measured **54.666656** → truncated.

Both edges sit essentially on the same pixel boundary, but a 0.00003 difference makes them round in opposite directions, shrinking the node by a full pixel.

## Fix

Never force-floor the trailing edge of a text node:

- `hasFractionalWidth == true` (clearly fractional): keep the existing force-ceil behavior.
- `hasFractionalWidth == false` (near-integer): fall through to natural rounding, which snaps 654.999891 up to 655 and preserves the measured width.

This also matches the intent of the existing comment in the function: *"If a node has a custom measure function we never want to round down its size as this could lead to unwanted text truncation."*

## Tests

- All existing tests pass unchanged (831 passed, 46 skipped, 0 failed).
- Adds `tests/YGRoundingTextTruncationTest.cpp` with four tests, including one reproducing the exact values from the real-world case above (fails before the fix, passes after).

## Related issues

- #1860 addresses the same class of bug for height only and notes that "a similar change for width may be helpful" — this PR covers both width and height at the rounding call sites.
- Likely also addresses #749 and #1574, which report measured text being rounded down by the pixel grid.

### DIFF
--- a/tests/YGRoundingTextTruncationTest.cpp
+++ b/tests/YGRoundingTextTruncationTest.cpp
@@ -1,0 +1,187 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#include <gtest/gtest.h>
+#include <yoga/Yoga.h>
+
+// This test reproduces a text truncation bug that occurs when:
+// 1. A text node is positioned at an offset where absoluteNodeLeft * scale
+//    has a fractional part very close to 1.0 (within 0.0001)
+// 2. The text node width * scale also has fractional part close to 1.0
+//    (hasFractionalWidth = false)
+// 3. But absoluteNodeRight * scale has fractional part that is close to
+//    but NOT within 0.0001 of 1.0
+//
+// In the buggy code, when hasFractionalWidth = false and textRounding = true,
+// forceFloor was set to true. This caused absoluteNodeRight to be floored,
+// while absoluteNodeLeft was correctly rounded up (due to inexactEquals check).
+// The mismatch caused the final width to be smaller than the measured width.
+//
+// Example from real-world bug:
+// - absoluteNodeLeft = 163.666641, * 3 = 490.999923 (fmod = 0.999923)
+//   |1.0 - 0.999923| = 0.000077 < 0.0001 → rounds UP to 491
+// - absoluteNodeRight = 218.333298, * 3 = 654.999894 (fmod = 0.999894)
+//   |1.0 - 0.999894| = 0.000106 > 0.0001 → floors DOWN to 654 (BUG!)
+// - Final width = 654/3 - 491/3 = 163/3 = 54.333... < measured 54.666... (TRUNCATED!)
+
+// Measure function that returns width causing the truncation bug
+// The exact value from the real-world bug case
+static YGSize measureTextForTruncationBug(
+    YGNodeConstRef /*node*/,
+    float /*width*/,
+    YGMeasureMode /*widthMode*/,
+    float /*height*/,
+    YGMeasureMode /*heightMode*/) {
+  // From the real bug: nodeWidth = 54.666656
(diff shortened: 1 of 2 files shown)


stleamist:
I've put together a minimal iOS reproduction of this bug that uses nothing but the yoga C API and a single `UILabel`, with snapshot tests: **https://github.com/stleamist/YogaTruncationRepro**

The label is measured by a yoga measure function and laid out on a @3x pixel grid (iPhone 15 Pro simulator) at an offset where the node's left edge scales to just *inside* the 0.0001 `inexactEquals` tolerance of a whole pixel (so it snaps up) while the right edge lands just *outside* it (so it gets force-floored). Yoga assigns the label exactly one pixel less than what its measure function returned, and the label renders with an ellipsis:

```
XCTAssertGreaterThanOrEqual failed: ("45.0") is less than ("45.33332266666667")
```

| yoga `main` (`f6206ec3`) | with this PR |
|---|---|
| truncated | not truncated |

The core of the trigger (full code in the repo):

```objc
// Measure the real label, then nudge the width up to the next value whose
// scaled size sits just inside the tolerance of a whole pixel — the same
// shape of value that real text measurement produces naturally. This makes
// hasFractionalWidth == false, enabling the force-floor path.
CGSize fitted = [view.label sizeThatFits:CGSizeMake(CGFLOAT_MAX, CGFLOAT_MAX)];
double wholePixels = ceil(fitted.width * 3.0);
double measuredWidth = (wholePixels - 0.000032) / 3.0;

// Left edge: scales to 490.999923, *within* 0.0001 of 491 → snaps UP.
// Right edge (left + width): ends up ~0.000109 away from a whole pixel,
// *outside* the tolerance → force-floored DOWN. The node loses one pixel.
YGNodeStyleSetPadding(root, YGEdgeLeft, (float)((491.0 - 0.000077) / 3.0));
```

The package pins yoga to current `main` (`f6206ec3`), so both tests fail out of the box. Swapping the dependency for this PR's branch (a one-line change noted in `Package.swift`) makes both pass. The snapshot test's committed reference is the *correct* rendering, so its failure diff visualizes the truncation directly.

stleamist:
Hi @NickGerleman — hope it's okay to tag you directly on this.

I realize Yoga's changes mostly flow through Meta's internal review, so this PR may just be waiting on someone to pick it up. Since it lands in the pixel-grid rounding code you maintain, I wanted to flag it — and I've tried to make it as low-effort to verify as possible:

- A minimal iOS repro using only the yoga C API + one `UILabel`, with snapshot tests that fail on `main` and pass with this PR: https://github.com/stleamist/YogaTruncationRepro
- The fix covers both width and height.
- All 831 existing tests pass (46 skipped, 0 failed), plus 4 new regression tests for the offset/tolerance edge case.

No urgency at all — happy to follow whatever process works best on your end, and to rebase or adjust however helps.

PR #1976, opened 2026-06-16
fix: yoga crash with display:none

javache:
Summary:
It fixes this issue: https://github.com/react/react-native/issues/52349
Basically, in some rare cases, having an element with `display:none` style caused the app to crash.

It was caused by a stale `hasNewLayout` flag on the hidden view
That flag was always set in `computeFlexBasisForChildren` for elements with `display:none` style, but could be never consumed/reset, because of a cache hit or something.
Since such elements contribute nothing to the layout sizes, I made the change to actually touch them only during actual layout passes

## Changelog:

[GENERAL] [FIXED] - Crash: YogaLayoutableShadowNode.cpp: function layout: assertion failed (YGNodeGetOwner(childYogaNode) == &yogaNode_) https://github.com/react/react-native/issues/52349

X-link: https://github.com/react/react-native/pull/57197

Reviewed By: christophpurrer

Differential Revision: D108796888

Pulled By: javache

### DIFF
--- a/yoga/algorithm/CalculateLayout.cpp
+++ b/yoga/algorithm/CalculateLayout.cpp
@@ -615,9 +615,18 @@ static float computeFlexBasisForChildren(
   for (auto child : children) {
     child->processDimensions();
     if (child->style().display() == Display::None) {
-      zeroOutLayoutRecursively(child);
-      child->setHasNewLayout(true);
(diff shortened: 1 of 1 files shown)


KrastanD:
Anyone know what version of react native this will be released in?

PR #1977, opened 2026-06-17
Fix testLoggerLeak by replacing Kotlin SAM lambda with anonymous object

ellemac123:
Summary:
Yoga test is failing after migration from java to kotlin due to the `testLoggerLeak` failing. The test creates a no-op `YogaLogger`, sets+unsets it on a config, drops all references, then asserts that a `WeakReference` to the logger gets cleared by GC. The Kotlin SAM lambda form (`YogaLogger { level, message -> }`) ends up with an implicit `this$0` capture of the enclosing test instance, so the logger never becomes weakly reachable and the test fails after 50 GC attempts.

Fix:
Replacing the SAM lambda with an explicit anonymous object expression to match what the Java version was doing.

Reviewed By: cortinico

Differential Revision: D108765169


cortinico (approved):

PR #1999, opened 2026-07-22
docs: note exception-safe cleanup of JS Yoga trees

greekr4:
## Summary

The JavaScript docs already warn that nodes must be freed manually, but don't mention two things we ran into in production (a WASM-based canvas editor that builds a throwaway Yoga tree per layout pass):

1. Nodes live in the WebAssembly heap, so leaked nodes are invisible to the JS garbage collector and accumulate until `Node.create()` fails.
2. If a measure function (or any code between create and free) throws, the whole tree leaks unless freeing happens in a `finally` block.

This adds a short paragraph and a `try`/`finally` example to the existing warning in the JavaScript tab of "Laying out a Yoga tree".

## Test Plan

Docs-only change. Verified the MDX renders (admonition + code fence) and the relative link to `external-layout-systems.mdx` resolves.

### DIFF
--- a/website/docs/getting-started/laying-out-a-tree.mdx
+++ b/website/docs/getting-started/laying-out-a-tree.mdx
@@ -99,6 +99,18 @@ root.insertChild(child1, 1);
 
 Yoga Nodes are not freed automatically and should be discarded when no longer needed. Individual nodes may be freed by calling `node.free()`, or an entire Yoga tree may be freed by calling `node.freeRecursive()`.
 
+Nodes live in the WebAssembly heap, which is not visible to the JavaScript garbage collector. An application which repeatedly leaks nodes will grow the heap until new nodes fail to allocate. If an exception may be thrown after nodes are created (e.g. from within a [measure function](../advanced/external-layout-systems.mdx) during layout), freeing in a `finally` block guarantees the tree is released:
+
+```ts
+const root = buildYogaTree();
+try {
+  root.calculateLayout(undefined, undefined, Direction.LTR);
+  readLayoutResults(root);
+} finally {
+  root.freeRecursive();
+}
+```
+
 A future revision of JavaScript bindings for Yoga may move to garbage collection to remove this requirement.
 
 :::


website/docs/getting-started/laying-out-a-tree.mdx
```diff
@@ -99,6 +99,18 @@ root.insertChild(child1, 1);
...
+Nodes live in the WebAssembly heap, which is not visible to the JavaScript garbage collector. An application which repeatedly leaks nodes will grow the heap until new nodes fail to allocate. If an exception may be thrown after nodes are created (e.g. from within a [measure function](../advanced/external-layout-systems.mdx) during layout), freeing in a `finally` block guarantees the tree is released:
+
+```ts
+const root = buildYogaTree();
+try {
+  root.calculateLayout(undefined, undefined, Direction.LTR);
+  readLayoutResults(root);
+} finally {
+  root.freeRecursive();
+}
```
andrewdacenko: The new snippet references buildYogaTree() and readLayoutResults(root), which are not defined anywhere in the docs, and uses Direction.LTR without an import (the snippet above this warning imports only Edge, FlexDirection, PositionType from yoga-layout). Other snippets in this page are self-contained and copy-pasteable; readers copying this block would hit unresolved identifiers. It also re-declares const root, which conflicts with the const root declared in the preceding snippet of the same JavaScript tab if both blocks are pasted together. Consider inlining the tree construction/result reading (or using comments as placeholders) and importing/mentioning Direction.

```suggestion
try {
  root.calculateLayout(undefined, undefined, Direction.LTR);
  // Read layout results, e.g. child0.getComputedLeft()
} finally {
  root.freeRecursive();
}
```

andrewdacenko (changes requested):

website/docs/getting-started/laying-out-a-tree.mdx
greekr4: Good catch, thanks — applied your suggestion in 51cbdce so the block reuses `root`/`child0` from the snippet above, and added `Direction` to that import (which also covers the `calculateLayout` snippet further down the page).

andrewdacenko (approved):

PR #2001, opened 2026-07-24
Remove disabled intrinsic-sizing test cases

ellemac123:
Summary: Removes the disabled (`data-disabled`) intrinsic-sizing test cases from the gentest fixtures and their generated Java/C++/JS tests. These cases (fit-content / max-content / stretch, several gated behind the `FixFlexBasisFitContent` experiment) were checked in disabled because Yoga's layout does not yet match the browser reference for them, so they only ever produce skipped tests. Remove them for now; re-add and enable once the behavior is implemented. Also removes the now-orphaned TODO comments that annotated those cases.

Reviewed By: GijsWeterings, cortinico

Differential Revision: D113528568


cortinico (approved):

